PR #265, opened 2024-04-29
Real Time Sync

MelonSpeedruns:


garrettjoecox (changes requested):
This should be "hookified"

garrettjoecox (approved):
There are some implications this has on gameplay, specifically with the 3 day cycle, but I'm fine with merging for now as more of a gimmick and attempting to rectify how one could actually do a real playthrough with this later.

garrettjoecox (changes requested):
Oh actually one last thing,

You'll need to call `RegisterRealTimeSync()` from the `InitEnhancements` function, so that it registers the hook on startup. Otherwise it gets out of sync with the cvar.

garrettjoecox (approved):

PR #289, opened 2024-05-02
Add time-moves-when-you-move mode

garrettjoecox:
Time speed is based on link's linear velocity, with time stopped when link isn't moving, and progressing the fastest when rolling as goron link

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1491844256.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1491849128.zip)


inspectredc (commented):
Looks good to me, what's our thoughts on adding modes etc. in general? In SoH it's at the point where a lot of these kinda things would usually cause hesitation due to the clutter of enhancements

garrettjoecox:
Moving this PR to my new fork

PR #383, opened 2024-05-13
Fix save editor dpad names

Archez:
The menu item names in the save editor right click equip did not match the actual button that gets assigned.


garrettjoecox (approved):

inspectredc (approved):
🥖

PR #396, opened 2024-05-15
Prevent infinite loop in save migration

Archez:
The save migrations were not gracefully handling saves that had a "higher" version than the current version for the build. The while loop would execute forever due to a strict inequal.

For now I changed it to a less than to prevent the loop and just allow the code to continue.

We probably need to think of what we actually want to do when this occurs.
Do nothing and just tell users that using newer saves on older builds aren't expected to work?
Do we set the version number back and make saves gracefully load zero values for things missing?
Do we gray out the save on the file menu? (a la SoH MQ) (assuming loading the unknown json doesn't crash)


inspectredc (approved):

garrettjoecox (approved):
There are several edge cases we need to handle more gracefully here, I think the MVP route will be “if we think it’s going to cause issues, rename the save file and don’t attempt to load it”

been on my todo list but that keeps growing

PR #397, opened 2024-05-15
Skip first cycle reset cutscene on debug created saves

Archez:
Setting the `threeDayResetCount` to >= 0 prevents the "first reset" cutscene which reverts link into deku form. Applying to both types of debug inventory saves since those cutscenes are long and not needed for debug saves.


inspectredc (approved):

garrettjoecox (commented):
Are we augmenting the debug save from the decomp here, or did we add it on our own?

PR #415, opened 2024-05-18
Add persistent bunny mask enhancement

garrettjoecox:
This enhancement allows you to always keep the bunny hood on, in combination with other masks and even other transformations.

I'm pretty satisfied with it's isolation from the rest of the codebase, ~the only thing I need help on is making the indicator on kaleidoscope look better, it looks a bit odd right now cc @Archez~

The user experience of using this should be intuitive, it allows you to equip/un-equip the mask like normal, and lets you press "A" on the mask in kaleido to equip/unequip it

I intend to also add an option for doing the same thing with the Great fairy mask, but want to keep this PR separate from that.


garrettjoecox:
Moving this PR to my new fork

PR #482, opened 2024-05-26
Add draw distance enhancements for actors and scene geometry

garrettjoecox:
Depends on https://github.com/Kenix3/libultraship/pull/605

- Adds a checkbox for disabling scene geometry culling
- Add slider for multiplying actor draw distance check
- Add slider for multiplying actor update distance check


garrettjoecox:
Moving this PR to my new fork

PR #539, opened 2024-05-27
Add draw distance enhancements for actors and scene geometry

garrettjoecox:
Depends on https://github.com/Kenix3/libultraship/pull/605

- Adds a checkbox for disabling scene geometry culling
- Add slider for multiplying actor draw distance check
- Add slider for multiplying actor update distance check

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1561105079.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1561107971.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1561109836.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenuBar.cpp
+++ b/mm/2s2h/BenGui/BenMenuBar.cpp
@@ -532,6 +532,37 @@ void DrawEnhancementsMenu() {
                              "minor visual glitches that were covered up by the black bars\nPlease disable this "
                              "setting before reporting a bug" });
 
+            ImGui::PushStyleColor(ImGuiCol_Text, ImVec4(255, 255, 0, 255));
+            ImGui::SeparatorText("Unstable");
+            ImGui::PopStyleColor();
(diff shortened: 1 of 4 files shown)


garrettjoecox:
Seems like this causes the clock town houses displayed from termina field to be hidden. Not a huge issue cause it's fairly minor and marked as "unstable" for now

inspectredc (approved):
Idk if im a huge fan of the copied function but thats probably just because it's not documented, and I can't think of any other way of doing it really.

garrettjoecox:
> Idk if im a huge fan of the copied function but thats probably just because it's not documented, and I can't think of any other way of doing it really.

Yeah that's fair, it's mostly to ensure we always preserve the vanilla behavior by default just in case we cause some sort of weird calculation error in our version.

PR #550, opened 2024-05-27
[Enhancement] Fast Song Playback

Patrick12115:
Speeds up all song playback, including skipping cutscenes for certain songs and elegy statue spawning.

This will most likely still need work, including moving everything that can be, into hooks. I am just still not knowledgeable enough to figure out the hooks myself, so I will need help with that part. Right now everything is just using CVars, which I know is undesirable.

Song of Time reset may be able to be done more elegantly, this was just the only way I, myself, could figure out how to get it to work properly.

Added lots of comments to explain everything as best as I could.

Reupload of #444 

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1541981281.zip)


mckinlee:
There is a conflict with Do not Reset Time Speed. timeSpeedOffset value is hardcoded so when SoT is played it doesn't retain the speed value.

PR #568, opened 2024-05-28
[Enhancement] Friendly dogs

PunishedBigs:
Adds an enhancement option that makes dogs react positively to all of link's forms just as they do for Zora Link normally. Also works for Fierce Deity Link when using the Fierce Deity anywhere enhancement. No issues with Bremen Mask. You just need to get some distance from the dog or change forms to update its behaviour when toggling the enhancement on or off.
Makes idling as Deku Link in South Clock Town while configuring stuff a little easier, but this will probably be obsolete if and when that dog follower mod from SoH gets ported.

Not sure how you guys wanna handle where i've put the enhancement checkbox if you use this. A "Misc" category would probably just end up as a dumping ground for menu bloat down the line but nothing else that was there already really fit so I just went with something generic pending feedback.

### DIFF
--- a/mm/2s2h/BenGui/BenMenuBar.cpp
+++ b/mm/2s2h/BenGui/BenMenuBar.cpp
@@ -574,6 +574,13 @@ void DrawEnhancementsMenu() {
             ImGui::EndMenu();
         }
 
+        if (UIWidgets::BeginMenu("Misc")) {
+            UIWidgets::CVarCheckbox("Friendly Dogs", "gEnhancements.Misc.FriendlyDogs", 
+                                    {.tooltip = "Dogs will always be friendly no matter what form Link is in."});
+
+            ImGui::EndMenu();
+        }
+
         if (mHudEditorWindow) {
             UIWidgets::WindowButton("Hud Editor", "gWindows.HudEditor", mHudEditorWindow,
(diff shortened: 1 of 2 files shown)


PunishedBigs:
Noticed an issue with not being able to pick up the dog while the enhancement is on, which will create issues with racetrack. Will have to sort it when back from work.

PunishedBigs:
Fixed not being able to pickup & throw while dog is following with enhancement active, tested and confirmed you can do racetrack without issue.

garrettjoecox:
In its current state, this is what we consider very “invasive” on the original source code. We should have a contributing.md on the repo soon that goes into more detail, but in summary, adding cvars in the original source code should be a last resort, it’s suggested to try and keep all of your changes in its own file under the enhancements directory using generic hooks. You can look at examples of this in that directory. This isn’t always possible, but will increase the likely hood of your change being merged in

inspectredc:
An Idea I can think for this one would be to hook onto the update function call, and check which function is being called. That way you can override things like `EnDg_ApproachPlayerToAttack` to just always call `EnDg_ApproachPlayer` instead, if the cvar is on.

PunishedBigs:
gotcha, ill come up with something

garrettjoecox:
Going to close this out for now, feel free to re-open after some more iteration, nice work by the way!

PR #578, opened 2024-05-28
Fix Dpad Ocarina L Btn & GI Dpad Stuff

### LINKED ISSUE #544: Ocarina songs that don't use the A button aren't cancelled by L button when playing with D-pad ocarina
Reported by Master L on Discord
Currently, when playing the ocarina with the D-pad using the D-pad ocarina option, holding the L button doesn't properly prevent song recognition _if_ the song doesn't use the A button (Song of Soaring, for example). Songs that use the A button are not affected by this.
This issue doesn't occur when playing the ocarina with the buttons that are mapped to the C button inputs

inspectredc:
Tried Following the Pattern from #441 Here

This should fix #544 

Few thoughts on my part:

- Can we cache the Cvar Result for each frame and reset them?
- Can this work as a hook which sets a bool only changing it when the cvar changes?
- Are Cvars expensive enough that the points above would make sense?

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1543736588.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1543747405.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1543870590.zip)

### DIFF
--- a/mm/2s2h/Enhancements/GameInteractor/GameInteractor.cpp
+++ b/mm/2s2h/Enhancements/GameInteractor/GameInteractor.cpp
@@ -230,3 +230,22 @@ int GameInteractor_InvertControl(GIInvertType type) {
 
     return result;
 }
+
+uint32_t GameInteractor_Dpad(GIDpadType type, uint32_t buttonCombo) {
+    uint32_t result = 0;
+
+    switch (type) {
+        case GI_DPAD_OCARINA:
+            if (CVarGetInteger("gEnhancements.Playback.DpadOcarina", 0)) {
+                result = buttonCombo;
+            }
+            break;
+        case GI_DPAD_EQUIP:
+            if (CVarGetInteger("gEnhancements.Dpad.DpadEquips", 0)) {
+                result = buttonCombo;
+            }
+            break;
+    }
+
+    return result;
+}
--- a/mm/2s2h/Enhancements/GameInteractor/GameInteractor.h
+++ b/mm/2s2h/Enhancements/GameInteractor/GameInteractor.h
@@ -58,6 +58,11 @@ typedef enum {
     GI_INVERT_CAMERA_RIGHT_STICK_Y,
 } GIInvertType;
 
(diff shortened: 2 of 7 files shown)


mm/2s2h/Enhancements/GameInteractor/GameInteractor.cpp
```diff
@@ -230,3 +230,22 @@ int GameInteractor_InvertControl(GIInvertType type) {
 
     return result;
 }
+
+uint32_t GameInteractor_Dpad(GIDpadType type, uint32_t buttonCombo) {
```
garrettjoecox, line 234: So what I had in mind here was more generic than being specific to just dpad, but to just controls in general. For instance, you call `GameInteractor_Btn(GI_BTN_OCARINA_UP)`, and a player could opt in to using their dpad, or even right stick to count as "OCARINA_UP".

Does that make sense? Also not sold on the name `GI_BTN`, just an example

inspectredc: I think sticks and buttons are a bit too different to handle them like that honestly but I get the generalisation yea.

garrettjoecox: I want to revisit this, but when we know more / have a better idea

garrettjoecox (approved):

mm/2s2h/Enhancements/GameInteractor/GameInteractor.cpp
inspectredc: Yea sure, do you think I should make a separate PR with the fix?

inspectredc: oh wait nvm ur approving

PR #579, opened 2024-05-28
Display GC Version String On Title

inspectredc:
null

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1544809115.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1544813966.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1544824544.zip)

### DIFF
--- a/mm/src/overlays/gamestates/ovl_title/z_title.c
+++ b/mm/src/overlays/gamestates/ovl_title/z_title.c
@@ -26,9 +26,11 @@ const char* GetGameVersionString() {
     uint32_t gameVersion = ResourceMgr_GetGameVersion(0);
     switch (gameVersion) {
         // BENTODO: Use enums from LUS once added and add rest of version names
-        case 0x5354631C:
+        case MM_NTSC_US_10:
         case 0xDA6983E7:
             return "MM-US 1.0";
+        case MM_NTSC_US_GC:
+            return "MM-US GC";
         default:
             return "UNKNOWN";
     }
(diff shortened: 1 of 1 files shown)


mm/src/overlays/gamestates/ovl_title/z_title.c
```diff
@@ -26,9 +26,11 @@ const char* GetGameVersionString() {
     uint32_t gameVersion = ResourceMgr_GetGameVersion(0);
     switch (gameVersion) {
         // BENTODO: Use enums from LUS once added and add rest of version names
-        case 0x5354631C:
+        case MM_NTSC_US_10:
         case 0xDA6983E7:
```
Archez: Would you wanna add this as `MM_NTSC_US_UNCOMPRESSED`, so then we can remove the TODO?

inspectredc: sure, although I think I'm going to have to redo some stuff later now and check against uncompressed game versions for certain sections (i.e. any checks that specifically target MM_NTSC_US_GC need to check the uncompressed version too). I was wondering if it would be better to just consolodate this on the exporter side so that the version would be set the same regardless of compression

Archez: My thinking is that we shouldn't bother with explicitly supporting uncompressed for all other roms. (By this I mean not having `_UNCOMPRESSED` handled throughout the code base).
N64 US uncompressed exists mostly because of decomp.

Ultimately I agree that it would be good if we could handle it at the extractor side, but until then, I'd say its not important to handle each rom.

inspectredc: I agree, so will just remove that case then if that's alright?

garrettjoecox (approved):

PR #586, opened 2024-05-28
Fix Half-Functional Text Control Codes

Qlonever:
Fixes a few text control codes (quicktext enable, SFX, delay) that were bugged in the original NA release. It seems these weren't properly adjusted to account for two characters being displayed each frame, so they only worked about half of the time.

Rika Alfa:
https://github.com/HarbourMasters/2ship2harkinian/assets/42286723/91f7743b-c159-4ceb-97d1-8189badc194e

This PR:
https://github.com/HarbourMasters/2ship2harkinian/assets/42286723/e8193c91-cb11-4d80-8e81-779533aae779

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1549522809.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1549524001.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1549532323.zip)

### DIFF
--- a/mm/src/code/z_message_nes.c
+++ b/mm/src/code/z_message_nes.c
@@ -431,7 +431,10 @@ void Message_DrawTextNES(PlayState* play, Gfx** gfxP, u16 textDrawPos) {
                     msgCtx->textColorR = msgCtx->textColorG = msgCtx->textColorB = 255;
                 }
 
-                if ((i + 1) == msgCtx->textDrawPos) {
+                if ((i + 1) == msgCtx->textDrawPos ||
+                    // 2S2H [Enhancement] To match how PAL and GC fixed character parsing, any code that checks for
+                    // (i + 1) == msgCtx->textDrawPos should also check for (i + 2) == msgCtx->textDrawPos
+                    ((i + 2) == msgCtx->textDrawPos && CVarGetInteger("gEnhancements.Fixes.ControlCharacters", 0))) {
                     msgCtx->textDrawPos++;
                 }
                 break;
@@ -495,7 +498,8 @@ void Message_DrawTextNES(PlayState* play, Gfx** gfxP, u16 textDrawPos) {
                     msgCtx->textColorB = D_801D080C[(s16)((u8)character - 1)].b;
                 }
 
-                if ((i + 1) == msgCtx->textDrawPos) {
+                if ((i + 1) == msgCtx->textDrawPos ||
+                    ((i + 2) == msgCtx->textDrawPos && CVarGetInteger("gEnhancements.Fixes.ControlCharacters", 0))) {
                     msgCtx->textDrawPos++;
                 }
                 break;
@@ -554,7 +558,8 @@ void Message_DrawTextNES(PlayState* play, Gfx** gfxP, u16 textDrawPos) {
                 return;
 
             case 0x17: // MESSAGE_QUICKTEXT_ENABLE
-                if ((i + 1) == msgCtx->textDrawPos) {
+                if ((i + 1) == msgCtx->textDrawPos ||
+                    ((i + 2) == msgCtx->textDrawPos && CVarGetInteger("gEnhancements.Fixes.ControlCharacters", 0))) {
                     if ((msgCtx->msgMode == MSGMODE_TEXT_DISPLAYING) ||
                         ((msgCtx->msgMode >= MSGMODE_OCARINA_STARTING) && (msgCtx->msgMode <= MSGMODE_26))) {
                         j = i;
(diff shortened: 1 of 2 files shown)


garrettjoecox (commented):
Ayyy nice! 

Should this be optional to account for the differences in versions? (I'm thinking yes..)

Qlonever:
> Should this be optional to account for the differences in versions? (I'm thinking yes..)

I'm not entirely sure what you mean by this, but I'm willing to add a checkbox just for the sake of fixes being optional.

Malkierian:
I think what Proxy was asking was whether we needed to account for the eventual support of PAL ROMs, or if the code will work with their control character flow automatically.

garrettjoecox:
Mostly just that out of the box, the default behavior (aside from rare exceptions like widescreen support or fixing crashes) should reflect US 1.0, anything else should be opt-in.

We may in the future do smart toggling of these options based on the rom they extracted.

Qlonever:
> Mostly just that out of the box, the default behavior (aside from rare exceptions like widescreen support or fixing crashes) should reflect US 1.0, anything else should be opt-in.

Yeah, that makes sense.
https://github.com/HarbourMasters/2ship2harkinian/assets/42286723/6cfb4f51-073e-41cd-9450-30d7c4f08145

>I think what Proxy was asking was whether we needed to account for the eventual support of PAL ROMs, or if the code will work with their control character flow automatically.

>We may in the future do smart toggling of these options based on the rom they extracted.

As far as I can tell, this functions the same as in PAL and GC, so toggling this on when extracting those ROMS would make sense.

garrettjoecox (approved):

inspectredc:
> Ayyy nice!
> 
> Should this be optional to account for the differences in versions? (I'm thinking yes..)

I imagine we can use z_message_nes (slightly modify it) for PAL too based one what small investigations I've done. z_message would need this treatment for JP support to have it too, but that's something I can add on my PRs end

mm/src/code/z_message_nes.c
```diff
@@ -621,19 +622,24 @@ void Message_DrawTextNES(PlayState* play, Gfx** gfxP, u16 textDrawPos) {
...
+                    (msgCtx->msgMode == MSGMODE_TEXT_DISPLAYING)) {
                     stateTimerHi = msgCtx->decodedBuffer.schar[i + 1] << 8;
                     stateTimerHi += (uint8_t)msgCtx->decodedBuffer.schar[i + 2];
                     Audio_PlaySfx(stateTimerHi);
                 }
-                if ((i + 1) == msgCtx->textDrawPos) {
-                    msgCtx->textDrawPos += 2;
+                if ((i + 1) == msgCtx->textDrawPos ||
+                    ((i + 2) == msgCtx->textDrawPos && CVarGetInteger("gEnhancements.Fixes.ControlCharacters", 0))) {
+                    msgCtx->textDrawPos += CVarGetInteger("gEnhancements.Fixes.ControlCharacters", 0) ? 3 : 2;
```
Archez: If I understand correctly, the way this works is a look-ahead is happening, and based on the result being true these actions are performed.

Since you are increasing the look-ahead to now considered two possible `i` values, I would think that the operations now need to handle that in each of these branch paths.

For example, you arbitrarily changed the `msgCtx->textDrawPos +=` from 2 -> 3, but other values in your branch paths don't change how `i` is used.

It seems like to me what should be happening is that if the case is `(i + 1) == msgCtx->textDrawPos` then the increment is 2, if the case is `(i + 2) == msgCtx->textDrawPos` then the increment should be 3.

Most likely the other paths need similar considerations.

Qlonever: > If I understand correctly, the way this works is a look-ahead is happening, and based on the result being true these actions are performed.

My understanding is that `i` represents the position of a character in the message data being read, and  `msgCtx->textDrawPos` represents the position of the last character that should be read on the current frame, plus one. The original code looks ahead to see if the control character is the last to be read, and perform its function if it is. However, since two characters are displayed per-frame in the US version, it should have checked if these control characters were second-to-last as well.

>It seems like to me what should be happening is that if the case is (i + 1) == msgCtx->textDrawPos then the increment is 2, if the case is (i + 2) == msgCtx->textDrawPos then the increment should be 3.

I think the purpose of this increment is to ensure that the SFX control character and the two parameter bytes after it don't count towards the characters displayed on the current frame. If that's the case, it makes sense to increment `msgCtx->textDrawPos` by 3, since three bytes of data are being ignored. This would be true for both the `(i+1)` and `(i+2)` cases, so I think the original increment of 2 is just a mistake.
I tested multiple values here, and only incrementing by 3 in both cases prevented SFX from disturbing the text flow.

>Most likely the other paths need similar considerations.

I don't think the other cases I modified (quicktext & delay) need similar considerations, since they're meant to alter the text flow anyway. It's possible I'm misunderstanding the purpose of some of this code, though.

mm/src/code/z_message_nes.c
```diff
@@ -554,7 +554,8 @@ void Message_DrawTextNES(PlayState* play, Gfx** gfxP, u16 textDrawPos) {
                 return;
 
             case 0x17: // MESSAGE_QUICKTEXT_ENABLE
-                if ((i + 1) == msgCtx->textDrawPos) {
+                if ((i + 1) == msgCtx->textDrawPos ||
+                    ((i + 2) == msgCtx->textDrawPos && CVarGetInteger("gEnhancements.Fixes.ControlCharacters", 0))) {
                     if ((msgCtx->msgMode == MSGMODE_TEXT_DISPLAYING) ||
```
louist103, line 563:
```
// 2S2H [Port]
```
Type comments should be added here. Kinda like this https://github.com/HarbourMasters/2ship2harkinian/blob/291973fffeb9c5c41e9269d5731c296edd2f1771/mm/src/overlays/actors/ovl_En_Door/z_en_door.c#L502

louist103 (commented):
Looks good to me. Just one thing.

mm/src/code/z_message_nes.c
Qlonever: Added a comment to my first change in the file.

mm/src/code/z_message_nes.c
Qlonever: After thinking about this more, I decided to revert this particular change. Looking at z_message.c, the original intent may have been to count the SFX control character towards the number of characters read, but not its parameter bytes for whatever reason. PAL seems to increment `msgCtx->textDrawPos` by 2 as well. 
This still checks for `(i + 2) == msgCtx->textDrawPos`, but only increments by 2 either way.

PR #551, opened 2024-05-27
Add 2ship-specific BUILDING.md and supportedHashes.json.

Malkierian:
null

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1555146650.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1555152687.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1555159396.zip)

PR #588, opened 2024-05-28
ci(linux): install libzip without crypto, bump sdl from 2.28.5 -> 2.30.3

briaguya0:
null

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1546269247.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1546270532.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1546276998.zip)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -152,7 +152,6 @@ jobs:
         cmake ..
         make
         sudo make install
-        sudo cp -av /usr/local/lib/libzip* /lib/x86_64-linux-gnu/
     - name: Generate 2ship.o2r
       run: |
         export PATH="/usr/lib/ccache:/usr/local/opt/ccache/libexec:$PATH"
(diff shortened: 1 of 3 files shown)


garrettjoecox (approved):
I assume we’ll need to update #551 ?

briaguya0:
> I assume we’ll need to update #551 ?

nope, if you have libzip with crypto support installed locally everything is fine, this is just because the appimage will add all linked libs and having libcrypto in there (or having it linked and not having it in there) was problematic.

.github/workflows/main.yml
```diff
@@ -114,7 +114,7 @@ jobs:
     - name: Install dependencies
       run: |
         sudo apt-get update
-        sudo apt-get install -y $(cat .github/workflows/apt-deps.txt)
+        sudo apt-get install -y $(cat .github/workflows/apt-deps.txt) libzip-dev zipcmp zipmerge ziptool
```
Archez, line 117: Why move this out of the deps.txt to here?

briaguya0: because the version installed from apt is links libcrypto

i could just do an apt remove in the build 2ship job before installing it from source if you'd prefer that

Archez: Ah, this is the diff for the o2r runner. I confused it with the Linux build down below. Now I understand 👍

Archez (approved):

PR #589, opened 2024-05-28
[Enhancement] Instant Recall

Patrick12115:
Pressing B will instantly recall Zora Link's Fin Boomerangs when thrown.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1550280541.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1550283112.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1550287043.zip)


garrettjoecox (commented):
Oh he’s a hooks kind of guy

mm/2s2h/BenGui/BenMenuBar.cpp
```diff
@@ -537,6 +537,16 @@ void DrawEnhancementsMenu() {
             }
             ImGui::EndMenu();
         }
+
+        if (UIWidgets::BeginMenu("Player Actions")) {
```
garrettjoecox: Maybe Items/Equipment might be better? I'm not sure what else would go in this category

Patrick12115: I originally started using that when I did Instant Putaway, but it could probably go under Items/Equipment too

mm/2s2h/Enhancements/Equipment/InstantRecall.cpp
```diff
@@ -0,0 +1,44 @@
...
+}
+
+void RegisterInstantRecall() {
+    GameInteractor::Instance->UnregisterGameHookForID<GameInteractor::OnActorUpdate>(onActorUpdateHookId);
+    onActorUpdateHookId = 0;
+
+    onActorUpdateHookId = GameInteractor::Instance->RegisterGameHookForID<GameInteractor::OnActorUpdate>(
+        ACTOR_EN_BOOM, [](Actor* outerActor) {
+            if (CVarGetInteger("gEnhancements.PlayerActions.InstantRecall", 0)) {
+                if (gPlayState->state.input->cur.button == BTN_B) {
```
inspectredc: this should be using CHECK_BTN_ALL, otherwise this will only return true if only the b button and nothing else is being held. I'm wondering if it should be using press instead of cur too

```suggestion
                if (CHECK_BTN_ALL(gPlayState->state.input->cur.button, BTN_B)) {
```

inspectredc (dismissed):
Looks good, just one small thing re: the input

mm/2s2h/Enhancements/Equipment/InstantRecall.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+    EnBoom* boomerangs = (EnBoom*)player->boomerangActor;
+
+    if (boomerangs != NULL) {
+        // Set the positions of both boomerangs
+        boomerangs->actor.world.pos = player->actor.world.pos;
+        if (boomerangs->actor.child != NULL) {
+            boomerangs->actor.child->world.pos = player->actor.world.pos;
+        }
+
```
Archez: It looks like moving the boomerangs position is unnecessary if we are also calling `Actor_Kill` on them at the same time. The boomerangs get cleaned up without issue.

```suggestion
```

mm/2s2h/Enhancements/Equipment/InstantRecall.cpp
Archez: I agree, and yeah I think `press` would be better to use as well.

```suggestion
                if (CHECK_BTN_ALL(gPlayState->state.input->press.button, BTN_B)) {
```

Archez (approved):
Pulled in latest develop and applied the suggested changes. This should be good to go now 👍

PR #590, opened 2024-05-29
[Graphics] Disable Black Bar Letterboxes

Patrick12115:
Removed the black bars at the top and bottom of the screen during cutscenes, Z targeting, etc

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1554096838.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1554102216.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1554116601.zip)


mm/src/code/z_shrink_window.c
```diff
@@ -21,7 +22,9 @@ ShrinkWindow sShrinkWindow;
 ShrinkWindow* sShrinkWindowPtr;
 
 void ShrinkWindow_Letterbox_SetSizeTarget(s32 target) {
-    sShrinkWindowPtr->letterboxTarget = target;
+    if (GameInteractor_Should(GI_VB_SHOW_BLACK_BARS, true, NULL)) {
+        sShrinkWindowPtr->letterboxTarget = target;
+    }
```
garrettjoecox: Any benefit to overriding this in the set target over it being in the draw? If it was in the draw, you could toggle it mid-cutscene to see the results immediately

Archez: Overriding in the shrink draw would also require overriding the view handling for letterbox in `View_ApplyLetterbox`

Patrick12115: Thanks @Archez! That was what I couldn't figure out with doing it in the draw

mm/src/code/z_shrink_window.c
```diff
@@ -79,35 +80,38 @@ void ShrinkWindow_Draw(GraphicsContext* gfxCtx) {
...
-                             OTRGetRectDimensionFromRightEdge(gScreenWidth), letterboxSize + 1);
-        gDPFillWideRectangle(gfx++, OTRGetRectDimensionFromLeftEdge(0), gScreenHeight - letterboxSize - 1,
-                             OTRGetRectDimensionFromRightEdge(gScreenWidth), gScreenHeight - letterboxSize);
-        // #endregion
-
-        gDPPipeSync(gfx++);
-        OVERLAY_DISP = gfx++;
-
-        CLOSE_DISPS(gfxCtx);
+    if (GameInteractor_Should(GI_VB_SHOW_LETTERBOX, true, NULL)) {
```
Archez: Since both paths don't do anything extra that we need to run, a cleaner approach would be to invert your "should" and just have the `if` condition be an early return. This way we aren't increasing the indentation of all this code.

```c
if (GameInteractor_Should(GI_VB_HIDE_LETTERBOX, false, NULL)) {
    return;
}
```

Archez (approved):

mm/2s2h/Enhancements/Graphics/DisableBlackBars.cpp
```diff
@@ -0,0 +1,11 @@
...
+#include "Enhancements/GameInteractor/GameInteractor.h"
+#include "DisableBlackBars.h"
+
+void RegisterDisableBlackBars() {
+    REGISTER_VB_SHOULD(GI_VB_DISABLE_LETTERBOX, {
+        if (CVarGetInteger("gEnhancements.Graphics.DisableBlackBars", 0)) {
+            *should = true;
+        }
+    });
+}
```
garrettjoecox: Whenever this little red circle symbol shows up that means you are missing a newline at the end of the file, it's considered a generally good practice to always have an empty newline at the end of of files

garrettjoecox (approved):

mm/2s2h/Enhancements/Graphics/DisableBlackBars.cpp
Patrick12115: Ooh, I was wondering what those were. Just assumed it meant end of file or something. I'll fix it when I get home

mm/2s2h/Enhancements/Enhancements.cpp
```diff
@@ -22,6 +22,9 @@ void InitEnhancements() {
 
     // Equipment
     RegisterSkipMagicArrowEquip();
+
```
garrettjoecox:
```suggestion

```

mm/2s2h/Enhancements/Enhancements.cpp
```diff
@@ -22,6 +22,8 @@ void InitEnhancements() {
 
     // Equipment
     RegisterSkipMagicArrowEquip();
+    // Graphics
```
garrettjoecox, line 26:
```suggestion

    // Graphics
```

PR #591, opened 2024-05-29
[Enhancement] Banker - Adds Min Max Functionality

Caladius:
Pressing Z or R while the Deposit Rupees dialogue is open will set the rupees to the amount Link is currently carrying or to 0 respectively.

Adds new "Actor" category segmented by Scene Location.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1559454943.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1559455614.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1559457466.zip)


mm/2s2h/Enhancements/Actor/BankerDialogue.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include "overlays/actors/ovl_En_Ginko_Man/z_en_ginko_man.h"
+
+void RegisterBankerDialogue() {
+
+    GameInteractor::Instance->RegisterGameHook<GameInteractor::OnActorUpdate>([](Actor* actor) {
+        if (!CVarGetInteger("gEnhancements.Actor.BankerDepositRupees", 0)) {
+            return;
+        }
+        if (actor->id == ACTOR_EN_GINKO_MAN) {
```
Archez: It seems like the only use of the `actor` is right here against the banker ID, which ends up translating to just an expensive "execute once per frame" (expensive because this is executing for all actors per frame).

Since the real check is just against the current text ID, and only the banker can make the text ID appear, then you should be able to switch to `OnGameStateUpdate` or similar hook  instead. Just remember to perform a null check on `gPlayState`

mm/2s2h/Enhancements/Actor/BankerDialogue.cpp
```diff
@@ -0,0 +1,41 @@
...
+                    char secondChar = ((gSaveContext.save.saveInfo.playerData.rupees / 10) % 10) + '0';
+                    char thirdChar = (gSaveContext.save.saveInfo.playerData.rupees % 10) + '0';
+                    const char rupeeChar[3] = { firstChar, secondChar, thirdChar };
+
+                    for (int i = 22; i <= 24; i++) {
+                        gPlayState->msgCtx.bankRupeesSelected = gPlayState->msgCtx.decodedBuffer.schar[i] =
+                            rupeeChar[i - 22];
+                        Font_LoadCharNES(gPlayState, gPlayState->msgCtx.decodedBuffer.schar[i],
+                                         gPlayState->msgCtx.unk120C4 + (0 + (i - 22) << 7));
+                    }
```
Archez: Hardcoding this logic for character position `22..24`, is prone to not work well with future other languages. Thankfully there is a variable that tracks the "rupee starting position" (`msgCtx.unk120C0`), and you are already using another variable that helps (`msgCtx.unk120C4`). Re-writing like this would be more appropriate and align with how the original code handles things.

```suggestion
                    for (int i = 0; i <= 2; i++) {
                        gPlayState->msgCtx.decodedBuffer.schar[gPlayState->msgCtx.unk120C0 + i] = rupeeChar[i];
                        Font_LoadCharNES(gPlayState,
                                         gPlayState->msgCtx.decodedBuffer.schar[gPlayState->msgCtx.unk120C0 + i],
                                         gPlayState->msgCtx.unk120C4 + (i << 7));
                    }
```

mm/2s2h/Enhancements/Actor/BankerDialogue.cpp
```diff
@@ -0,0 +1,41 @@
...
+                    Audio_PlaySfx(NA_SE_SY_RUPY_COUNT);
+                }
+                if (CHECK_BTN_ALL(gPlayState->state.input[0].cur.button, BTN_R)) {
+                    for (int i = 22; i <= 24; i++) {
```
Archez: And a similar change to this for loop as well

mm/2s2h/Enhancements/Actor/BankerDialogue.cpp
```diff
@@ -0,0 +1,41 @@
...
+                        gPlayState->msgCtx.bankRupeesSelected = gPlayState->msgCtx.decodedBuffer.schar[i] = '0';
```
Archez: I think you aren't setting `msgCtx.bankRupeesSelected` correctly. Rather than assigning a char value of a number, it looks like this should just be set to the real number `0`, and it doesn't need to be in the loop.

Similarly above you would want to set `msgCtx.bankRupeesSelected` equal to the players rupees.

But actually, the original message handler `func_80148D64` will automatically update `msgCtx.bankRupeesSelected` for you, so you probably could just drop it.

Archez (commented):
Nice work so far, left a few comments about opportunities to clean things up.

mm/2s2h/Enhancements/Actor/BankerDialogue.cpp
garrettjoecox: Additionally, you can filter this hook by actor ID so it’s only ever run for this actor

mm/2s2h/Enhancements/Actor/BankerDialogue.cpp
Caladius: From what I found, setting bankRupeesSelected to a real number instead of a char value caused the text box to become corrupted and do some very strange things.

It wasnt until I figured out it needed the char value did the text box behave as expected but without the Font_LoadCharNES function it would not update the text box display with the numbers from bankRupeesSelected.

mm/2s2h/Enhancements/Actor/BankerDialogue.cpp
```diff
@@ -0,0 +1,64 @@
...
+                        if (gPlayState->msgCtx.currentTextId == 1104) {
+                            if (CHECK_BTN_ALL(gPlayState->state.input[0].cur.button, BTN_Z) &&
+                                gPlayState->msgCtx.bankRupeesSelected != gSaveContext.save.saveInfo.playerData.rupees) {
+                                char firstChar = (gSaveContext.save.saveInfo.playerData.rupees / 100) + '0';
+                                char secondChar = ((gSaveContext.save.saveInfo.playerData.rupees / 10) % 10) + '0';
+                                char thirdChar = (gSaveContext.save.saveInfo.playerData.rupees % 10) + '0';
+                                const char rupeeChar[3] = { firstChar, secondChar, thirdChar };
+
+                                for (int i = 0; i <= 2; i++) {
+                                    gPlayState->msgCtx.bankRupeesSelected =
```
Archez: Bringing this up again: `msgCtx.bankRupeesSelected` should not be assigned the "char value" of a number, it should be real numbers that match the desired rupee count.

Right now if my current rupees are 500, this loop will run three times, assigning over `msgCtx.bankRupeesSelected` each time and then it will end up with a value of `48` which is the char value of `0`.

You aren't seeing any negative effect because `func_80148D64` runs every frame which overwrites your changes  based on the current chars stored in `msgCtx.decodedBuffer.schar`.

It should be moved out of the loop and look like
```c
gPlayState->msgCtx.bankRupeesSelected = gSaveContext.save.saveInfo.playerData.rupees;
```
(and assign zero for the minimum case)

Caladius: I understand now! Updated and thank you.

lscholte:
I am not trying to add scope to this pull request, but have you considered implementing the same idea but for withdrawals, where max corresponds to the size of your wallet (e.g. 99, 200, or 500 rupees)?

Caladius:
> I am not trying to add scope to this pull request, but have you considered implementing the same idea but for withdrawals, where max corresponds to the size of your wallet (e.g. 99, 200, or 500 rupees)?

My mental scope was just to get the thing I was after operational. Adding the same functionality to the Withdrawals never occurred to me but I do think it would make a great addition and +1 to the idea.

I'll take a stab at it and see what works out, thank you for the suggestion <3.

PR #596, opened 2024-05-29
Add options for first person aiming

garrettjoecox:
## Summary
- Adds ability to invert x/y axis for first person aiming
- Adds ability to adjust sensitivity for first person aiming
- Adds ability to control first person aiming with Gyro
- Adds ability to control first person aiming with Right Stick

## Todo
- [x] Make gyro feel better
- [x] clean up code
- [x] Add "Disable auto center"
- [x] Add "Move while in first person"

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2021644357.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2021654051.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2021664353.zip)


jordanbtucker:
I've been testing this out, and the non-inverted look when aiming is working great so far except when using the telescope in the Astral Observatory.

cclark25:
I took a shot at adjusting the logic for camera tilting, which should work as a basis for improved gyro control:
https://github.com/garrettjoecox/2ship2harkinian/pull/4

cclark25:
I wrote up a prototype for the logic that will improve gyro controls.
Basically instead of continuously moving the camera, the game will track "tilt" values, which only move the camera once when the joystick (or gyro) changes position, and will rest the camera to an un-tilted state when the joystick (or gyro) is released.
The actual gyro implementation is left unimplemented and instead it uses the secondary joystick to apply the tilt, but that is enough to demo the functionality.
https://github.com/garrettjoecox/2ship2harkinian/pull/4

cclark25:
@garrettjoecox I realized I can connect my switch joycons to my computer over bluetooth to test the gyro controls.
After playing with my own branch and comparing it to yours, it seems that the clunkiness of the controls on your branch is due to the gyro offset getting passed through the smoothing function `Math_SmoothStepToS`.

The gyro offsets should be added directly onto the final rotation values.

jordanbtucker:
While these are all great enhancements, is the reason this PR is taking so long to merge because it's trying to do too much? Maybe it should be split up so we can get some of the already implemented features (like x/y axis for first person aiming) merged without having to wait on the rest of the this PR.

garrettjoecox:
> While these are all great enhancements, is the reason this PR is taking so long to merge because it's trying to do too much? Maybe it should be split up so we can get some of the already implemented features (like x/y axis for first person aiming) merged without having to wait on the rest of the this PR.

The primary reason is I've been on a bit of a hiatus lately (some burnout + new baby) Anyone is open to pick up where I left off, otherwise I'll get to it eventually

jordanbtucker:
@garrettjoecox I completely understand that. I don't mean to come off pushy or ungrateful. I'm just excited to get all of the QoL improvements as quick as possible. I've tried looking into the codebase, but it's a little over my head. I appreciate all your hard work, and I will wait patiently.

garrettjoecox:
Did some more iteration on this, I feel it's ready to be merged. Someone may come in later with a better gyro implementation but thanks to the tip from @cclark25 I think I have it feeling much better than before.

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -688,12 +698,110 @@ void AddEnhancements() {
...
+                  "gEnhancements.Camera.FirstPerson.SensitivityX",
+                  "Adjusts the Sensitivity of the X Axis in First Person Mode.",
+                  WIDGET_CVAR_SLIDER_FLOAT,
+                  { 10.0f, 200.0f, 100.0f, .format = "%.0f%%", .isPercentage = true },
+                  nullptr },
+                { "Y Axis Sensitivity: %.0f%%",
+                  "gEnhancements.Camera.FirstPerson.SensitivityY",
+                  "Adjusts the Sensitivity of the Y Axis in First Person Mode.",
+                  WIDGET_CVAR_SLIDER_FLOAT,
+                  { 10.0f, 200.0f, 100.0f, .format = "%.0f%%", .isPercentage = true },
```
Archez: All these options initializers here and below can't have mixed unnamed and named designators. It has to be all named to be valid (and in constructor order)
https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers

garrettjoecox: Aha thanks, all the more reason to just always have them. I’ll fix tomorrow

garrettjoecox: Fixed

Archez (approved):
Concise and controls feel good. LGTM 👍

cclark25:
The gyro controls still don't quite feel right to me.

The default sensitivity is way too high at 100% in both directions. The X sensitivity should be more like 25% of what it is currently, and the Y sensitivity should be more like 10%.

Also I think the axis inversion is backwards. I had to invert both the X and Y directions to make it work in the natural direction.

I am using my Nintendo Switch Joycons, in case that is some kind of hardware difference.

cclark25:
Hmm oddly enough, the issues with the sensitivity and the axis inversion I listed before are now gone after closing the game and re-opening it. Perhaps something was wrong with my calibration the first time.

garrettjoecox:
> Hmm oddly enough, the issues with the sensitivity and the axis inversion I listed before are now gone after closing the game and re-opening it. Perhaps something was wrong with my calibration the first time.

That is odd, but good to hear, so you think it feels alright with the new changes?

mm/src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -13004,7 +13004,121 @@ void Player_Destroy(Actor* thisx, PlayState* play) {
...
+        stickX += rightStickX * CVarGetFloat("gEnhancements.Camera.FirstPerson.RightStickSensitivityX", 1.0f);
+        stickY += rightStickY * CVarGetFloat("gEnhancements.Camera.FirstPerson.RightStickSensitivityY", 1.0f);
+    }
+
+    stickX = CLAMP(stickX, -60, 60);
+    stickY = CLAMP(stickY, -60, 60);
+
+    if (!func_800B7128(this) && !func_8082EF20(this) && !arg2) { // First person without weapon
+        var_s0 = stickY * 0xF0;
+        if (CVarGetInteger("gEnhancements.Camera.FirstPerson.DisableFirstPersonAutoCenterView", 0)) {
```
cclark25: IMO the auto-center should also ALWAYS be disabled if gyro controls are enabled.
Enabling it while gyro is active serves no purpose and may only confuse the user, and the user might not be familiar with the setting unless they have already played around with it without gyros enabled in the first place.

cclark25:
@garrettjoecox Yeah it looks good. I would just recommend that one thing about the auto-center functionality.
The gyro controls feel way better than before, although they still feel somewhat slow to respond to me, though I cannot figure out why, so that may just be in my head.

mm/src/overlays/actors/ovl_player_actor/z_player.c
garrettjoecox: Good call, made that change

cclark25 (approved):
I tested it again after the last pushed change and it is still working great.
Disabling the auto-center automatically when gyro is enabled works as expected.
It also doesn't seem like there is a delay in the gyro controls, so perhaps my controller was facing some kind of interference causing a delay before.

PR #597, opened 2024-05-29
Display a MSAA value of 1x as "Off"

sheepytina:
Rename `MSAA` to `Anti-aliasing (MSAA)` and display a value of `1` as `Off`. Sample counts of 2anti2aliasing and above display as normal. Tooltip has also been slightly rewritten.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1548779294.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1548783927.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1548793858.zip)


inspectredc (approved):
I think maybe we should also include off in the recommended, or say recommended usage, since we default to off; but it's not really a big deal so otherwise LGTM

Archez (approved):

PR #603, opened 2024-05-29
Added F3DEX_GBI_2 define

KiritoDv:
null


briaguya0:
this should make it in via https://github.com/HarbourMasters/2ship2harkinian/pull/614

PR #608, opened 2024-05-30
Fix Skybox star colors

### LINKED ISSUE #570: Stars seem to be displaying incorrectly
May be related to #492, but not sure.

### Description of the inaccuracy:

Stars in the sky are much more visible (bigger?) than they should be, they jitter and don't move smoothly when rotating the camera (which may be the issue raised in #492) and they appear to be much more saturated (pink stars looking red).

In comparison, Zelda64Recomp displays them differently, which looks accurate to original hardware.

To get a good comparison go into Termina field, they start to appear at 7:30pm.

Archez:
The Skybox star colors were appearing generally pinkish due to endianness ordering with `Color_RGBA8_u32`

```c
typedef union {
    struct {
#ifdef IS_BIGENDIAN
        uint8_t r, g, b, a;
#else
        uint8_t a, b, g, r;
#endif
    };
    uint32_t rgba;
} Color_RGBA8_u32;
```

~~To fix this I've added designator field names so that the values are explicitly tied to the correct field regardless of the machines native endianness.~~

~~I would've preferred a macro solution as a wrapper around all these variables, but I couldn't come up with something elegant due to the initializer brackets. Anything to get around that was equally as invasive as adding these designators, but less readable.~~

Edit: Using a solution that switches `Color_RGBA8_u32` -> `Color_RGBA8` and `gDPSetColor` -> `gDPSetPrimColor`

Screen Shot 2024-05-29 at 9 49 29 PM

Fixes #570 

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1550836485.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1550837904.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1550843388.zip)


garrettjoecox (commented):
Would another solution just be to use `Color_RGBA8` instead?

Archez:
> Would another solution just be to use `Color_RGBA8` instead?

`Color_RGBA8_u32` is a union with `u32`. That u32 value is what gets passed to `gDPSetColor`

`Color_RGBA8` does not have that union type, so we would then have to bitpack all the individual values back together for `gDPSetColor`, or switch to using `gDPSetPrimColor` which handles the packing inside.

Not sure which is more desirable amongst all this.

Archez:
@garrettjoecox heres what RGBA8 with bitpacking could look like https://github.com/Archez/2ship2harkinian/commit/2010595561e30afe8a1d70ff22747e6ec2eaab3f
Or `gDPSetPrimColor` https://github.com/Archez/2ship2harkinian/commit/e8f1ff4cbd3c7d05279605cf520c2fd574fae25a

Thoughts? (and of course adding `2S2H` style comments)

garrettjoecox:
> @garrettjoecox heres what RGBA8 with bitpacking could look like [Archez@2010595](https://github.com/Archez/2ship2harkinian/commit/2010595561e30afe8a1d70ff22747e6ec2eaab3f) Or `gDPSetPrimColor` [Archez@e8f1ff4](https://github.com/Archez/2ship2harkinian/commit/e8f1ff4cbd3c7d05279605cf520c2fd574fae25a)
> 
> Thoughts? (and of course adding `2S2H` style comments)

I think I prefer either of these over the designators. Are we going to run into this issue anywhere else? Or is this a fairly isolated issue?

Archez:
> I think I prefer either of these over the designators. Are we going to run into this issue anywhere else? Or is this a fairly isolated issue?

This issue only occurs when an initializer list is used with `Color_RGBA8_u32` which these two instances are the only ones in the code base.

Archez:
Updated for the alternate solution

inspectredc (approved):

PR #612, opened 2024-05-30
Fix taking transformation mask off setting unexpected flags

### LINKED ISSUE #490: Removing a transformation mask sets flags that prevents learning opening for Goron Lullaby[Linux]
EDIT: I initially thought this issue was related to the use of Debug mode, but after release additional users reported having this problem. If I can find more details I will edit them into this post.

### Description of the bug:
After messing around and seeing a tip from user Bobafett. It seems on the Linux release builds for 2ship removing a transformation mask and going back to Human Link sets 3 arbitrary flags in the 24th row in weekEventReg.

Two of these set flags relate to the Goron Elder(specifically 4 and 6) and prevents you from learning his song. These flags reset on reverting time so if you manage to free him without reverting back to human form it should work as intended.

Having fast transformations enabled seems to bypass this issue.

1. Run the Linux appimage of 1.0.0 2ship.
2. Put on a transformation mask, then remove it.(This sets 3 arbitrary flags in the 24th row of weekEventReg)
3. Attempt to learn the opening to the Goron Lullaby.

### System Information:
- OS: Garuda Linux Bird of Prey
- GFX Backend: OpenGL
- Game Version: 1.0.0
- Rom Version: NTSC US Gamecube

Archez:
The game sets some flags when putting on a transformation mask. These flags are what allow the player to press a button to "skip" the cutscene after the first time. An array tracks the flag for each player form.

The problem is this logic also executes for taking a mask _off_ which leads to `PLAYER_FORM_HUMAN` being passed into the same array. This ends up as a OOB read, causing garbage data to be passed into `SET_WEEKEVENTREG` meaning random flag(s) are potentially set. It's been observed on LInux that this sets a few flags pertaining to learning Goron Lullaby.

Checking on console reveals that the OOB value that gets used on console is `0x100A` which translates to two flags being set: `WEEKEVENTREG_16_02` and `WEEKEVENTREG_16_08`
The later of the two appears to be unused. The first however is used by the Gorman Brothers to present or skip a specific text ID that would show on the third day if the player completed the alien quest.
The text is the "I head some garos attacked the milk wagon".

I've opted to add a value to the array that matches console, so these flags will be set when taking a mask off, restoring original behavior with the OOB read.

Alternatively, we could fix the bug by checking for the human form and just not call `SET_WEEKEVENTREG` at all if that is more preferred.

Fixes #490 

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1553793249.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1553795755.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1553802221.zip)

### DIFF
--- a/mm/src/overlays/actors/ovl_player_actor/z_player.c
+++ b/mm/src/overlays/actors/ovl_player_actor/z_player.c
@@ -18258,10 +18258,11 @@ void func_80855218(PlayState* play, Player* this, struct_8085D910** arg2) {
 }
 
 u16 D_8085D908[] = {
-    WEEKEVENTREG_30_80, // PLAYER_FORM_FIERCE_DEITY
-    WEEKEVENTREG_30_20, // PLAYER_FORM_GORON
-    WEEKEVENTREG_30_40, // PLAYER_FORM_ZORA
-    WEEKEVENTREG_30_10, // PLAYER_FORM_DEKU
+    WEEKEVENTREG_30_80,               // PLAYER_FORM_FIERCE_DEITY
+    WEEKEVENTREG_30_20,               // PLAYER_FORM_GORON
+    WEEKEVENTREG_30_40,               // PLAYER_FORM_ZORA
+    WEEKEVENTREG_30_10,               // PLAYER_FORM_DEKU
+    PACK_WEEKEVENTREG_FLAG(16, 0x0A), // 2S2H [Port] Added to match OOB value read on console for human form
 };
 struct_8085D910 D_8085D910[] = {
     { 0x10, 0xA, 0x3B, 0x3F },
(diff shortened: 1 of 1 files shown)


garrettjoecox (approved):
wow what a wild ride it was reading that one. Solid find

garrettjoecox (approved):
> Alternatively, we could fix the bug by checking for the human form and just not call SET_WEEKEVENTREG at all if that is more preferred.

I like the fix of preserving hardware behavior, it doesn't cause any user issues.

PR #633, opened 2024-06-01
Item drops 3D!

Captainkittyca2:
Adds a toggle-able feature within `Enhancements->Graphics` to make Item drops 3D and spin. Most of the code is handled within a hook function in `3DItemDrops.cpp` file

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1559082527.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1559082983.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1559085617.zip)


Caladius:
Using kafeis cpp file is a no no lol. You'll need to create your own cpp and h file following the other enhancement examples and self contain your code within it.

Captainkittyca2:
> Using kafeis cpp file is a no no lol. You'll need to create your own cpp and h file following the other enhancement examples and self contain your code within it.

But It's within its own functions. I left the kafei functions untouched

Caladius:
Each enhancement needs it's own separate file.

Captainkittyca2:
> Each enhancement needs it's own separate file.

Alright, I'll make a h and cpp file for it

Captainkittyca2:
Alright, added two files

garrettjoecox (commented):
You should be able to pull this off without touching the actor’s file at all using the actor update & actor init hooks

Captainkittyca2:
> You should be able to pull this off without touching the actor’s file at all using the actor update & actor init hooks

Ah I'm not familiar with those tbh.

garrettjoecox:
Updated to use generic hooks & a few other tweaks

Archez:
I've updated this PR with latest develop and applied some fixes/changes:
* Add rotation reset when disabling the enhancement so that bill-boarded sprites face the camera correctly.
* Refined which items should get the custom spin applied. Some where getting double rotation.
* Huge rupees weren't using the huge rupee model.
* Rupee scales are adjusted to match SoH.
* Removal of the rupee model gfx patch. This was impacting rupees from chests and in theory would not work with custom models.

If we wanted to render just the inner part of these rupees, we should setup new custom GID draw methods that just render the inner rupee part, but I don't think that is required for the scope of this PR.

Archez (approved):
LGTM 👍 Thanks to everyone who worked on this!
A good example of taking something from SoH, but porting it into hooks.

The changes for the slime actor in my opinion are fine as is, as we don't really have a good pattern for pulling something like that out.

mm/src/overlays/actors/ovl_En_Slime/z_en_slime.c
```diff
@@ -1239,11 +1239,29 @@ void EnSlime_Draw(Actor* thisx, PlayState* play) {
...
+                    GetItem_Draw(play, GID_ARROWS_MEDIUM);
+                    break;
+            }
+        } else {
+            Matrix_Scale(0.03f, 0.03f, 0.03f, MTXMODE_APPLY);
+
+            gSPSegment(POLY_OPA_DISP++, 8, (uintptr_t)this->itemDropTex);
+            gSPMatrix(POLY_OPA_DISP++, Matrix_NewMtx(play->state.gfxCtx), G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
+            gSPDisplayList(POLY_OPA_DISP++, gItemDropDL);
+        }
```
inspectredc: Could this be VB hooked? smth like VB_DRAW_SLIME_ITEM(_DROP) ?

Archez: Yeah you're right. I was only thinking about regular actor hooks. I've pushed an update with a VB hook register/unregister.

PR #652, opened 2024-06-03
Add widescreen actor culling option

garrettjoecox:
null

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1563550279.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1563551341.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1563560879.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenuBar.cpp
+++ b/mm/2s2h/BenGui/BenMenuBar.cpp
@@ -540,11 +540,10 @@ void DrawEnhancementsMenu() {
                 { .tooltip =
                       "Disables the distance check for scene geometry, allowing it to be drawn no matter how far "
                       "away it is from the player. This may have unintended side effects." });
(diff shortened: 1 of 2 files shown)


Archez (approved):
After testing through some of the math and trying other formulas, this appears to be the most accurate for >=95% of the cases. Great work!

I think there is potential for also tweaking `actor->uncullZoneScale` in a slight exponential way to account for the field of view stretching at the extreme edges, but I'm not confident in how that value translates to the projection matrix to suggest anything. (and like I side the current proposed changes are good enough in my opinion).

A few actors do their own extra culling checks (bushes, bomb rocks, and other actors that fade out at a distance). These can be addressed in follow up PRs.

PR #658, opened 2024-06-04
[Enhancement] Preseve time speed during cycle reset

balloondude2:
This PR adds an enhancement that will preserve the current time speed through a cycle reset. So you won't need to play the Inverted Song of Time after each reset if you prefer having a slower time speed. 

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1680404202.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1680406934.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1680416535.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Cycle/EndOfCycle.cpp
+++ b/mm/2s2h/Enhancements/Cycle/EndOfCycle.cpp
@@ -7,7 +7,7 @@ extern "C" {
 }
 
 SaveInfo saveInfoCopy;
-s32 timeSpeedOffsetCopy;
+s32 timeSpeedOffsetCopy = 3;
 
 void RegisterEndOfCycleSaveHooks() {
     GameInteractor::Instance->RegisterGameHook<GameInteractor::BeforeEndOfCycleSave>([]() {
--- a/mm/2s2h/Enhancements/Cycle/EndOfCycle.cpp
+++ b/mm/2s2h/Enhancements/Cycle/EndOfCycle.cpp
@@ -7,10 +7,12 @@ extern "C" {
 }
 
 SaveInfo saveInfoCopy;
+s32 timeSpeedOffsetCopy;
 
 void RegisterEndOfCycleSaveHooks() {
     GameInteractor::Instance->RegisterGameHook<GameInteractor::BeforeEndOfCycleSave>(
-        []() { memcpy(&saveInfoCopy, &gSaveContext.save.saveInfo, sizeof(SaveInfo)); });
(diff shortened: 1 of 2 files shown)


garrettjoecox (commented):
Nice! Does this work/conflict with the slower first-cycle mechanic?

balloondude2:
I did a little testing this morning, and it doesn't seem to conflict with the faster first cycle. It doesn't seem that timeSpeedOffset is used to make the first cycle run faster. 

Edit: REG(15) is what sets the first cycle faster. It's set to 5 on the first cycle and 3 on subsequent cycles. REG(15) and timeSpeedOffset together contribute to the time speed in the game.

balloondude2:
This enhancement will also retain a time speed set through the Save Editor over a cycle reset.

mm/2s2h/Enhancements/Cycle/EndOfCycle.cpp
```diff
@@ -7,10 +7,13 @@ extern "C" {
...
 
 SaveInfo saveInfoCopy;
+s32 timeSpeedOffsetCopy;
 
 void RegisterEndOfCycleSaveHooks() {
-    GameInteractor::Instance->RegisterGameHook<GameInteractor::BeforeEndOfCycleSave>(
-        []() { memcpy(&saveInfoCopy, &gSaveContext.save.saveInfo, sizeof(SaveInfo)); });
+    GameInteractor::Instance->RegisterGameHook<GameInteractor::BeforeEndOfCycleSave>([]() {
+        memcpy(&saveInfoCopy, &gSaveContext.save.saveInfo, sizeof(SaveInfo));
+        memcpy(&timeSpeedOffsetCopy, &gSaveContext.save.timeSpeedOffset, sizeof(s32));
```
Archez: Doing a `memcpy` for a primitive value like an integer is overkill. You can just do a `var = value` assignment instead.
```suggestion
        timeSpeedOffsetCopy = gSaveContext.save.timeSpeedOffset;
```

Archez (approved):

inspectredc (approved):
May want an initial value for timeSpeedOffsetCopy for sanity's sake but it doesn't look like it should be necessary

balloondude2:
> May want an initial value for timeSpeedOffsetCopy for sanity's sake but it doesn't look like it should be necessary

I'll get an initial value set. I think I should have time to get it added later today.

balloondude2:
Added an initial value of 0 to match what is set during a vanilla cycle reset.

PR #670, opened 2024-06-06
[Enhancement] Fierce Deity Anchored Backwalk Animation

Eblo:
For whatever reason, Fierce Deity has an anchored backwalk animation that cuts off too early, causing Link to stop and go repeatedly. This fix extends the cutoff, allowing FD to do anchored backwalks smoothly. I wasn't too sure what to categorize this as, but I went with "Restoration" because the animation data *is* in the game, but it is not normally fully utilized here.

Here's a video of it being toggled:

https://github.com/HarbourMasters/2ship2harkinian/assets/7004497/56fc06ed-41b9-494a-b213-5632ee085d4c

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1886804407.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1886808208.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1886813310.zip)


mm/2s2h/Enhancements/Restorations/FierceDeityBackwalk.cpp
```diff
@@ -0,0 +1,14 @@
+#include <libultraship/bridge.h>
+#include "Enhancements/GameInteractor/GameInteractor.h"
+
+void RegisterFierceDeityBackwalk() {
+    REGISTER_VB_SHOULD(GI_VB_ANCHORED_BACKWALK, {
+        Player* player = (Player*)opt;
+        // If the player is Fierce Deity, the anchored backwalk animation has a longer duration.
+        if (CVarGetInteger("gEnhancements.Restorations.FierceDeityBackwalk", 0) &&
+            player->transformation == PLAYER_FORM_FIERCE_DEITY) {
+            f32 currentFrame = player->skelAnime.curFrame;
```
inspectredc:
```suggestion
```

Doesn't look like you ended up using this

inspectredc (approved):
Just a small cleanup but LGTM

mm/2s2h/Enhancements/Restorations/FierceDeityBackwalk.cpp
Eblo: Ah yep, nice catch. That's a leftover from when I wasn't doing hooks correctly. Thanks!

mm/2s2h/Enhancements/Restorations/FierceDeityBackwalk.cpp
```diff
@@ -0,0 +1,13 @@
+#include <libultraship/bridge.h>
+#include "Enhancements/GameInteractor/GameInteractor.h"
+
+void RegisterFierceDeityBackwalk() {
+    REGISTER_VB_SHOULD(GI_VB_ANCHORED_BACKWALK, {
+        Player* player = (Player*)opt;
+        // If the player is Fierce Deity, the anchored backwalk animation has a longer duration.
+        if (CVarGetInteger("gEnhancements.Restorations.FierceDeityBackwalk", 0) &&
+            player->transformation == PLAYER_FORM_FIERCE_DEITY) {
+            *should = player->skelAnime.curFrame > 9.0f;
```
Archez: Could you explain how you came up with `9.0` as the value here?

Archez (approved):
Looks good. Will need to bring in latest develop/conflict resolution before merge.
Left a non-blocking question.

mm/2s2h/Enhancements/Restorations/FierceDeityBackwalk.cpp
Eblo: As best as I can tell, all other forms in the game use a sort of skewed backwalk when Z-targeting something. Fierce Deity is different and uses the direct backwalk animation even when anchored to a target. My suspicion is that the `Player_Action_15` function assumes the frame count for the anchored backwalk animation, which is shorter than the one for the regular backwalking that FD uses here.

I had looked at the frame data in `link_animetion.xml`. The normal backwalk has a frame count of 29, but the anchored backwalk has a frame count of 16:

https://github.com/HarbourMasters/2ship2harkinian/blob/136a928883ac4ef3efc5bba48b48a047fc1a1de0/mm/assets/xml/N64_US/misc/link_animetion.xml#L480
https://github.com/HarbourMasters/2ship2harkinian/blob/136a928883ac4ef3efc5bba48b48a047fc1a1de0/mm/assets/xml/N64_US/misc/link_animetion.xml#L515

While I don't know exactly what the existing `5.0f` in the code is (looping info?), I used it with the above frame data to do some admittedly arbitrary math:

```
(29 / 16) * 5.0 = 9.0625
```

Rounded to 9.0, this value worked pretty well out of the box. Further experimentation showed that lower whole number values will still run into the stop-and-restart issue found with the vanilla logic. Higher values will let Link continue to move along, but the last frame of the animation will freeze for a bit before looping again. 9.0 is a sweet spot.

Eblo:
Apologies for the late response. I have rebased off of develop.

Archez:
Thank you for the detailed explanation! I was very curious about all this and wanted to confirm things better so that it didn't feel "too arbitrary", so I did some digging into the action funcs.

The first thing I did was log out which action func is executed every frame, and what function is setting a new action. What I found is that only the FD transformation was setting `Action_15` when backwalking, all the other transformations remain in `Action_9` which appears to be the main z-target follow action. This lead me to think that its not actually related to the animation itself.

`Action_15` is set because of a result from `func_8083E404`, which appears to do so math around the players current speed and yaw rotation. It seems like whats happening is that FD's speed is just too high and triggers a return value that changes the action. To confirm this suspicion, I hardcoded the incoming arg to be FD's value for all transformations, and then I was able to see the same "incomplete" backwalk for human Link, etc.

https://github.com/user-attachments/assets/e49b316f-9a31-4045-bf1a-9f9474c373f7

Conversely, clamping FD's speed arg to match the other transformations completely resolves his z-target movement, and `Action_15` is never called. This can be noticed by FD link continuing to hold his sword with two hands the entire time, and no sudden jerky movements when going left/right. (My video is using keyboard controls, so it might not look as smooth as controller stick movement).

https://github.com/user-attachments/assets/59595262-4e0d-4159-bdb5-30a030483c39

---

Given that `Action_15` on its own is behaving identically across all transformations, I don't think we will want to change the animation frame check anymore.

Would you like me to submit these changes to update your branch/PR, or just open a new PR for the different approach instead?

Eblo:
I appreciate the added research! So ultimately, these changes make FD actually behave like the other forms, correct? If so, I'm more than fine with using your approach for consistency's sake. I have no preference for whether you submit to my branch or open a PR. Whichever seems cleaner/easier to you.

Archez:
> So ultimately, these changes make FD actually behave like the other forms, correct?

Correct, with regards to this z-target related function. The speed clamping only happens here as part of the math. FD's actual movement speed is unaffected, and remains _fast_.

I'll go ahead and push these changes to your branch, then this can be merged in. And based on the new approach, I think I'll also move the toggle to the "Fixes" menu.

Eblo:
I tried your change out for myself; it plays nicely in my opinion. I'm more than happy with this approach over what I initially had.

Archez:
> I tried your change out for myself; it plays nicely in my opinion. I'm more than happy with this approach over what I initially had.

Good to hear!

PR #679, opened 2024-06-09
[Enhancement] Two-handed sword spin attacks

Eblo:
This establishes parity with the magic spin attack for the Fierce Deity sword and the Great Fairy's Sword. Do note that I had to do some refactoring in `z_player.c` to make the GFS work, as there are various points where only the B button was checked for input, while the GFS is a C-button item.

https://github.com/HarbourMasters/2ship2harkinian/assets/7004497/cc63727c-38cd-4009-9a71-52045088b3d2

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2041890059.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2041903907.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2041944009.zip)


smashbro596:
this is godlike! this was one of my big asks! have fd link be able to use the spin attack!

mm/src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -8185,14 +8187,33 @@ s32 func_8083A4A4(Player* this, f32* arg1, s16* arg2, f32 arg3) {
...
+    uint16_t buttonToCheck;
+    if (IS_HELD_DPAD(this->heldItemButton)) {
+        buttonToCheck = sDpadItemButtons[HELD_ITEM_TO_DPAD(this->heldItemButton)];
+    } else {
+        buttonToCheck = sPlayerItemButtons[this->heldItemButton];
+    }
+
+    return CHECK_BTN_ALL(sPlayerControlInput->cur.button, buttonToCheck);
+}
+// #endregion
```
garrettjoecox: It looks like this function is only ever called from your enhancement file in C++, can it just be moved there?

mm/src/overlays/actors/ovl_En_M_Thunder/z_en_m_thunder.c
```diff
@@ -547,7 +547,15 @@ void EnMThunder_Draw(Actor* thisx, PlayState* play2) {
         Matrix_Translate(0.0f, 220.0f, 0.0f, MTXMODE_APPLY);
         Matrix_Scale(-1.2f, -0.8f, -0.6f, MTXMODE_APPLY);
         Matrix_RotateXS(0x4000, MTXMODE_APPLY);
-    } else {
+    }
+    // #region 2S2H [Enhancements] Two Handed Sword Magic Charge Effect
```
garrettjoecox: Seems harmless enough, it's rather difficult doing stuff like this from hooks. Let's wrap it in a cvar at least though?

mm/2s2h/Enhancements/GameInteractor/GameInteractor.h
```diff
@@ -55,6 +55,8 @@ typedef enum {
     GI_VB_FLIP_HOP_VARIABLE,
     GI_VB_DISABLE_LETTERBOX,
     GI_VB_CLOCK_TOWER_OPENING_CONSIDER_THIS_FIRST_CYCLE,
+    GI_VB_HELD_ITEM_BUTTON_PRESS,
+    GI_VB_MAGIC_SPIN_ATTACK_FORM_CHECK,
```
garrettjoecox: Both of these could be a bit better, should read as a question (Should held item button press?) maybe "Should held item consider button pressed?" Bit of a mouthful but it's clearer

For the other you can just swap the last two words "Should magic spin attack check form?"

garrettjoecox (approved):
a few comments but mostly looks good, needs some updating

mm/src/overlays/actors/ovl_player_actor/z_player.c
Eblo: Sure, I'm not against that. In my head I thought this might be something others may want to use so I left it in z_player.c, but since there is nobody calling for it now, that's silly. I'll move it.

mm/src/overlays/actors/ovl_En_M_Thunder/z_en_m_thunder.c
Eblo: It actually wasn't too troublesome to hookify this. There exists a visual bug in the original game where the player could get the GFS to do a charged magic spin attack, but it would display the Kokiri Sword's visual effect. The way I did this enhancement previously had fixed that effect regardless of whether the enhancement was enabled or disabled. With the new hook, the effect is only fixed with the enhancement on, so purists can still see the original bug if the enhancement is disabled.

mm/2s2h/Enhancements/GameInteractor/GameInteractor.h
Eblo: I renamed `GI_VB_HELD_ITEM_BUTTON_PRESS` to `VB_CHECK_HELD_ITEM_BUTTON_PRESS`. Vanilla behavior is to *only* check B, regardless of what button the item is actually assigned to. Incidentally, this means vanilla behavior is that the GFS can do a charged magic spin attack if the player quickly holds B after swinging the GFS. What this hook does is instead check the currently held item's assigned button, which may be B or some other button. From what I tested, that also means that, not only does this enhancement render the GFS magic charge glitch pointless, but it is also not possible when enabled. The glitch can still be performed when the enhancement disabled.

Eblo:
I addressed the feedback. For players, the main change of note is that the new two-handed sword magic charge visual effect will only display when this enhancement is enabled.

garrettjoecox (approved):

PR #682, opened 2024-06-10
[Enhancement] Make sword beams work on any enemy

Eblo:
This extends the Fierce Deity Anywhere enhancement to make the sword beams work for regular enemies. For a given enemy, the sword beam multiplier is set to 1 if it is 0. The effect will match light arrows if none are already set. Additionally, sword beams will always trigger a collision, which makes them work on enemies with special properties, such as Skulltulas and Big Octos. Bosses should retain their existing behavior for sword beams, such as the splash effect when Odolwa blocks sword beams.

https://github.com/HarbourMasters/2ship2harkinian/assets/7004497/431c1f66-95ce-48df-a284-bdeb0e31fa07

Known enemies that still ignore sword beams:
- **Snapper**
- **Deep Python**, which I don't think is possible for FD to even encounter?

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2240573645.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2240578201.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2240594527.zip)


smashbro596:
YES! now to wait for this and the two handed sword attacks to be merged into the main nightly branch and i'll be ready to cut the moon in half!

all enemies found in termina field work beautifully. ill test in other areas.

enemies the beams go right through:
big octos
big skulltulas (when facing you. they work when their backs are turned)
leevers? *(they won't sit still long enough for me to hit them. but better safe then sorry)

inspectredc:
Hey! Awesome work so far on this! Just wanted to note that the intention of using hooks is to have a single enhancement be able to subscribe to multiple hooks rather than a hook being used multiple times for a single enhancement. In this case the idea would be to instead create a new VB hook where you have used the FD anywhere one, and then register that hook similarly inside the same RegisterFierceDeityAnywhere function. If you have any more questions about this or need any help lmk in the development channel!

Eblo:
> big skulltulas

In this case, skulltulas only take damage in their exposed state, like they would with other sword attacks.

> leevers? *(they won't sit still long enough for me to hit them. but better safe then sorry)

They *seem* to work from my own testing, but I've only managed to get a sword beam to target a leever when the leever is so close that it's hard to say it wasn't the sword itself that connected.

> In this case the idea would be to instead create a new VB hook where you have used the FD anywhere one, and then register that hook similarly inside the same RegisterFierceDeityAnywhere function

Are you recommending to do something similar to the [clock tower cutscene skip hook](https://github.com/HarbourMasters/2ship2harkinian/blob/develop/mm/2s2h/Enhancements/Cutscenes/StoryCutscenes/SkipClockTowerOpen.cpp)?

smashbro596:
> > big skulltulas
> 
> In this case, skulltulas only take damage in their exposed state, like they would with other sword attacks.
> 
> > leevers? *(they won't sit still long enough for me to hit them. but better safe then sorry)
> 
> They _seem_ to work from my own testing, but I've only managed to get a sword beam to target a leever when the leever is so close that it's hard to say it wasn't the sword itself that connected.
> 
> > In this case the idea would be to instead create a new VB hook where you have used the FD anywhere one, and then register that hook similarly inside the same RegisterFierceDeityAnywhere function
> 
> Are you recommending to do something similar to the [clock tower cutscene skip hook](https://github.com/HarbourMasters/2ship2harkinian/blob/develop/mm/2s2h/Enhancements/Cutscenes/StoryCutscenes/SkipClockTowerOpen.cpp)?

i noticed the big skulltulas are killed when hit by a beam to the back. but i feel like they should react like getting hit by an arrow. yknow because it passes through.
interestingly the bubble enemies die whether on fire or not when struck by a sword beam.

Eblo:
The Skulltulas, Big Octos, and They actually seem to be fixable by ORing the sword beam damage effect with the arrow damage effect when setting the sword beam damage flags. So far, this seems to keep boss sword beam effects intact while maintaining the damaging effect on enemies. However a comment specifically recommends not combining these flags for the time being...

https://github.com/HarbourMasters/2ship2harkinian/blob/e60ca15ae6a9e5381c9675530828e7a1c246f643/mm/include/z64collision_check.h#L397-L398

inspectredc:
> Are you recommending to do something similar to the [clock tower cutscene skip hook](https://github.com/HarbourMasters/2ship2harkinian/blob/develop/mm/2s2h/Enhancements/Cutscenes/StoryCutscenes/SkipClockTowerOpen.cpp)?

Yes that should be a good example to follow

mm/src/overlays/actors/ovl_En_M_Thunder/z_en_m_thunder.c
```diff
@@ -174,6 +175,11 @@ void EnMThunder_Init(Actor* thisx, PlayState* play) {
             this->timer = 1;
             this->scaleTarget = 12;
             this->collider.info.toucher.dmgFlags = DMG_SWORD_BEAM;
+            // #region 2S2H - Enhancements - Give sword beams arrow damage effect
+            if (!GameInteractor_Should(GI_VB_SWORD_BEAMS_ON_REGULAR_ENEMIES, true, NULL)) {
+                this->collider.info.toucher.dmgFlags |= DMG_NORMAL_ARROW;
+            }
+            // #endregion
```
inspectredc: For things like this you should be able to use the OnActorInit Hook to then add the DMG_NORMAL_ARROW flag there

mm/src/code/z_collision_check.c
```diff
@@ -59,8 +60,35 @@ f32 CollisionCheck_GetDamageAndEffectOnBumper(Collider* at, ColliderInfo* atInfo
...
+                 */
+                u8 spinAttackEffect = (ac->actor->colChkInfo.damageTable->attack[24] >> 4) & 0xF;
+                u8 swordEffect = (ac->actor->colChkInfo.damageTable->attack[9] >> 4) & 0xF;
+                *effect = spinAttackEffect == 0 ? swordEffect : spinAttackEffect;
+            } else {
+                /**
+                 * An effect for sword beams has been defined, so just use that.
+                 */
+                *effect = defaultEffect;
+            }
```
inspectredc: This is code that is ideally handled in a function in the FierceDeityAnywhere file. You can call a function with pointers to `damage` and the `effect` ptr to affect the values and pass `i` (or just use 25) as well.

inspectredc: I think also instead of doing 

`if ((!GameInteractor_Should(GI_VB_SWORD_BEAMS_ON_REGULAR_ENEMIES, true, NULL) && i == 25)) {`

you should try 

`if ((!GameInteractor_Should(GI_VB_SWORD_BEAMS_ON_REGULAR_ENEMIES, true, &i))) {`

and only set `*should = false` when `i` is 25
```cpp
if (*opt == 25) {
    *should = false;
}
```

inspectredc: also if there is any enum to represent what 25 would mean here please use that! if not please explain in a comment maybe what the 25 means

smashbro596:
i wonder. could the check be applied to boulders? my thought is: imagine blowing up rocks with your blade beam. or melt ice? or destroy sun blocks?

Eblo:
> i wonder. could the check be applied to boulders? my thought is: imagine blowing up rocks with your blade beam. or melt ice? or destroy sun blocks?

Now that I have a better understanding of how hooks should be applied, that's probably an easy thing to add. I'm not sure if that should be a standard feature as part of this, though. The idea of this is to make the sword beam behavior for normal enemies similar to how it works on bosses.

I applied the normal arrow effect to sword beams so that they could kill big octos and non-prone skulltulas. A side effect of this is that they can now break the big icicle on Termina Field in 2 hits. If someone else wanted to, they could add in whatever effects they please to sword beams, like the ones you listed.

Eblo:
I just pushed a new update that makes enemies display the blue light orbs effect upon taking sword beam damage. I also completely overhauled the collision detection of sword beams. Instead of adding the light arrow damage flag, it will instead always register a collision. This keeps the one odd effect from Odolwa intact while also letting the beams connect with Skulltulas and Big Octos.

Various hurdles, as commented in `FierceDeityAnywhere.cpp` required having some sort of per-actor variable to use. I went with the actor's ActorShape face, since that is something that exists for all actors but is not normally used for anything but the player. This shouldn't interfere with how anything else in the game behaves, unless other enhancements decide to use the same thing for other purposes.

Eblo:
I updated this to use the new variadic argument syntax for the hooks. In particular this has cleaned up the damage multiplier and damage effect hooks, resulting in one less file modified for this feature.

mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
```diff
@@ -1,10 +1,142 @@
...
+    REGISTER_VB_SHOULD(GI_VB_DRAW_DAMAGE_EFFECT, {
+        if (CVarGetInteger("gEnhancements.Masks.FierceDeitysAnywhere", 0)) {
+            if (lightArrowsAsSwordBeams) {
+                u8* type = va_arg(args, u8*);
+                if (*type == ACTOR_DRAW_DMGEFF_LIGHT_ORBS) {
+                    *type = ACTOR_DRAW_DMGEFF_BLUE_LIGHT_ORBS;
+                }
+            }
+        }
+    });
```
garrettjoecox: Am I missing something, or can you just pass the Actor* that's being damaged into the `GI_VB_DRAW_DAMAGE_EFFECT` hook and check their `actor->shape.face` here, instead of doing the `ShouldActorDraw` and storing it in a static variable. 

Also the way it is now seems like it could have issues when you're talking about multiple actors taking different kinds of damage at the same time.

mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
```diff
@@ -1,10 +1,142 @@
...
+    /*
+     * If this is a sword beam collision, just hand wave it as a valid collision. This allows for sword beams to hit
+     * enemies in a damaging way, such as Skulltulas and Big Octos. The isSwordBeamCollision flag exists for the sole
+     * purpose of having the blue light orbs effect on Big Octos; that actor handles damage effects differently from
+     * most other enemies.
+     */
+    REGISTER_VB_SHOULD(GI_VB_CHECK_BUMPER_COLLISION, {
+        if (CVarGetInteger("gEnhancements.Masks.FierceDeitysAnywhere", 0)) {
+            ColliderInfo* bumper = va_arg(args, ColliderInfo*);
+            *should = isSwordBeamCollision = bumper->bumper.dmgFlags & DMG_SWORD_BEAM;
```
garrettjoecox: It looks like when you have `FierceDeitysAnywhere` enabled, you are overriding _all_ `CollisionCheck_NoSharedFlags` collision checks with this, is that intended? Is the sword beam the only usage of `CollisionCheck_NoSharedFlags`?

mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
Eblo: > Am I missing something, or can you just pass the Actor* that's being damaged into the GI_VB_DRAW_DAMAGE_EFFECT hook and check their actor->shape.face here, instead of doing the ShouldActorDraw and storing it in a static variable.

As stated in the comment, some calls to `Actor_DrawDamageEffects` pass in `NULL` for the actor, such as Leevers:

https://github.com/HarbourMasters/2ship2harkinian/blob/fa702ed90c3d28093dfb95cb9069018e4fa03a48/mm/src/overlays/actors/ovl_En_Neo_Reeba/z_en_neo_reeba.c#L659-L660

Not having this results in the converted sword beam visual effect reverting to the one for light arrows for those actors:

https://github.com/user-attachments/assets/0c4187be-e8c4-4772-ba2a-59bdd2e7ce1b

> Also the way it is now seems like it could have issues when you're talking about multiple actors taking different kinds of damage at the same time.

You're right. If a sword beam hits one actor, but another actor gets hit by something else, it might cancel out this static variable. This variable only exists for edge cases like the Leever. I argued with myself when reminding myself of what I did for those, and in the process I came up with some other ideas I want to try. I'll get back to you.

mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
Eblo: Not intended, not the only usage, big goof. Thanks for the catch.

mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
Eblo: I have streamlined it so there is only a single static variable to store the actor currently being drawn. I'd love if I didn't have to use one at all, but I don't know of another way to handle null cases like with Leevers. The new approach should hopefully minimize/eliminate potential weirdness with actors' collision interfering.

mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
Eblo: Thanks to varadiac args, I changed it so that the collision check only applies if the toucher has the sword beams flag.

Eblo:
Addressed feedback. Collision should have less potential weirdness now.

Archez (approved):
I love how the work for this has evolved over time and adapted to the new variadic args pattern! I feel this is in a really good state now with minimal src footprint.

My only additional feedback would be that I think it would be good to have this behavior behind its own toggle, but that work can be saved for a follow up PR around all the FD toggles (as we discussed in Discord).

mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
```diff
@@ -1,10 +1,139 @@
...
+                actor->shape.face = NOT_HIT_BY_SWORD_BEAM;
+            }
+        }
+    });
+
+    /*
+     * Use a hook before an actor draws to store which actor is being drawn. Some actors call Actor_DrawDamageEffects()
+     * with a null actor (e.g. leevers), so we cannot just check the actor in the draw damage hook directly.
+     */
+    GameInteractor::Instance->RegisterGameHook<GameInteractor::ShouldActorDraw>(
```
garrettjoecox: Still don't love this, but I will feel much better about it with the conditional hooks pattern.

garrettjoecox (approved):
noice

mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
```diff
@@ -1,10 +1,139 @@
...
+        [](Actor* actor, bool* result) { currentlyDrawnActor = actor; });
+
+    /*
+     * If we're drawing the light arrow damage effect, but we know it's from a sword beam, then quietly change the type
+     * to the blue lights effect.
+     */
+    REGISTER_VB_SHOULD(VB_DRAW_DAMAGE_EFFECT, {
+        if (CVarGetInteger("gEnhancements.Masks.FierceDeitysAnywhere", 0)) {
+            if (currentlyDrawnActor->shape.face & HIT_BY_SWORD_BEAM) {
```
inspectredc: Can the actor be passed in here instead of setting up the static ptr?

Eblo: Some actors call `Actor_DrawDamageEffects` with NULL for the actor, such as Keese and Leevers. I would absolutely love to avoid having to use this static variable, but I don't know of any other way to account for those cases. I am all ears for any ideas.

Eblo:
Since only three actors (Keese, Freezards, Leevers) seem to have this unusual null case for drawing damage effects, I decided to specially handle those cases for drawing the sword beam damage effect. I added a new hook for those calls. In this hook, I check if it's a sword beam collision, and if it is one, it will instead call `Actor_DrawDamageEffects` with the actor instead of null. This allows the visual effect to work without need for a static variable or the `ShouldActorDraw` hook.

mm/src/overlays/actors/ovl_En_Firefly/z_en_firefly.c
```diff
@@ -844,9 +845,11 @@ void EnFirefly_Draw(Actor* thisx, PlayState* play) {
         POLY_OPA_DISP = gfx;
     }
 
-    Actor_DrawDamageEffects(play, NULL, this->bodyPartsPos, KEESE_BODYPART_MAX,
-                            this->drawDmgEffScale * this->actor.scale.y * 200.0f, this->drawDmgEffFrozenSteamScale,
-                            this->drawDmgEffAlpha, this->drawDmgEffType);
+    if (GameInteractor_Should(VB_USE_NULL_FOR_DRAW_DAMAGE_EFFECTS, true, this)) {
```
Archez: Because the maximum possible variadic args for this VB_ enum is 2, I'd recommend at least passing `NULL` for the two cases that don't currently. (or you could also pass the bodyParts for all 3 uses for consistency, this might be better even if redundant).

(but also see my next comment)

mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
```diff
@@ -79,19 +77,47 @@ void RegisterFierceDeityAnywhere() {
...
+                if (actor->id == ACTOR_EN_FIREFLY) { // Keese
+                    EnFirefly* enFireFly = (EnFirefly*)actor;
+                    Actor_DrawDamageEffects(gPlayState, actor, enFireFly->bodyPartsPos, KEESE_BODYPART_MAX,
+                                            enFireFly->drawDmgEffScale * actor->scale.y * 200.0f,
+                                            enFireFly->drawDmgEffFrozenSteamScale, enFireFly->drawDmgEffAlpha,
+                                            enFireFly->drawDmgEffType);
+                } else if (actor->id == ACTOR_EN_FZ) { // Freezard
+                    EnFz* enFz = (EnFz*)actor;
+                    Vec3f* bodyPartsPos = va_arg(args, Vec3f*);
+                    Actor_DrawDamageEffects(gPlayState, actor, bodyPartsPos, ARRAY_COUNT(bodyPartsPos),
```
Archez: I don't think you can do this correctly here `ARRAY_COUNT(bodyPartsPos)`, this relies on the compiler knowing the declared size of the array, which it no longer knows now that the type is a `Vec3f*` in this scope. This most likely is returning some number that is incorrect.

You will probably need to pass in the arg count as a va_arg.

This is why `Actor_DrawDamageEffects` itself requires passing the count separately from the array pointer as well.

mm/src/overlays/actors/ovl_En_Firefly/z_en_firefly.c
Eblo: Done, now passing in the Vec3f and its size for all three.

mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
Eblo: Good catch, and thanks for the info. Fixed.

PR #695, opened 2024-06-15
Add Manual Jump from Patcher64+

Quorsor:
Adds the Manual Jump option from Patcher64+

Z + A to Jump manually
B while midair to Jump Attack

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1650506228.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1650507299.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1650513917.zip)


garrettjoecox (dismissed):
Nice work! You are part of the way there with your VB hook, the idea is that the only change you make to the source code is adding your `GameInteractor_Should(` to the right spots, and the additional behavior you are introducing should be handled in the hook handler itself (ManualJump.cpp in your case)

garrettjoecox (approved):
Updated this PR to conform to the newer patterns, moved it to the new menu, and QA'd

PR #700, opened 2024-06-18
[Cheat] Stop time in Temples and Dungeons

balloondude2:
This PR will add a cheat that will stop time while in temples. Toggling the cheat will require a room change to take effect. 
The clock will also be hidden while in the temple.

EDIT: Cheat has 3 settings: 
-  Off
- Temples (Woodfall Temple, Snowhead Temple, Great Bay Temple, and Stone Tower Temple)
- Temples and Dungeons (includes the four temples, both spider houses, Pirates' Fortress, Beneath the Well, Ikana Castle, and Secret Shrine)

KNOWN ISSUES:

- [x] Doesn't work in Igos's throne room, despite working in the entryway.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1748423500.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1748427181.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenuBar.cpp
+++ b/mm/2s2h/BenGui/BenMenuBar.cpp
@@ -664,8 +664,9 @@ void DrawCheatsMenu() {
                                     { .tooltip = "Holding L makes you float into the air" })) {
             RegisterMoonJumpOnL();
         }
-        UIWidgets::CVarCheckbox("Stop Time in Temples", "gCheats.TempleTimeStop",
-                                { .tooltip = "Time will not advance while in Temples" });
+        UIWidgets::CVarCheckbox(
+            "Stop Time in Temples", "gCheats.TempleTimeStop",
+            { .tooltip = "Time will not advance while in Temples. Requires a scene change to update." });
 
         ImGui::EndMenu();
     }
--- a/mm/2s2h/BenGui/BenMenuBar.cpp
+++ b/mm/2s2h/BenGui/BenMenuBar.cpp
@@ -664,6 +664,8 @@ void DrawCheatsMenu() {
                                     { .tooltip = "Holding L makes you float into the air" })) {
             RegisterMoonJumpOnL();
         }
+        UIWidgets::CVarCheckbox("Stop Time in Temples", "gCheats.TempleTimeStop",
+                                { .tooltip = "Time will not advance while in Temples" });
 
         ImGui::EndMenu();
     }
--- a/mm/2s2h/BenGui/BenMenuBar.cpp
+++ b/mm/2s2h/BenGui/BenMenuBar.cpp
@@ -60,9 +60,8 @@ static const std::unordered_map<int32_t, const char*> alwaysWinDoggyraceOptions 
 
 static const std::unordered_map<int32_t, const char*> timeStopOptions = {
     { TIME_STOP_OFF, "Off" },
-    { TIME_STOP_TEMPLES, "Time stops in the 4 main temples. Requires a scene update to take effect." },
-    { TIME_STOP_TEMPLES_DUNGEONS,
-      "Time stops in the 4 main temples and the mini-dungeons. Requires a scene update to take effect." },
+    { TIME_STOP_TEMPLES, "In the 4 main temples. Requires a room change to update." },
+    { TIME_STOP_TEMPLES_DUNGEONS, "In the 4 main temples and mini-dungeons. Requires a room change to update." },
 };
 
 namespace BenGui {
--- a/mm/2s2h/BenGui/BenMenuBar.cpp
+++ b/mm/2s2h/BenGui/BenMenuBar.cpp
@@ -58,6 +58,13 @@ static const std::unordered_map<int32_t, const char*> alwaysWinDoggyraceOptions 
     { ALWAYS_WIN_DOGGY_RACE_ALWAYS, "Always" },
 };
 
+static const std::unordered_map<int32_t, const char*> timeStopOptions = {
+    { TIME_STOP_OFF, "Off" },
+    { TIME_STOP_TEMPLES, "Time stops in the 4 main temples. Requires a scene update to take effect." },
+    { TIME_STOP_TEMPLES_DUNGEONS,
+      "Time stops in the 4 main temples and the mini-dungeons. Requires a scene update to take effect." },
+};
+
 namespace BenGui {
 
 void DrawMenuBarIcon() {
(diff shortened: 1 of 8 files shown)


SuperBro64:
This is a really cool idea. Could make the dungeon crawling less stressful for some.

I actually made a discussion post about a cheat that would stop time in general everywhere, like in SoH (https://github.com/HarbourMasters/2ship2harkinian/discussions/673). Perhaps this could also be a dropdown, where the options are:
- Don't Stop Time (default)
- Stop Time in Temples Only
- Stop Time Everywhere

Just a suggestion, unless you intended this cheat to be distinct from stopping time everywhere. Still great work!

balloondude2:
I thought about having an option to stop time everywhere, but since some actors/npcs operate on a schedule, I thought it might not be worth adding. 

I could add it if people would prefer having the option anyway.

Archez:
In my opinion, I don't think that should be an option. Dungeons works because no actor inside a dungeon has a time schedule. Freezing time as a whole is bound to just create support issues and requires players to have to turn it on and off repeatedly in a playthrough.

There is an option available to freeze time via the save editor, which is a debug tool, so the expectation is that it is provided as-is for testing/debugging. Not really intended for casual playthroughs.

SuperBro64:
> I thought about having an option to stop time everywhere, but since some actors/npcs operate on a schedule, I thought it might not be worth adding.

Oh, wow. I completely forgot that there are NPCs whose movements depend on time moving forward. That does throw a wrinkle into stopping time everywhere...

In that case, as an alternate suggestion, rather than the third option of the dropdown I proposed stopping time everywhere, it could simply extend your base cheat to stopping time in the mini-dungeons, as well. Meaning not just the four temples but also:
- Swamp Spider House
- Oceanside Spider House
- Pirates' Fortress
- Beneath the Well
- Ancient Castle of Ikana
- Secret Shrine

Correct me if I'm wrong, but as far as I can remember, those areas don't have any time-based actor/NPC movements. (The man who comes into the Oceanside Spider House after clearing it does give you a different reward depending on the current day, but he doesn't move, so I think it would probably be fine?)

SuperBro64:
The additions you made for stopping time in the mini-dungeons look really good! Just one small thing. You have the code for Pirates' Fortress as:

```
// Pirates' Fortress
case SCENE_KAIZOKU:
```

But according to [MM's scene table](https://wiki.cloudmodding.com/mm/Scene_Table/Internal/U_1.0), I think "KAIZOKU" only refers to the initial entrance area of the fortress where the moat is. "TORIDE" refers to the main exterior scene of the fortress, while "PIRATE" refers to the interior areas, so perhaps those should be added, too?

Archez:
> Doesn't work in Igos's throne room, despite working in the entryway

If you weren't aware, this is actually happening in all dungeons/temples when going through doors or respawning. This is because `R_TIME_SPEED` is (re)set on room load, not on scene init.

You would need to switch your logic to use the `OnRoomInit` hook, but currently it executes before the scene commands execute. You could move the hook execution to inside `OTRfunc_800973FC`, before the inner `return 1`  to resolve that.

balloondude2:
> > Doesn't work in Igos's throne room, despite working in the entryway
> 
> If you weren't aware, this is actually happening in all dungeons/temples when going through doors or respawning. This is because `R_TIME_SPEED` is (re)set on room load, not on scene init.
> 
> You would need to switch your logic to use the `OnRoomInit` hook, but currently it executes before the scene commands execute. You could move the hook execution to inside `OTRfunc_800973FC`, before the inner `return 1` to resolve that.

I could have sworn that it was working through room changes before, but I haven't done extensive testing since noticing the issue with Igos. I had tried using `OnRoomInit` but had hit a roadblock, so thanks for pointing me to `OTRfunc_800973FC`. I'll see if I can get it actually working

mm/2s2h/Enhancements/Cheats/TimeStop.cpp
```diff
@@ -0,0 +1,64 @@
...
+                    // Ancient Castle of Ikana + Igos's Lair
+                    case SCENE_CASTLE:
+                    case SCENE_IKNINSIDE:
+                    // Secret Shrine
+                    case SCENE_RANDOM:
+                        R_TIME_SPEED = 0;
+                        break;
+                    default:
+                        break;
+                }
```
Archez, line 36: With the double switch statements and large blocks, it would be great to indicate that you are intentionally falling through with a comment.

```suggestion
                }
                // fallthrough
```

Archez (approved):
Looks good! Thanks for adding the extra hook 🚀

mm/2s2h/BenGui/BenMenuBar.cpp
```diff
@@ -58,6 +58,12 @@ static const std::unordered_map<int32_t, const char*> alwaysWinDoggyraceOptions
     { ALWAYS_WIN_DOGGY_RACE_ALWAYS, "Always" },
 };
 
+static const std::unordered_map<int32_t, const char*> timeStopOptions = {
+    { TIME_STOP_OFF, "Off" },
+    { TIME_STOP_TEMPLES, "In the 4 main temples. Requires a room change to update." },
```
Archez: These values are what get used directly in the drop down. I recommend using short names here as they are currently overflowing input when closed. The longer descriptions can be moved to a tooltip that gets set on the combo box widget. You can refer to the "Debug Save File Mode" for an example of what I mean.

Maybe `Off / Temples / Temples + Mini Dungeons`

mm/2s2h/Enhancements/Cheats/TimeStop.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "Enhancements/Enhancements.h"
+
+extern "C" {
+#include <variables.h>
+#include <functions.h>
+}
+
+void RegisterTimeStopInTemples() {
+    GameInteractor::Instance->RegisterGameHook<GameInteractor::AfterRoomSceneCommands>([](s16 sceneId, s8 roomNum) {
+        uint8_t selectedOption = CVarGetInteger("gCheats.TempleTimeStop", 0);
```
Archez:
```suggestion
        uint8_t selectedOption = CVarGetInteger("gCheats.TempleTimeStop", TIME_STOP_OFF);
```

Archez (commented):
Hey, just noticed an opportunity to improve the messaging of the UI element. Left a suggestion below. Let me know if that is something you'd be able to address.

balloondude2:
> Hey, just noticed an opportunity to improve the messaging of the UI element. Left a suggestion below. Let me know if that is something you'd be able to address.

Great suggestion. I got it updated and it looks a lot better. Let me know if you want any of the wording changed.

PR #701, opened 2024-06-20
Modern menu

Malkierian:
This adds a new, more modern-styled menu to use instead of the menubar. The goal was to have it extensible, easier to navigate, and capable of holding a lot more enhancements. I think I accomplished that.

- 3 buttons on the menu header for touch use (when porting to mobile platforms, for instance) for closing the menu, restarting the game, and quitting the program.
- Virtually full size flexibility. Designed around 1280x800, but still functional down to around 400x200.
- All normally extra windows start docked in the menu (for ease of use), but can be popped out to their normal windows.
- The menu itself can be popped out instead of being an overlay, and as such can be docked (could use some refinement, though).
- [Esc] is now the key used to bring it up instead of F1.
- Menu retains its section and subsection selections, meaning you can open and close the menu to toggle an option quickly.
- Menu retains its popped size and position when popped back in.

This also reimplements the gamepad navigation toggle from SoH (which works better in the new menu than on the menubar by several orders of magnitutde), and includes a new enhancement that forces the cursor to be visible even in full screen without the menu or menubar open.

Thanks to @Caladius, this also comes with a search function. It's setup to search widget names and tooltips, and displays each option's location in the rest of the menus.

Note to other devs: the LUS changes will need a bit of tweaking before being made a PR, so that it's not relying on 2ship settings and CVars directly, but is configurable and abstract.

image
image
image

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1979690160.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1979697628.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1979697989.zip)


retropieuser:
This is looking really good. 

Would it be possible to implement a hotkey to bring the menu up from the gamepad rather than Esc? As you still need a keyboard whereas with this menu we're almost at Big Picture style where this can be played on the tv without a mouse and keyboard required

retropieuser:
This is fantastic work, I can finally play things like 2s2h (and hopefully soon soh) on my TV without needing a mouse and keyboard! Thanks for a great job on this

One small tidbit, right now when you have the menu open, pressing gamepad_back only hides the menu while the toggle btn is pressed down, when you let go the menu reappears. To close the menu you have to select the button on the GUI.

Can this be changed so when you press the toggle btn in the menu it closes it please?

Malkierian:
It was specifically designed to do that, just like the menubar before it. I'm not experiencing what you're describing. In fact, if I hold the toggle button, it rapidly shows and hides it over and over again.

retropieuser:
Ah yes maybe I didn't explain it well. if I hold the toggle button it rapidly turns off and on.

In terms of a user case though and what they're likely to do, just two suggestions:- 

You could disable the toggle button when the menu is visible and re-enabling it when the menu isn't visible? So pressing it once brings up the menu and then you need to select close Menu inside the GUI?

Or you could have the toggle button always able to switch the menu from visible to invisible with a slight wait for presses so it doesn't happen too fast?

Just food for thought anyway. It's an awesome job all the same and thank you for working on it

Computerfan23:
Hi!

This is really great work and it is really amazing. Will this be able to work better for accessibility using the NVDA screen reader? Like using a text to speech?

Malkierian:
It actually should be possible to hook this up directly to the TTS system currently built into SoH when it gets ported over, without too much hassle.

Computerfan23:
That sounds great to hear. It would make it much easier for me to access the enhancements if I can get it readable with tts. Since it would be easier to access using the esc button i believe. Looking forward to when this will be available.

mm/2s2h/search/SearchMenu.h
```diff
@@ -0,0 +1,1377 @@
...
+                        gPlayState->frameAdvCtx.enabled = false;
+                    };
+                } },
+              { "Advance 1",
+                "",
+                "Advance 1 frame.",
+                WIDGET_BUTTON,
+                {},
+                [](widgetInfo& info) { CVarSetInteger("gDeveloperTools.FrameAdvanceTick", 1); } },
+              { "Advance (Hold)", "", "Advance frames while the button is held.", WIDGET_BUTTON } } } });
```
Archez: Frame advancing "advance (Hold)" is not doing anything/

mm/2s2h/search/SearchMenu.h
```diff
@@ -0,0 +1,1377 @@
...
+                "This notably reduces the overall motion blur strength but smooths out the trails.",
+                WIDGET_CHECKBOX },
+              { "On/Off",
+                "gEnhancements.Graphics.MotionBlur.Toggle",
+                "Enables Motion Blur.",
+                WIDGET_CHECKBOX,
+                {},
+                [](widgetInfo& info) {
+                    if (CVarGetInteger("gEnhancements.Graphics.MotionBlur.Toggle", 0) == 0) {
+                        R_MOTION_BLUR_ENABLED;
```
Archez: All the widget handling around motion blur seems to be incorrect/strange.

`On/Off` is only meant to be used when the setting is Dynamic, and is not supposed to be a CVar as its supposed to reflect the games register values. Similarly, `Strength` when set to dynamic, is also supposed to reflect the games register value, not a CVar.

Additionally, the changes in `MotionBlur.cpp` should be reverted.

mm/2s2h/search/SearchMenu.h
```diff
@@ -0,0 +1,1377 @@
...
+            disabledValue = true;
+            for (auto option : widget.activeDisables) {
+                disabledTempTooltip += std::string("- ") + disabledMap[option].second.reason + std::string("\n");
+            }
+            disabledTooltip = disabledTempTooltip.c_str();
+        }
+    }
+
+    switch (widget.widgetType) {
+        case WIDGET_CHECKBOX:
```
Archez: I think all the cvar widgets should probably have CVAR in the name, and then things like `WIDGET_CHECKBOX` can just be the regular checkbox variant (something the motion blur elements will need)

mm/2s2h/search/SearchMenu.h
Malkierian: Done.

mm/2s2h/search/SearchMenu.h
Malkierian: Done.

mm/2s2h/search/SearchMenu.h
Malkierian: Done.

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -0,0 +1,1514 @@
...
+                "",
+                "Motion Blur strength.",
+                WIDGET_SLIDER_INT,
+                { 0, 255, 180, {}, &motionBlurStrength },
+                [](widgetInfo& info) { R_MOTION_BLUR_ALPHA = motionBlurStrength; },
+                [](widgetInfo& info) {
+                    info.isHidden = disabledMap.at(DISABLE_FOR_MOTION_BLUR_MODE).value != 0 ||
+                                    disabledMap.at(DISABLE_FOR_MOTION_BLUR_OFF).active;
+                },
+                [](widgetInfo& info) { motionBlurStrength = R_MOTION_BLUR_ALPHA; } },
```
Archez: I believe the assignment of `motionBlurStrength` should happen as part of the `preFunc`, not the `postFunc`, so that the current frame's value of `R_MOTION_BLUR_ALPHA` is what gets displayed in the slider.

`postFunc` means it is delayed by a frame.

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -0,0 +1,1514 @@
...
+
+    if (widget.widgetOptions.sameLine) {
+        ImGui::SameLine();
+    }
+
+    switch (widget.widgetType) {
+        case WIDGET_CHECKBOX: {
+            bool* pointer = std::holds_alternative<bool*>(widget.widgetOptions.valuePointer)
+                                ? std::get<bool*>(widget.widgetOptions.valuePointer)
+                                : nullptr;
```
Archez: Unfortunately, passing in `nullptr` to `::Checkbox` is unsafe. This will lead to a null pointer dereference and crash.

Instead, we would need to either use a dummy value pointer as the fallback, or bail out of rendering this widget to be safe. We can add a debug `assert()` to at least trip this for local dev.

---

Separately, there are many `std::get()` in this file that are not wrapped with a `std::holds_alternative()` check so things are already inconsistent. Maybe just drop that part?

If we wanted to actually guard against all this, we could just wrap everything with a try/catch for `std::bad_variant_access`

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -0,0 +1,1514 @@
...
+            }
+            ImGui::SeparatorText(widget.widgetName);
+            if (widget.widgetOptions.color != COLOR_NONE) {
+                ImGui::PopStyleColor();
+            }
+            break;
+        case WIDGET_COMBOBOX: {
+            int32_t* pointer = std::holds_alternative<int32_t*>(widget.widgetOptions.valuePointer)
+                                   ? std::get<int32_t*>(widget.widgetOptions.valuePointer)
+                                   : nullptr;
```
Archez: Same thing here.

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -0,0 +1,1514 @@
...
+                { .valuePointer = (bool*)&gPlayState->frameAdvCtx.enabled },
+                nullptr,
+                [](widgetInfo& info) {
+                    if (disabledMap.at(DISABLE_FOR_NULL_PLAY_STATE).active) {
+                        info.isHidden = true;
+                    }
+                    if (!std::holds_alternative<bool*>(info.widgetOptions.valuePointer) ||
+                        std::get<bool*>(info.widgetOptions.valuePointer) != (bool*)&gPlayState->frameAdvCtx.enabled) {
+                        info.widgetOptions.valuePointer = (bool*)&gPlayState->frameAdvCtx.enabled;
+                    }
```
Archez: This logic looks very confusing and is dangerously setting the `valuePointer` with the memory address that reflects the member offset of `frameAdvCtx` against the value of `nullptr` (which means it becomes a non-nullptr value).

I think this should be re-worked as something like:

```c++
if (gPlayState != nullptr) {
    info.widgetOptions.valuePointer = (bool*)&gPlayState->frameAdvCtx.enabled
} else {
    info.widgetOptions.valuePointer = (bool*)nullptr // or dummy pointer
}
```

Additionally, we probably don't want the initializer to assign with `gPlayState` as that is going to be null during init and will experience the same behavior.

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -0,0 +1,1514 @@
...
+// `color` can be used in any way, depending on the widget. color pickers (COLOR_24 and COLOR_32) will eventually
+// use this for storing its current value; WIDGET_SEPARATOR_TEXT uses this to change the text color `windowName`
+// is used when using WIDGET_WINDOW_BUTTON, which optionally draws windows, like Stats, in the menu with a button
+// to pop the window out of the main menu overlay
+struct WidgetOptions {
+    CVarVariant min;
+    CVarVariant max;
+    CVarVariant defaultVariant;
+    std::unordered_map<int32_t, const char*> comboBoxOptions;
+    std::variant<bool*, int32_t*, float*> valuePointer;
```
Archez, line 127: [non-blocking]

The default initialized value for `valuePointer` ends up as `(bool*)nullptr` and the variant type as `bool*`.

We might want to consider allowing `nullptr_t` as a type so that we can explicitly deal with `nullptr` values.

Marking as non-blocking for this merge, but probably something to re-visit for when upstreaming to LUS.

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -0,0 +1,1514 @@
...
+                      .labelPosition = widget.widgetOptions.labelPosition == UIWidgets::LabelPosition::None
+                                           ? UIWidgets::LabelPosition::Above
+                                           : widget.widgetOptions.labelPosition })) {
+                if (widget.widgetCallback != nullptr) {
+                    widget.widgetCallback(widget);
+                }
+            }
+            break;
+        case WIDGET_SLIDER_INT:
+            if (UIWidgets::SliderInt(widget.widgetName, std::get<int32_t*>(widget.widgetOptions.valuePointer),
```
Archez: Same thing above about `nullptr` is bad for Sliders cause of dereference (and below for SliderFloat)

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -0,0 +1,1514 @@
...
+                            menuSearchText.erase(std::remove(menuSearchText.begin(), menuSearchText.end(), ' '),
+                                                 menuSearchText.end());
+                            std::transform(widgetStr.begin(), widgetStr.end(), widgetStr.begin(), ::tolower);
+                            widgetStr.erase(std::remove(widgetStr.begin(), widgetStr.end(), ' '), widgetStr.end());
+                            if (widgetStr.find(menuSearchText) != std::string::npos) {
+                                SearchMenuGetItem(info);
+                                ImGui::PushStyleColor(ImGuiCol_Text, UIWidgets::Colors::Gray);
+                                std::string origin =
+                                    fmt::format("  ({} -> {}, Clmn {})", menuLabel, sidebar.label, column);
+                                ImGui::Text(origin.c_str());
```
Archez: `warning: format string is not a string literal (potentially insecure) [-Wformat-security]`

```suggestion
                                ImGui::Text("%s", origin.c_str());
```

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -0,0 +1,1514 @@
...
+        case WIDGET_BUTTON:
+            if (UIWidgets::Button(widget.widgetName, { menuTheme[menuThemeIndex], widget.widgetOptions.size,
+                                                       widget.widgetTooltip, disabledValue, disabledTooltip })) {
+                if (widget.widgetCallback != nullptr) {
+                    widget.widgetCallback(widget);
+                }
+            }
+            break;
+        case WIDGET_WINDOW_BUTTON: {
+            if (widget.widgetOptions.windowName == "") {
```
Archez: `result of comparison against a string literal is unspecified (use an explicit string comparison function instead) [-Wstring-compare]`

`strlen` or this to be safer

```suggestion
            if (widget.widgetOptions.windowName == nullptr || widget.widgetOptions.windowName[0] == '\0') {
```

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -0,0 +1,1514 @@
...
+    WIDGET_AUDIO_BACKEND, // needed because of special operations that can't be handled easily with the normal combobox
+                          // widget
+    WIDGET_VIDEO_BACKEND  // same as above
+} WidgetType;
+
+typedef enum {
+    MOTION_BLUR_DYNAMIC,
+    MOTION_BLUR_ALWAYS_OFF,
+    MOTION_BLUR_ALWAYS_ON,
+};
```
Archez: Missing typedef name
```suggestion
} MotionBlurOption;
```

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -0,0 +1,1514 @@
...
+
+typedef enum {
+    DEBUG_LOG_TRACE,
+    DEBUG_LOG_DEBUG,
+    DEBUG_LOG_INFO,
+    DEBUG_LOG_WARN,
+    DEBUG_LOG_ERROR,
+    DEBUG_LOG_CRITICAL,
+    DEBUG_LOG_OFF,
+};
```
Archez: Missing typedef name

```suggestion
} DebugLogOption;
```

Archez (commented):
Ok just a few more things (mainly around the new changes and some warning fixes)

mm/2s2h/BenGui/SearchableMenuItems.h
Malkierian: To be fair, these typedef names weren't there when they were originally added, so it's not my fault, but I'll add them XD.

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -0,0 +1,1539 @@
...
+                ImGui::EndChild();
+            } break;
+            default:
+                break;
+        }
+        if (widget.postFunc != nullptr) {
+            widget.postFunc(widget);
+        }
+    } catch (const std::bad_variant_access& e) {
+        SPDLOG_ERROR(e.what());
```
Archez:
```suggestion
        SPDLOG_ERROR("Failed to draw menu item \"{}\" due to: {}", widget.widgetName, e.what());
```

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -0,0 +1,1539 @@
...
+                {},
+                [](widgetInfo& info) {
+                    int hz = Ship::Context::GetInstance()->GetWindow()->GetCurrentRefreshRate();
+                    if (hz >= 20 && hz <= 360) {
+                        CVarSetInteger("gInterpolationFPS", hz);
+                        Ship::Context::GetInstance()->GetWindow()->GetGui()->SaveConsoleVariablesOnNextTick();
+                    }
+                },
+                [](widgetInfo& info) {
+                    if (disabledMap.at(DISABLE_FOR_NOT_DIRECTX).active) {
```
Archez: Looks like in last commit  removed the hide for the non-DX match refresh rate checkbox.

Probably should add back the other hide so there aren't two match refresh options on DX.

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -0,0 +1,1539 @@
...
+      { [](disabledInfo& info) -> bool {
+           return !Ship::Context::GetInstance()->GetWindow()->GetGui()->SupportsViewports();
+       },
+        "Multi-viewports not supported" } },
+    { DISABLE_FOR_NOT_DIRECTX,
+      { [](disabledInfo& info) -> bool {
+           return Ship::Context::GetInstance()->GetWindow()->GetWindowBackend() !=
+                  Ship::WindowBackend::FAST3D_DXGI_DX11;
+       },
+        "Multi-viewports not supported" } },
```
Archez: Copy paste of previous disabled message. Should instead be something like "Only available on DirectX"

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -0,0 +1,1539 @@
...
+                WIDGET_CVAR_SLIDER_INT,
+                { 1, 8, 1 },
+                [](widgetInfo& info) {
+                    Ship::Context::GetInstance()->GetWindow()->SetMsaaLevel(CVarGetInteger(CVAR_MSAA_VALUE, 1));
+                } },
+#endif
+
+              { "Current FPS: %d",
+                "gInterpolationFPS",
+                tooltip.c_str(),
```
Archez: The pointer value produced by `std::string::c_tr()` is only valid while the parent string is alive. Because `topic` is a function scope variable, it gets cleaned up once `AddSettings()` finishes, and then the resulting text displayed in the tool tip is garbage.

You will either need to keep this variable alive by moving it outside of the function scope, or marking it as `static`.

Malkierian: Now that I look at it, there was no reason for it to be std::string to begin with, changing `tooltip` to `const char*` seems to have resolved the issue.

Archez (approved):
Thanks for your patience with the reviewing, I feel good with where this PR is at now. Once in, I'm sure we will get good feedback on further improvements we can make.

garrettjoecox:
~Idk what setting I had set, but I couldn't interact with the new menu until I deleted my JSON/Imgui.init files. Probably want to figure out what setting that is before we have to deal with it in support.~ Unable to reproduce unfortunately

Blocking:
- There seems to be some weirdness regarding whether or not the menu shows up when you press your controller's dedicated button. When I open up the game, my controller's `menu` button doesn't do anything. If I then click ESC on my keyboard to open the menu, then "grab focus" with my controller by moving the dpad, and then navigate to and close the menu with the "Close Menu" button in the UI, my controller seems to retain focus, and then the `menu` button on my controller works to re-open the menu. Additionally, regardless of if the menu has focus or not, when it is open and I try to close it using my controller's `menu` button, it closes and re-opens.

Not blocking:
- The icons within buttons are not vertically centered. I think this might have been an issue before, but with the larger font size it seems more exaggerated.
- Volume sliders on audio page are wrong format
- Some things in the wrong section or missing, eg 2 options under "time savers > dialog" don't belong, and warp point developer tool is missing
- I know this is obvious, but the navigation buttons & search should use our themed widgets
- We should just get rid of the "Press f1" popup all together, pressing your menu button or escape seems intuitive enough
- When the search page is opened, the input should be auto focused
- Menu.cpp/.h should both be under the BenGui/ folder
- Not sure how I feel yet about the pattern in which we define menu items. Probably won't know how I feel until I have to work with it for a while.

What is the plan for the old Navbar? It's going to get annoying real quick to have to add UI stuff in two places

Malkierian:
For the major problem, it's going to require some investigation. As usual, what I was developing on, WASAPI, has no issues with the way things are. Those issues only show up with SDL as the window manager, and the fact that they're handled differently between the two annoys me to no end, but that's where they end up manifesting.

For the non-blocking stuff:
1) I'm not even sure how to make that center. The problem seems to be that FA renders the icons at the text line floor, and I don't know if there's a way to change that.
2) Easily enough handled.
3) Yeah, the dialog things were supposed to go in Gameplay, I think. Not sure how that happened. And I knew I'd be missing some things XD
4) The nav buttons are easy enough, but I don't think the search widget has a themed widget variant
5) I can see the the prompt skip default to on, once we decide to yeet the navbar. The only reason it isn't is because I didn't know whether to yeet that in this PR or not. I figured people would want to have the option of it while getting used to the new menu.
6) I'm really not sure how to give the search bar focus only on first draw when changing to the search tab...
7) Also easily done
8) Any suggestions for improvement in that flow will always be welcome. It's already gone through 4 or 5 iterations between what I originally did and what Cal did for the search menu, and then what I did to further extend what he did. This is very much still WIP, it's just functional so that more people can test it.

garrettjoecox (approved):
Let's make an issue to track and detail the SDL thing

Malkierian:
> Let's make an issue to track and detail the SDL thing

Handled in https://github.com/Kenix3/libultraship/pull/677

PR #714, opened 2024-06-25
PowerShell clang fix

Malkierian:
Fixes script running in commandline mode by adding System.Net inclusion at the top.
Also adds clang-format.exe to gitignore.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1634302839.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1634306070.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1634308476.zip)

### DIFF
--- a/run-clang-format.ps1
+++ b/run-clang-format.ps1
@@ -1,3 +1,4 @@
+Using Namespace System.Net
 $url = "https://github.com/llvm/llvm-project/releases/download/llvmorg-14.0.6/LLVM-14.0.6-win64.exe"
 $llvmInstallerPath = ".\LLVM-14.0.6-win64.exe"
 $clangFormatFilePath = ".\clang-format.exe"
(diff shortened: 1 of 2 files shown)


Archez (approved):

run-clang-format.ps1
```diff
@@ -1,3 +1,4 @@
+Using Namespace System.Net
```
Archez: Can you confirm if just `Using Namespace System` works for you?
Down below we already do `net.webclient` so I think just the `System` namespace would be sufficient.

Malkierian: It does indeed.

PR #716, opened 2024-06-25
[Enhancement] Add First Cycle Skip

balloondude2:
This PR adds an enhancement that extends the Skip Intro Sequence enhancement to also skip the first cycle, starting the game with magic, Ocarina, Song of Time, Song of Healing, and Deku Mask. 

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2038324378.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2038330334.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2038350606.zip)


Archez:
Nice extension!

Similar to the "Skip Intro Sequence", I think we should strive to also set any of the corresponding permanent flags that belong to the first cycle events. Its possible that there are none, but if you haven't checked yet, there is an event log you can open to see what flags are set as you receive each of these items. The event log is a little noisy, and not all flags are permanent either, so you would have to perform some deduction on what should be set.

garrettjoecox (approved):
LGTM, as long as you're certain you didn't miss any flags/state.

We'll have to circle back on how this works/if it works with the eventual rando

garrettjoecox (commented):
Oh and also needs updating to modern menu

balloondude2:
I added the modern menu option. I went through the first cycle path again, and I realized I had missed the entrance cutscene flags for clocktown. I got those added, and am now confident I haven't missed any persistent flags from the first cycle. 

I also agree that this will need some iteration to be integrated with an eventual rando.

PR #717, opened 2024-06-25
[Enhancement] Enable putaway as Fierce Deity

Eblo:
This allows Fierce Deity to put away his sword instead of brandishing it at all times. Like other forms, this can be used in tandem with the restricted items cheat to use items the form isn't meant to use (often with visual bugs)

image
image
image

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2036288094.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2036298291.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2036342139.zip)


mm/src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -3761,8 +3761,10 @@ void Player_ProcessItemButtons(Player* this, PlayState* play) {
         ItemId item;
         EquipSlot i = func_8082FDC4();
 
-        i = ((i >= EQUIP_SLOT_A) && (this->transformation == PLAYER_FORM_FIERCE_DEITY) &&
-             (this->heldItemAction != PLAYER_IA_SWORD_TWO_HANDED))
+        i = GameInteractor_Should(GI_VB_FD_ALWAYS_WIELD_SWORD,
+                                  (i >= EQUIP_SLOT_A) && (this->transformation == PLAYER_FORM_FIERCE_DEITY) &&
+                                      (this->heldItemAction != PLAYER_IA_SWORD_TWO_HANDED),
+                                  NULL)
```
garrettjoecox: Don't need these trailing NULL's anymore

mm/2s2h/Enhancements/Player/FierceDeityPutaway.cpp
```diff
@@ -0,0 +1,20 @@
...
+        if (CVarGetInteger("gEnhancements.Player.FierceDeityPutaway", 0)) {
+            Player* player = GET_PLAYER(gPlayState);
+            if (player->transformation == PLAYER_FORM_FIERCE_DEITY)
+                *should = true;
+        }
+    });
+
+    REGISTER_VB_SHOULD(GI_VB_FD_ALWAYS_WIELD_SWORD, {
+        if (CVarGetInteger("gEnhancements.Player.FierceDeityPutaway", 0)) {
+            Player* player = GET_PLAYER(gPlayState);
```
garrettjoecox:
```suggestion
```

mm/2s2h/Enhancements/Player/FierceDeityPutaway.cpp
```diff
@@ -0,0 +1,20 @@
+#include <libultraship/bridge.h>
+#include "Enhancements/GameInteractor/GameInteractor.h"
+#include "variables.h"
+
+void RegisterFierceDeityPutaway() {
+    REGISTER_VB_SHOULD(GI_VB_SHOULD_PUTAWAY, {
+        if (CVarGetInteger("gEnhancements.Player.FierceDeityPutaway", 0)) {
+            Player* player = GET_PLAYER(gPlayState);
+            if (player->transformation == PLAYER_FORM_FIERCE_DEITY)
```
garrettjoecox, line 9: The original condition that you are overriding here is as follows:
```
(
  (this->transformation == PLAYER_FORM_HUMAN) || (this->transformation == PLAYER_FORM_ZORA)
) && (
  (this->heldItemAction >= PLAYER_IA_SWORD_KOKIRI) ||
  ((this->stateFlags2 & PLAYER_STATE2_100000) && (play->actorCtx.targetCtx.fairyActor == NULL))
)
```

Here it looks like we're ignoring the entire second part of that condition, is that intentional? Is that okay?

garrettjoecox (commented):
Needs a modern menu option, along with a question about the VB condition

mm/2s2h/Enhancements/Player/FierceDeityPutaway.cpp
Eblo: I think it may have been intentional, but it's been a while so I don't remember why if so. I'll investigate that when rebasing and addressing your other feedback.

Eblo: I looked into it. This was not intentional. I changed it to only circumvent the transformation check.

garrettjoecox (approved):

PR #720, opened 2024-06-26
Fix: Prevent crash on loading OOB info panel textures in kaleido

Archez:
This adds some bounds checking on the texture loading for kaleido's info panel to prevent crashes when OOB values are passed in. Notably this prevents a crash when performing index warp with the cursor point over the Stone Tower map point.

This uses our empty texture, which unfortunately means we can't preserve the garbage texture that normally gets presented on console. But I don't think that is worth trying to account for.

This also adjusts the texture handling so that resource strings are sent to the renderer, enabling HD texture support for the info panel names.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1638735656.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1638738717.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1638740408.zip)

### DIFF
--- a/mm/src/overlays/kaleido_scope/ovl_kaleido_scope/z_kaleido_scope_NES.c
+++ b/mm/src/overlays/kaleido_scope/ovl_kaleido_scope/z_kaleido_scope_NES.c
@@ -253,7 +253,7 @@ u8 gAreaGsFlags[] = {
 // TODO: Also applies to owl warps
 s16 sGameOverRectPosY = 66;
 
-void Kaleido_LoadMapNameStatic(void* segment, u32 texIndex) {
+void Kaleido_LoadMapNameStatic(void** segment, u32 texIndex) {
     static const char* gMapNameStatics[] = {
         gMapPointGreatBayENGTex,      gMapPointZoraHallENGTex,        gMapPointRomaniRanchENGTex,
         gMapPointDekuPalaceENGTex,    gMapPointWoodfallENGTex,        gMapPointClockTownENGTex,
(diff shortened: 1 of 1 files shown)


mm/src/overlays/kaleido_scope/ovl_kaleido_scope/z_kaleido_scope_NES.c
```diff
@@ -262,7 +262,15 @@ void Kaleido_LoadMapNameStatic(void* segment, u32 texIndex) {
         gMapPointSouthernSwampENGTex, gMapPointMountainVillageENGTex, gMapPointMilkRoadENGTex,
         gMapPointZoraCapeENGTex,
     };
-    void* tex = ResourceMgr_LoadTexOrDListByName(gMapNameStatics[texIndex]);
+
+    // 2S2H [Port] Bounds check texture to load to prevent crashes when using index warp
+    void* tex;
+    if (texIndex >= 0 && texIndex < ARRAY_COUNTU(gMapNameStatics)) {
```
louist103: I think the better long term fix is what I did here:
https://github.com/HarbourMasters/2ship2harkinian/pull/430/commits/0f2e7c0e9b0c795b50b1f4fd56f5e720b182680d

Archez: Sure, I can cherry pick that over instead, and that should also support HD textures properly.

Archez: Done!

PR #723, opened 2024-06-27
[Enhancement] Skip the scarecrow song

Eblo:
Much like the setting in SoH, this allows Link to summon Pierre just by pulling out his instrument near a spawn point. This incidentally means that the 3-day cycle reset won't impact Link's ability to call Pierre.

https://github.com/HarbourMasters/2ship2harkinian/assets/7004497/f2458033-0be9-4afd-81a6-d0b41412af07

If and when entrance randomizer becomes a thing, this will have to be taken into consideration, as the song is typically gated by being able to reach Pierre in Clock Town.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2041046363.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2041059830.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2041113072.zip)

### DIFF
--- a/mm/src/overlays/actors/ovl_En_Kakasi/z_en_kakasi.c
+++ b/mm/src/overlays/actors/ovl_En_Kakasi/z_en_kakasi.c
@@ -7,6 +7,7 @@
 #include "prevent_bss_reordering.h"
 #include "z_en_kakasi.h"
 #include "objects/object_ka/object_ka.h"
+#include "2s2h/GameInteractor/GameInteractor.h"
 
 #define FLAGS (ACTOR_FLAG_TARGETABLE | ACTOR_FLAG_FRIENDLY | ACTOR_FLAG_10 | ACTOR_FLAG_2000000)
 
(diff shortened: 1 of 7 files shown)


mm/src/overlays/actors/ovl_En_Kakasi/z_en_kakasi.c
```diff
@@ -1053,8 +1054,10 @@ void EnKakasi_SetupIdleUnderground(EnKakasi* this) {
 }
 
 void EnKakasi_IdleUnderground(EnKakasi* this, PlayState* play) {
-    if (CHECK_WEEKEVENTREG(WEEKEVENTREG_79_08) && (this->picto.actor.xzDistToPlayer < this->songSummonDist) &&
-        ((BREG(1) != 0) || (play->msgCtx.ocarinaMode == OCARINA_MODE_PLAYED_SCARECROW_SPAWN))) {
+    if ((this->picto.actor.xzDistToPlayer < this->songSummonDist) &&
+        ((BREG(1) != 0) ||
+         (GameInteractor_Should(GI_VB_NEED_SCARECROW_SONG,
+                                play->msgCtx.ocarinaMode == OCARINA_MODE_PLAYED_SCARECROW_SPAWN, NULL)))) {
```
Archez: The logic here looks to be messed up. The original `weekeventreg` was dropped, which means it is now impossible to summon the scarecrow with the enhancement off.

Thinking in hooks, the `weekeventreg` probably should be what gets passed in as the condition to the hook, and then setting it to true would bypass the need for the reg to be set.

You may also need to replicate this whole condition within your hook handler so you don't exit the ocarina mode incorrectly (when not close enough to the scarecrow).

Eblo: Wow, don't know how I dropped that. It does seem to still work with the hook disabled and playing the song normally, but you're right that it's not correct.

Eblo: Based on more testing, I believe that register is the flag that says the player created a scarecrow song for this cycle. When I restore this condition but leave the hook where it is, Pierre only appears if I taught him a song this cycle.

I have refactored my logic based on your input. Thanks for the catch.

garrettjoecox (approved):
LGTM, just needs a modern menu entry and moving the header over to using the `cycle.h`

Eblo:
> LGTM, just needs a modern menu entry and moving the header over to using the `cycle.h`

Done, but moved to `Songs.h`

PR #728, opened 2024-06-28
Fix: Clean up flags when using debug warp

Archez:
This adds some flag cleanups when using the debug warp.

* For all saves, we clear two flags relating to the day transition. This should resolve issues around not being able to pause after activating debug warp during a day/night transition.
* For debug saves we want debug warp to completely reset the save. The `Sram_InitDebugSave` would memset `gSaveContext.save.saveInfo` which has all the persistent flags, but there are few other flags higher up on the save context that wouldn't clear out. So I copied over some of the logic from `Sram_OpenSave` to help prevent things from sticking around.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1650070261.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1650077147.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1650078933.zip)

### DIFF
--- a/mm/src/overlays/gamestates/ovl_select/z_select.c
+++ b/mm/src/overlays/gamestates/ovl_select/z_select.c
@@ -21,7 +21,7 @@ void MapSelect_LoadGame(MapSelectState* this, u32 entrance, s32 spawn) {
     if (gSaveContext.fileNum == 0xFF) {
         Sram_InitDebugSave();
 
-        // #region 2S2h [Debug] Clear all flags when using debug file slot.
+        // #region 2S2H [Debug] Clear all flags when using debug file slot.
         // This is mostly copied from Sram_OpenSave.
         for (size_t i = 0; i < ARRAY_COUNT(gSaveContext.eventInf); i++) {
             gSaveContext.eventInf[i] = 0;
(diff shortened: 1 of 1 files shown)


garrettjoecox (approved):
> This should resolve issues around not being able to pause after activating debug warp during a day/night transition.

So I'm not just crazy. nice find

mm/src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -20,13 +20,33 @@ void MapSelect_LoadConsoleLogo(MapSelectState* this) {
 void MapSelect_LoadGame(MapSelectState* this, u32 entrance, s32 spawn) {
     if (gSaveContext.fileNum == 0xFF) {
         Sram_InitDebugSave();
+
+        // #region 2S2h [Debug] Clear all flags when using debug file slot.
```
inspectredc:
```suggestion
        // #region 2S2H [Debug] Clear all flags when using debug file slot.
```

inspectredc (approved):

mm/src/overlays/gamestates/ovl_select/z_select.c
Archez: The amount of times that I incorrectly type these 4 characters is hilarious

Eblo:
> This should resolve issues around not being able to pause after activating debug warp during a day/night transition.

This also occurs after warping to Odolwa's Lair even if it's not during a transition. I can confirm that does not happen in this build.

PR #732, opened 2024-07-02
[Enhancement] Get Both Letter to Mama rewards in one cycle

Eblo:
This removes the need to go through the Anju and Kafei sidequest twice to get both of the mutually exclusive rewards from delivering the Letter to Mama. Now the player can either get the Postman's Hat when hand delivering the letter to Madame Aroma or get the Chateau Romani from the Postman after he delivers the letter.

https://github.com/HarbourMasters/2ship2harkinian/assets/7004497/4163b202-450a-4255-ba93-35d57f6a5ac7

https://github.com/HarbourMasters/2ship2harkinian/assets/7004497/93015751-3211-4381-8e08-314373277898

I filed this under Time Cycle, since it doesn't strictly allow the player to circumvent any requirements needed to get either item, with one edge case exception. Getting the Chateau Romani requires Romani's Mask only if it's after 10 P.M., while the Postman can be interacted with at any time on the night of the final day. Madame Aroma is, however, normally reachable without the mask if done so before 10 P.M.

This will require another look when randomizer considerations are on the table. I made each actor mirror the other's conditions when deciding whether to give their rewards. This made me realize that the Postman himself doesn't check any persistent flags when deciding on whether to give his hat to the player; he checks the player's inventory directly for the hat. A persistent flag does exist in the Bomber's Notebook (`WEEKEVENTREG_BOMBERS_NOTEBOOK_EVENT_RECEIVED_POSTMANS_HAT`), but the Postman does not check that when deciding on whether to give the Postman's Hat.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1660570197.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1660572013.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1660580999.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenuBar.cpp
+++ b/mm/2s2h/BenGui/BenMenuBar.cpp
@@ -465,6 +465,10 @@ void DrawEnhancementsMenu() {
                 { .tooltip = "Playing the Song Of Time will not reset the Sword back to Kokiri Sword." });
             UIWidgets::CVarCheckbox("Do not reset Rupees", "gEnhancements.Cycle.DoNotResetRupees",
                                     { .tooltip = "Playing the Song Of Time will not reset the your rupees." });
+            UIWidgets::CVarCheckbox(
+                "Both Letter to Mama Rewards", "gEnhancements.Cycle.BothLetterToMamaRewards",
+                { .tooltip =
+                      "Giving the Letter to Mama to either Madame Aroma or the Postman will yield both rewards." });
 
             ImGui::PushStyleColor(ImGuiCol_Text, ImVec4(255, 255, 0, 255));
             ImGui::SeparatorText("Unstable");
--- a/mm/2s2h/Enhancements/Cycle/BothLetterToMamaRewards.cpp
+++ b/mm/2s2h/Enhancements/Cycle/BothLetterToMamaRewards.cpp
@@ -1,0 +1,47 @@
+#include <libultraship/bridge.h>
+#include "Enhancements/GameInteractor/GameInteractor.h"
+
+extern "C" {
+#include "functions.h"
+#include "z64message.h"
+#include "variables.h"
+extern PlayState* gPlayState;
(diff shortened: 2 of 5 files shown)


inspectredc (approved):
Awesome! LGTM

Archez:
What about instead of having each NPC check and grant both rewards, we could just make the Priority Mail persist in the players inventory?

This way players would still need to interact with both NPCs in the expected way/at the correct times, but the player can do both in the same cycle. No item replaces the Priority Mail item in it's slot, so there shouldn't be any conflicts there. Probably would simplify future rando handling as we don't have to worry about if the player has "access" to the NPCs or what items they give.

Curious what others think. Personally I would lean towards this.

inspectredc:
That sounds like a good idea to me yea, not too familiar with all the mechanics behind these things, but if its only checking for the item being shown to them that sounds good

Eblo:
> No item replaces the Priority Mail item in it's slot, so there shouldn't be any conflicts there.

The Room Key uses that same slot. In vanilla this isn't a problem because the letter replaces it anyway, and all the key grants access to is an optional cutscene and a rupee chest. However, depending on whatever a rando would do, persisting the letter could present a problem. If there were a seed where one of these characters gave the Room Key, but it got replaced with the letter again due to this setting, the player would be unable to access whatever reward the Knife Chamber might have.

This does not matter right now, and whether we go with what is currently here or persist the letter, this setting will need another once-over when that time comes. I do like the idea of making things simpler, but as it is, this does mirror existing logic and seems clearer to me.

Malkierian:
That sounds like something the way adult trade in OoT is handled would solve.

Eblo:
Superseded by #791.

Issue #525, opened 2024-05-27
Hang on launch when non-ascii character in path


thistle-witch:
### Description of the crash:
When launching 2ship when there is a non-ascii character in the path an error message pops up on launch and the game window hangs indefinitely.

### Can you reproduce? If so please list the steps:
- [ ] No, I cannot reproduce this crash
- [x] Yes, here are the steps to reproduce the crash:
    - Place a newly extracted 2ship folder in a path that contains a non-ascii character (example: þ)
    - Generate the o2r files by providing a MM rom
    - Attempt to launch 2ship

### Screenshots or videos:
![image](https://github.com/HarbourMasters/2ship2harkinian/assets/129108882/28e560f1-e65f-4de7-9681-b0176762844f)

### Logs:
No log file was generated

### System Information:
- OS: Windows 11
- GFX Backend: DirectX
- Game Version: 1.0.0 Rika Alfa
- Rom Version: 1.0 NTSC Cartridge dump

garrettjoecox:
Fixed by #743

PR #743, opened 2024-07-10
Allow non-ascii characters in filepath on Windows

balloondude2:
This PR fixes Issue #525 and allows 2ship to properly launch when there are non-ascii characters (such as þ) in the file path. I'm not sure if this was a Windows-only problem, but Windows is the only thing I can test. 

I'm not sure if I put the changes in the best spot (src > code > main.c), so I'm open to moving those lines somewhere else if it's more appropriate. 

Before:
before

After:
after

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1684814238.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1684819633.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1684822884.zip)

### DIFF
--- a/mm/src/code/main.c
+++ b/mm/src/code/main.c
@@ -1,6 +1,7 @@
 #ifdef _WIN32
 #include <Windows.h>
 #include <stdio.h>
+#include <locale.h>
 #endif
 
 #include "audiomgr.h"
@@ -66,6 +67,8 @@ void SDL_main(int argc, char** argv /* void* arg*/) {
 #ifndef _DEBUG
     ShowWindow(GetConsoleWindow(), SW_HIDE);
 #endif
+    // Allow non-ascii characters for Windows
+    setlocale(LC_ALL, ".UTF8");
 #endif // _WIN32
 
     InitOTR();


Archez:
I've tried to do some reading around this and other options. I found two main links from Microsoft about this:
* `setLocale` https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/setlocale-wsetlocale?view=msvc-160#utf-8-support
* App Manifest changes https://learn.microsoft.com/en-us/windows/apps/design/globalizing/use-utf8-code-page

Both achieve similar goals in different ways. The manifest one "feels" like a better approach in the sense that UTF-8 is enabled the moment the program executes, where as `setLocale` isn't valid until it is called in code at runtime. However, the manifest change requires users to be on a specific version of Windows to leverage it, where as `setLocale` can be statically linked so that it supports more Windows versions.

Considering you call `setLocale` basically as early as possible, I think it would probably be safe to just try this first and see if user reports die down.

Archez (approved):

PR #746, opened 2024-07-15
[Enhancement] Actor Nametags

Archez:
Ports over Actor Nametags from SoH https://github.com/HarbourMasters/Shipwright/pull/3083

### Notable differences from SoH include:
* Nametags render roughly 2x farther (this could be changed to any value really, but if its too far away you cant really read it anyways)
* Individual nametags can set whether ZBuffer should be used (obscured by geometry)
  * Debug nametags from the actor viewer currently set this so that it is easier to see them. We can always make this a configurable option in the actor viewer.
* Hooks are optimized to (un)registered when there exists nametags

https://github.com/user-attachments/assets/eb3fb847-9e86-4469-a1f6-e1e8a3f42f97

---

### New Hooks:
* `OnPlayDrawWorldEnd` - Fires after the world is done drawing, but before the interface. This allows enhancements to render things that only need to render while the world is visible (etc, collision viewer, nametags). This also is within the frame interpolation recording bounds.
* `OnActorDestroy` - Called right before an actor is freed from the arena. When actors are removed via room transitions or during day/night transitions, they will not always call `Actor_Kill`, but `Actor_Destroy` will always be called in both cases. `OnActorDestroy` is a better hook to use over `OnActorKill`

I've also regrouped some things in the actor viewer and fixed a bug where having a selected actor could crash through room transitions or day/night transitions.

Additionally this added a new `ShipUtils` where I have currently added some helper methods for things I needed. If we feel this should just belong to `BenPort`, I can move them there too, but I was trying to avoid that becoming a catch all file.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1897745192.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1897757443.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1897759781.zip)


balloondude2 (approved):
Looks great to me! I also didn't encounter any issues with either the Actor Viewer or the Collision Viewer during playtesting.

mm/2s2h/DeveloperTools/ActorViewer.cpp
```diff
@@ -77,6 +111,69 @@ void ResetVariables() {
...
+    if (CVarGetInteger("gDeveloperTools.ActorViewer.NameTags", ACTORVIEWER_NAMETAGS_NONE) ==
+        ACTORVIEWER_NAMETAGS_NONE) {
+        return;
+    }
+
+    actorInitHookID =
+        GameInteractor::Instance->RegisterGameHook<GameInteractor::OnActorInit>(ActorViewer_AddTagForActor);
+}
+
+void ActorViewer_RegisterHooks() {
```
briaguya0: is the idea here to put a bunch of different hooks in here? if so the name seems good as-is, if not it might be nice to have a less generic name

mm/src/code/z_play.c
```diff
@@ -1514,8 +1514,9 @@ void Play_DrawMain(PlayState* this) {
                 }
             }
 
-            // Draw collision before the PostWorldDraw label so that collision is not drawn during pause
-            DrawCollisionViewer();
+            // Draw Enhancements that need to be placed in the world. This happens before the PostWorldDraw
+            // so that they aren't drawn when the pause menu is up (e.g. collision viewer, actor name tags)
+            GameInteractor_ExecuteOnPlayDrawWorldEnd();
```
briaguya0, line 1519: :heart:

briaguya0 (approved):
one comment but LGTM

mm/2s2h/DeveloperTools/ActorViewer.cpp
Archez: I suppose I was mainly following the pattern of each enhancement file having a `_RegisterHooks()` method, but I'm not sure what other hooks there could be in the future.

Considering this is just meant for preventing crashes with the actor viewer selected actor being killed, I think I can rework this hook to be registered dynamically instead like the other actor hooks in here (whenever an actor is selected/changed from the dropdown), since majority of people won't have the actor viewer open anyways so no need to run the hook.

mm/2s2h/DeveloperTools/ActorViewer.cpp
```diff
@@ -16,6 +18,20 @@ typedef enum Method {
     HOLD,
 } Method;
 
+typedef enum {
+    ACTORVIEWER_NAMETAGS_NONE,
+    ACTORVIEWER_NAMETAGS_DESC,
+    ACTORVIEWER_NAMETAGS_NAME,
+    ACTORVIEWER_NAMETAGS_BOTH,
```
garrettjoecox: For what it’s worth, I frequently have a local change that makes the actor params also visible in the name tag for soh, would be useful to just have out of the box

Archez: Did you have a preference for hex display, decimal, or both? I was thinking both like: `0x00A1, (161)`

I'd probably change this from a drop down to a multi-select as well then.

garrettjoecox: No preference, both works

Archez:
Updates since last review:
* Refactor a bit of the actor viewer to get rid of unused variables
* Improved dynamic hook (un)registration for selected actor
* Added more options for actor name tag display

image

garrettjoecox (approved):
merge pls

PR #749, opened 2024-07-21
Implement save backups and better invalid json detection

Archez:
This PR enables the builtin save backup system to work properly.

There were two main pieces missing on our end to make this happen:

First, when a new cycle save is persisted to file, the game writes the save data twice to the sram buffer and uses double the "page" size of a regular new cycle save. This translates into sram flashing both the main save and the backup save in one operation.

I've added a check within our save manager write method to check if the "page" size is this "special" value, and if so, perform a second save operation to the matching backup slot.

The second piece is that the game computes checksum values on saves and performs multiple check points. The file select menu does the following:
* If a checksum fails, the save is discarded
* If the main save and the backup save have mismatching values, the backup save is updated to match the main save

We were hard coding the checksum calculation to `1` which meant that backup saves where not created/updated. Additionally, in some of the variables that the game compares with the checksum, a value of `1` is used to indicated no backup save exists.

I've reverted the checksum calculation back to vanilla so that all this logic can run properly again. Because the checksum can be invalidated if someone modifies their save file externally, or a save migration changes the save, I've opted to always recompute the checksum upon parsing the save json. This way the game never thinks the save is "invalid". I think this is a worthy compromise to not make editing save files "painful".

---

This PR also expands the json parsing/typing try/catches to handle a few move edge cases with invalid json files.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1722595798.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1722597716.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1722597829.zip)

### DIFF
--- a/mm/src/code/z_sram_NES.c
+++ b/mm/src/code/z_sram_NES.c
@@ -690,10 +690,6 @@ void Sram_IncrementDay(void) {
 }
 
 u16 Sram_CalcChecksum(void* data, size_t count) {
-    // #region 2S2H [Port] I'm not really sure what this is doing or how to port it, for now always return the same
-    return 1;
-    // #endregion
-
     u8* dataPtr = data;
     u16 chkSum = 0;
 
--- a/mm/2s2h/SaveManager/SaveManager.cpp
+++ b/mm/2s2h/SaveManager/SaveManager.cpp
@@ -12,10 +12,12 @@
 extern "C" {
 #include "src/overlays/gamestates/ovl_file_choose/z_file_select.h"
 extern FileSelectState* gFileSelectState;
+
+u16 Sram_CalcChecksum(void* data, size_t count);
 }
 
 // This entire thing is temporary until we have a more robust save system that
-// supports backwards compatability, migrations, threaded saving, save sections, etc.
+// supports backwards compatibility, migrations, threaded saving, save sections, etc.
 typedef enum FlashSlotFile {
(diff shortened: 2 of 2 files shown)


mm/src/code/z_sram_NES.c
```diff
@@ -690,10 +690,6 @@ void Sram_IncrementDay(void) {
 }
 
 u16 Sram_CalcChecksum(void* data, size_t count) {
-    // #region 2S2H [Port] I'm not really sure what this is doing or how to port it, for now always return the same
```
garrettjoecox, line 693: So we’re okay using the vanilla calculation? Originally I thought we might have needed to modify it cause it looked like pointer math, but now I see it’s seems to just be a sum of all bytes casted to u8 or something

garrettjoecox (approved):

mm/src/code/z_sram_NES.c
Archez: Yup. And the size passed in accounts for our ShipSaveInfo struct additions too.

I don't see an issue with using it. Ultimately, with this PR recomputing the value on save load, the checksum is only used for deciding if a new backup should be made.

inspectredc (approved):

PR #756, opened 2024-08-14
Various custom asset support

Archez:
This PR adds various support of loading custom assets and toggling between assets (read: this is not meant to be all encompassing of every possible custom asset)

List so far of changes:
* `Lib_SegmentedToVirtual` no longer loads resource data and instead will just pass through whatever pointer it is given. This will ensure that asset path strings end up to the renderer or later in some of the games systems (animations/skeletons)
  * Fixes things like eye/mouth textures, effects/particles, Tektites, snow trees, etc
* Title cards render as one rectangle now
* Debug GfxPrint now uses font data from archives if available
* Support pause menu overworld map with multi tile load that handles asset paths

New issues introduced by these changes that need to be resolved:
- [x] ~~Animation timings are incorrect~~
- [x] ~~Manually load resources that are no longer freely loaded by `Lib_SegmentedToVirtual` (mostly Vertex data)~~ 

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1833939754.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1833948719.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1833955771.zip)


Archez:
Marking this as ready to review. The listed changes wrap up most of the reported issues with the mm reloaded pack.

Remaining issues will be addressed in follow up PRs as these changes will be extensive enough to warrant individual PRs
* Igos Curtains (CPU modified)
* Couple authentic texture overflows (Arrow tips, Freezards, Iron Knuckle effect) (this would resemble the similar SoH authentic texture overflow enhancement)
* Sakon hideout conveyor belt door (Seems like irregular texture UV coordinates that are not scaling correctly for HD texture sizes)
* Minimaps (requires pulling in minimap updates from decomp to get extracted textures)

mm/src/code/c_keyframe.c
```diff
@@ -187,20 +184,22 @@ void Keyframe_FlexPlayLoopWithMorph(KFSkelAnimeFlex* kfSkelAnime, KeyFrameAnimat
...
         skeleton = ResourceMgr_LoadKeyFrameSkelByName(skeleton);
     }
-
-    if (kfSkelAnime->skeleton != skeleton) {
-        kfSkelAnime->skeleton = skeleton;
-    }
-
     if (ResourceMgr_OTRSigCheck(animation)) {
         animation = ResourceMgr_LoadKeyFrameAnimByName(animation);
+        // Overwrite the end time after fetching the animation data has the parent caller most liekly
```
inspectredc:
```suggestion
        // Overwrite the end time after fetching the animation data as the parent caller most likely
```

mm/src/code/z_lib.c
```diff
@@ -678,10 +677,7 @@ void Lib_Nop801004FC(void) {
 }
 
 void* Lib_SegmentedToVirtual(void* ptr) {
-    if (ResourceMgr_OTRSigCheck(ptr)) {
-        return ResourceGetDataByName(ptr); // SEGMENTED_TO_VIRTUAL(ptr);
-    }
-    return ptr;
+    return SEGMENTED_TO_K0(ptr);
```
inspectredc, line 684: I thought this was a bit confusing to have rather than just returning ptr (which is what the macro does) just because what SEGMENTED_TO_K0 does in decomp implies some kind of lookup being involved to me. Idk if a comment or just removing the macro would feel more appropriate to me

inspectredc (approved):
LGTM just some small things

mm/src/code/z_lib.c
Archez: I more so wanted to just restore the lines back to their decomp equivalent. The macro has already been adjusted by us, and many things use this macro directly in code instead of using this function.

I can add a comment here that states roughly these 4 methods in this file all essentially return the given value back unmodified.

Archez:
Updated for PR feedback.

Also found a few more resources that needed to be loaded manually so those have been added to the PR.

mm/src/code/z_lib.c
Archez: Left a comment, hopefully that addresses your concerns 👍

inspectredc (approved):
Looks good and thanks for the added comment! New changes all good too 👍🏼

PR #765, opened 2024-09-06
Add restoration for woodfall mountain cleared state

Archez:
There is an unused model for the Woodfall Mountain without poison when seen from Terminal Field. Nothing in N64 MM shows this model, but in 3DS MM it was fixed to display properly.

This PR adds a restoration toggle so that the Mountain model will react to the current cycle Woodfall clear flag and display properly.

image

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1903091571.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1903099191.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1903102029.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Restorations/WoodfallMountainAppearance.cpp
+++ b/mm/2s2h/Enhancements/Restorations/WoodfallMountainAppearance.cpp
@@ -7,6 +7,11 @@ extern "C" {
 #include "overlays/actors/ovl_Bg_Breakwall/z_bg_breakwall.h"
 }
 
+typedef enum {
+    /*  2 */ BGBREAKWALL_F_2 = 3, // Poisoned Woodfall Mountain
+    /*  3 */ BGBREAKWALL_F_3 = 3, // Spring Woodfall Mountain
+} BgBreakwallParamEx;
+
 void RegisterWoodfallMountainAppearance() {
     static HOOK_ID breakwallInitID = 0;
     GameInteractor::Instance->UnregisterGameHookForID<GameInteractor::OnActorInit>(breakwallInitID);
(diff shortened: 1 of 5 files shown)


mm/2s2h/Enhancements/Restorations/WoodfallMountainAppearance.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+    if (!CVarGetInteger("gEnhancements.Restorations.WoodfallMountainAppearance", 0)) {
+        return;
+    }
+
+    breakwallInitID = GameInteractor::Instance->RegisterGameHookForID<GameInteractor::OnActorInit>(
+        ACTOR_BG_BREAKWALL, [](Actor* actor) {
+            if (BGBREAKWALL_GET_F(actor) == 2 && CHECK_WEEKEVENTREG(WEEKEVENTREG_CLEARED_WOODFALL_TEMPLE)) {
+                actor->params = (actor->params & 0xFFF0) | 3;
+            }
```
inspectredc: wondering if we should add our own BGBREAKWALL_F_2 and BGBREAKWALL_F_3 here

inspectredc (approved):
LGTM left a small comment on adding enums but not sure if this is more of a decomp will probably doc this better kinda problem

mm/2s2h/Enhancements/Restorations/WoodfallMountainAppearance.cpp
Archez: Like just at the top of this file, or within the decomp header?
I wouldn't be opposed to it in this file, where we can also add a comment, but for header I think it might be overkill/better left for decomp.

inspectredc: I think for now just in its own file like youve done is good yea 👍🏼

PR #766, opened 2024-09-06
Use variadic arguments with GameInteractor_Should

Archez:
This PoC shows one way of supporting variadic arguments from C -> C++ for the `GameInteractor_Should` hook.
It doesn't seem possible to be able to `extern` `template <....Args>` style functions, so I've opted for the "variadic function" approach which is C friendly.

- This works by using the `...` operator in the external hook execute function, then assigning the arguments into a `va_list` variable via `va_start`, then passing the va_list to each hook subscriber.
- Because we can have multiple subscribers for one execution, we have to make sure that subscribers make a copy of the incoming va_list using `va_copy` so that the original list is untouched between each subscriber.
- Then `va_arg` is used to read the next argument into the desired type. These arguments can be of any pointer type, or a standard type after any "default argument promotions".
- At the end, `va_end` must be used to signal that the va_list is no longer needed

I've opted to keep the existing `void* opt` argument before the `va_list` as most likely 90% of our hooks won't need more than just that, so it would be nice to not have to deal with the boiler plate of `va_list` in every hook.

In the second commit I show how we can also move some of the boiler plate within the `REGISTER_VB_SHOULD` macro.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1974922888.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1974927483.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/1974932819.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Camera/DebugCam.cpp
+++ b/mm/2s2h/Enhancements/Camera/DebugCam.cpp
@@ -232,7 +232,7 @@ void RegisterDebugCam() {
 
     if (CVarGetInteger("gEnhancements.Camera.DebugCam.Enable", 0)) {
         freeCamVBHookId = REGISTER_VB_SHOULD(GI_VB_USE_CUSTOM_CAMERA, {
-            Camera* camera = static_cast<Camera*>(opt);
+            Camera* camera = va_arg(args, Camera*);
             Camera_DebugCam(camera);
             *should = false;
         });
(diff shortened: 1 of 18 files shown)


Archez:
Based on conversations in discord, I've gone ahead and dropped the explicit `void* opt` argument in the `GameInteractor_Should` hook in favor of handling it through `va_arg`

This means two things:
* Dropping support for the `ForPtr` variant of hook handler for this specific hook. This is because there is no safe way to "peak" at the `va_list` to find a suitable pointer type. There is no example of this handler usage in 2ship or soh, so this was deemed fine.
* The external C function for `GameInteractor_Should` now has to type the `result` arg as an integer type of at least `int32_t` due to default argument promotions for `va_start`'s `param_N` arg. A down cast is used to still create the actual bool pointer that gets passed to the handlers, so this should have no visible impact to consumers.

garrettjoecox (approved):

inspectredc (approved):

PR #776, opened 2024-09-23
[Enhancement] Disable Takkuri Steal

mckinlee:
It just simply returns false in func_80C10B0C if the the checkbox is checked. Nothing fancy.

Put it in Time Savers section for now, not sure where this fits best.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2024757628.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2024773769.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2024843666.zip)


garrettjoecox (commented):
I think a new "difficulty options" section similar to SoH's might make sense to house this?

Additionally we should use the VB pattern to prevent this behavior. I would suggest adding a new VB enum for `GI_VB_THIEF_BIRD_STEAL`, and within thiefbird.c around line 588 wrap the `(Rand_ZeroOne() < 0.5f)` with a `GameInteractor_Should`. This will allow us to rely on the vanilla condition (50% randomness) and allow us to override it when we want

mckinlee:
> I think a new "difficulty options" section similar to SoH's might make sense to house this?
> 
> Additionally we should use the VB pattern to prevent this behavior. I would suggest adding a new VB enum for `GI_VB_THIEF_BIRD_STEAL`, and within thiefbird.c around line 588 wrap the `(Rand_ZeroOne() < 0.5f)` with a `GameInteractor_Should`. This will allow us to rely on the vanilla condition (50% randomness) and allow us to override it when we want

Thanks for the feedback. I've made those changes. :)

garrettjoecox (approved):
Perfect! Nice work

inspectredc (approved):
LGTM!

mm/2s2h/GameInteractor/GameInteractor.h
```diff
@@ -63,6 +63,7 @@ typedef enum {
     VB_CLOCK_TOWER_OPENING_CONSIDER_THIS_FIRST_CYCLE,
     VB_DRAW_SLIME_BODY_ITEM,
     VB_ZTARGET_SPEED_CHECK,
+    GI_VB_THIEF_BIRD_STEAL,
```
Archez: We recently dropped the `GI_` prefix to all these enums, just need to update it here and the other files.

```suggestion
    VB_THIEF_BIRD_STEAL,
```

mm/2s2h/Enhancements/DifficultyOptions/DisableTakkuriSteal.cpp
```diff
@@ -0,0 +1,10 @@
+#include <libultraship/bridge.h>
+#include "2s2h/GameInteractor/GameInteractor.h"
+
+void RegisterDisableTakkuriSteal() {
+    REGISTER_VB_SHOULD(GI_VB_THIEF_BIRD_STEAL, {
```
Archez: Since you have this register func called whenever the checkbox state is changed, it would be great if you could capture the hook ID from this register call and add logic to unregister the hook. This is in the effort to be efficient about only having hooks registered when expected to run.

This also allows you to drop the CVarGet from the inside of the hook body and instead have it as a one-time cost at time of registration, since you know the hook is only registered if the CVar is toggled on.

You can look at `Restorations/WoodfallMountainAppearance.cpp` for an example of what I mean

Archez (commented):
Just a few more things, then this should be good (pending any request to move the menu option)

mm/2s2h/GameInteractor/GameInteractor.h
mckinlee: Dropped the GI_

mm/2s2h/Enhancements/DifficultyOptions/DisableTakkuriSteal.cpp
mckinlee: Done

Archez (approved):

mm/2s2h/Enhancements/DifficultyOptions/DisableTakkuriSteal.h
```diff
@@ -0,0 +1,6 @@
+#ifndef DISABLE_TAKKURI_STEAL_H
+#define DISABLE_TAKKURI_STEAL_H
+
+void RegisterDisableTakkuriSteal();
+
+#endif // DISABLE_TAKKURI_STEAL_H
```
garrettjoecox: Looks like you accidentally left this file around, shouldn't be necessary now with the `DifficultyOptions.h`

mckinlee: Yep, sure did. Removed. :)

PR #779, opened 2024-09-24
Add swordsman school options

garrettjoecox:
null

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2106278049.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2106293415.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2106359354.zip)


Archez:
Mind including a description for this? Trying to understand why there are two different options in this PR. What does "finish early" mean?

garrettjoecox:
> Mind including a description for this? Trying to understand why there are two different options in this PR. What does "finish early" mean?

So the score is pretty self explanatory, you have to cut every log with a jump slash to get the perfect score of 30, you can dial that down to like 20 to give yourself some wiggle room for misses, but the minigame has a static amount of logs that come up, and if you reach the target score, you still have to wait till the rest of the logs come till it ends. This is especially bad when you set the score to like 1, you then have to wait for a while for the rest of the logs to go up then down. These could be combined to simplify the user experience I suppose

garrettjoecox:
Oh, it looks like I did add descriptions in the modern menu entries, just not on the legacy menu

balloondude2:
I think it might be better to combine the two options into one. I don't really see a situation where I'd want to lower the winning score and not finish early. 

That said, if you think there are users who would want a lower score requirement and still want all the logs to come out, I'd understand keeping the two options.

garrettjoecox:
Combined the options into one 👍

Archez (approved):

balloondude2 (approved):

PR #780, opened 2024-09-27
Extract pause maps and minimaps, support alt assets, fix pause map icons

Archez:
This PR makes it so we extract the pause menu maps and minimaps as actual textures instead of blobs. The map display now uses texture resource path strings instead of loading the raw bytes, allowing them to be overridden in texture packs and support HD sizes.

Also fixes missing door transition icons on the pause menu map due to a missing scene command callback.

image

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2025102389.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2025112530.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2025158127.zip)

### DIFF
--- a/mm/2s2h/z_scene_2SH.cpp
+++ b/mm/2s2h/z_scene_2SH.cpp
@@ -258,6 +258,7 @@ void Scene_CommandTransiActorList(PlayState* play, SOH::ISceneCommand* cmd) {
 
     play->transitionActors.count = list->numTransitionActors;
     play->transitionActors.list = (TransitionActorEntry*)list->GetRawPointer();
+    MapDisp_InitTransitionActorData(play, play->transitionActors.count, play->transitionActors.list);
 }
 
 void Scene_CommandEnvLightSettings(PlayState* play, SOH::ISceneCommand* cmd) {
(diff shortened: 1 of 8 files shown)


mm/2s2h/z_scene_2SH.cpp
```diff
@@ -258,6 +258,7 @@ void Scene_CommandTransiActorList(PlayState* play, SOH::ISceneCommand* cmd) {
 
     play->transitionActors.count = list->numTransitionActors;
     play->transitionActors.list = (TransitionActorEntry*)list->GetRawPointer();
+    MapDisp_InitTransitionActorData(play, play->transitionActors.count, play->transitionActors.list);
```
garrettjoecox, line 261: It looks like this was in the original `Scene_CommandTransitionActorList`, did we just miss it or was it not necessary?

garrettjoecox (approved):

mm/2s2h/z_scene_2SH.cpp
Archez: This is necessary for the white room transition squares and the boss door icon to show up correctly on the pause map. Seems like it was a miss from before.

PR #783, opened 2024-10-01
Add HD support for Igos room curtains

Archez:
This PR brings HD support to CPU modified textures used by the Igos curtains in the boss room fight.

Originally the game works by copying the texture data to the actor struct so that each curtain can have their own unique copy of the texture. As the curtain burns, the actor blacks out pixels in their copy. This texture is loaded as a segment value for the general display list.

To support HD textures and to avoid complex memory allocation/copying in the actor for different HD texture sizes, I've opted to leverage the existing Blended texture registration system which allows us to use a texture mask to avoid modifying the original texture. Because there are two curtains with separate mask, and the blended system only supports one mask per resource string, a new opcode was added to allow registering a blended texture as part of a graphics instruction instead of having to register in code before rendering.

In these demo shots, I have removed the lens flare and fire effect so it is easier to see the curtain itself

image

https://github.com/user-attachments/assets/17ff2ff2-fa83-4c0c-83a3-d713761bfae3

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2002666556.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2002685585.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2002737219.zip)

### DIFF
--- a/mm/src/overlays/actors/ovl_Boss_06/z_boss_06.c
+++ b/mm/src/overlays/actors/ovl_Boss_06/z_boss_06.c
@@ -10,7 +10,8 @@
 #include "overlays/actors/ovl_En_Knight/z_en_knight.h"
 #include "objects/gameplay_keep/gameplay_keep.h"
 #include "objects/object_knight/object_knight.h"
-#include "BenPort.h"
+
+#include "2s2h/BenPort.h"
 #include "2s2h/Enhancements/FrameInterpolation/FrameInterpolation.h"
 
 #define FLAGS (ACTOR_FLAG_TARGETABLE | ACTOR_FLAG_UNFRIENDLY | ACTOR_FLAG_10 | ACTOR_FLAG_20)
@@ -493,6 +494,9 @@ void Boss06_Update(Actor* thisx, PlayState* play) {
                 } else {
                     this->unk_200[idx] &= 0xF;
                 }
+
+                // 2S2H [Port] Map the coordinates to the blended mask
+                this->curtainTexMask[((s32)sp7C.y * 64) + x] = true;
             }
         }
     }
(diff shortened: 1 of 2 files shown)


mm/src/overlays/actors/ovl_Boss_06/z_boss_06.c
```diff
@@ -588,7 +592,19 @@ void Boss06_Draw(Actor* thisx, PlayState* play2) {
     }
 
     if (this->unk_144 & 1) {
-        gSPSegment(POLY_OPA_DISP++, 0x08, Lib_SegmentedToVirtual(&this->unk_200));
+        // 2S2H [Port][HD Textures] To support CPU modified textures for any arbitrary texture size and alt asset
+        // toggling, we need to set the original texture resource as the segment value and register each curtain's mask
+        // as a blended texture. The blended tex opcode is necessary since there are two curtains in the scene, but only
+        // one mask can be registered to a resource at a time.
+        if (true) {
```
garrettjoecox, line 599: Did you mean for this to be `true`? If so, and that was just your way of preserving the original line, do you need the `gSPInvalidateTexCache` in the else?

garrettjoecox (approved):

mm/src/overlays/actors/ovl_Boss_06/z_boss_06.c
Archez: That was my intention yes, and if we were to ever want the other code path to run, the invalidate would be needed as well.

PR #790, opened 2024-10-02
Add initial Sail implementation

garrettjoecox:
A ground up rewrite from SoH's impl. Includes the GI Action system (which replaces some patterns we had on SoH and some patterns we were starting here for rando traps)

Intended to be used with the Sail companion app, or your own usage of the sail protocol.

### Companion App Downloads
- **Linux** — [AppImage](https://github.com/HarbourMasters/Sail/releases/download/nightly/Sail-x86_64.AppImage) (recommended) or the [raw binary](https://github.com/HarbourMasters/Sail/releases/download/nightly/Sail-linux-amd64.tar.gz)
- **macOS** — [universal `.app`](https://github.com/HarbourMasters/Sail/releases/download/nightly/Sail-macos-universal.zip)
- **Windows** — [portable `.exe`](https://github.com/HarbourMasters/Sail/releases/download/nightly/Sail-windows-amd64.exe)

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/2ship2harkinian/2ship2harkinian/actions/artifacts/9848365745.zip)
  - [2ship-mac.zip](https://nightly.link/2ship2harkinian/2ship2harkinian/actions/artifacts/9848309025.zip)
  - [2ship-windows.zip](https://nightly.link/2ship2harkinian/2ship2harkinian/actions/artifacts/9848184920.zip)

### DIFF
--- a/mm/CMakeLists.txt
+++ b/mm/CMakeLists.txt
@@ -113,6 +113,11 @@ endif()
 
 list(FILTER ship__ EXCLUDE REGEX "2s2h/Extractor/*")
 
+option(SKIP_NETWORKING "Skip networking support (Sail remote control), avoiding the SDL2_net 2.2.0 dependency" OFF)
+if (SKIP_NETWORKING)
+    list(FILTER ship__ EXCLUDE REGEX "2s2h/Network/*")
+endif()
+
 if(NOT CMAKE_SYSTEM_NAME MATCHES "NintendoSwitch|CafeOS")
     file(GLOB_RECURSE ship__Extractor RELATIVE ${CMAKE_CURRENT_SOURCE_DIR}
         "2s2h/Extractor/*.c"
@@ -223,8 +228,8 @@ endif()
 find_package(SDL2)
 set(SDL2-INCLUDE ${SDL2_INCLUDE_DIRS})
 
-if (BUILD_CROWD_CONTROL)
-    find_package(SDL2_net)
+if (NOT SKIP_NETWORKING)
+    find_package(SDL2_net 2.2.0 REQUIRED)
     set(SDL2-NET-INCLUDE ${SDL_NET_INCLUDE_DIRS})
 endif()
 
@@ -253,9 +258,7 @@ if (CMAKE_SYSTEM_NAME STREQUAL "Windows")
 		"$<$<CONFIG:Release>:"
 			"NDEBUG;"
 		">"
-           #"$<$<BOOL:${BUILD_CROWD_CONTROL}>:ENABLE_CROWD_CONTROL>"
 		"INCLUDE_GAME_PRINTF;"
-           #"ENABLE_CROWD_CONTROL;"
 		"F3DEX_GBI_2"
 		"UNICODE;"
 		"_UNICODE"
@@ -289,7 +292,6 @@ elseif ("${CMAKE_CXX_COMPILER_ID}" MATCHES "GNU|Clang|AppleClang")
 			"NDEBUG;"
 		">"
         "F3DEX_GBI_2;"
-		# "$<$<BOOL:${BUILD_CROWD_CONTROL}>:ENABLE_CROWD_CONTROL>"
 		"_CONSOLE;"
 		"_CRT_SECURE_NO_WARNINGS;"
 		"ENABLE_OPENGL;"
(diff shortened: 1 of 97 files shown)


mm/CMakeLists.txt
```diff
@@ -365,9 +373,15 @@ endif()
...
+if (BUILD_NETWORKING)
     find_package(SDL2_net)
-    set(SDL2-NET-INCLUDE ${SDL_NET_INCLUDE_DIRS})
+
+    if(NOT SDL2_net_FOUND)
+        message(STATUS "SDL2_net not found (it's possible the version installed is too old). Disabling BUILD_NETWORKING.")
+        set(BUILD_NETWORKING 0)
+    else()
+        set(SDL2-NET-INCLUDE ${SDL_NET_INCLUDE_DIRS})
+    endif()
```
Archez: I don't think we should be doing fallbacks like this. If a user sets the network flag, I think it would make more sense to have it error if they don't have the dependency.

In `find_package` you can set `2.2.0 REQUIRED` so that it throws human readable messaging about the dependency, instead of something more cryptic in the build process.

mm/2s2h/Network/Sail.cpp
```diff
@@ -0,0 +1,292 @@
...
+
+    static std::string ip = CVarGetString("gNetwork.Sail.Host", "127.0.0.1");
+    static uint16_t port = CVarGetInteger("gNetwork.Sail.Port", 43384);
+    bool isFormValid = !isStringEmpty(CVarGetString("gNetwork.Sail.Host", "127.0.0.1")) && port > 1024 && port < 65535;
+
+    ImGui::BeginDisabled(isEnabled);
+
+    ImGui::Text("Remote IP & Port");
+    UIWidgets::PushStyleSlider();
+    if (ImGui::InputText("##gNetwork.Sail.Host", (char*)ip.c_str(), ip.capacity() + 1)) {
```
Archez: This appears to be an illegal use of std::string with `ImGui::InputText`
Please refer to the comment for ImGui about this:
```
if you want to use InputText() with std::string or any custom dynamic string type, see misc/cpp/imgui_stdlib.h and comments in imgui_demo.cpp
```

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -1533,6 +1619,43 @@ void SearchMenuGetItem(widgetInfo& widget) {
...
                 }
                 break;
+            case WIDGET_CVAR_INPUT_TEXT: {
+                ImGui::BeginDisabled(disabledValue);
+                UIWidgets::PushStyleSlider(menuTheme[menuThemeIndex]);
+                ImGui::PushItemWidth(ImGui::GetWindowWidth());
+                ImGui::Text(widget.widgetName.c_str());
+                std::string cvarValue =
+                    CVarGetString(widget.widgetCVar, std::get<const char*>(widget.widgetOptions.defaultVariant));
+                if (ImGui::InputText(("##" + widget.widgetName).c_str(), (char*)cvarValue.c_str(),
```
Archez: Same thing here

mm/2s2h/Network/Sail.cpp
garrettjoecox: Should be good now 👍

mm/2s2h/BenGui/SearchableMenuItems.h
```diff
@@ -347,7 +353,16 @@ static std::map<DisableOption, disabledInfo> disabledMap = {
...
+        "Intro Skip Not Selected" } },
+    { DISABLE_FOR_SAIL_FORM_INVALID,
+      { [](disabledInfo& info) -> bool {
+           return !(!isStringEmpty(CVarGetString("gNetwork.Sail.Host", "127.0.0.1")) &&
+                    CVarGetInteger("gNetwork.Sail.Port", 43384) > 1024 &&
+                    CVarGetInteger("gNetwork.Sail.Port", 43384) < 65535);
+       },
+        "Invalid Host/Port" } },
+    { DISABLE_FOR_SAIL_ENABLED,
+      { [](disabledInfo& info) -> bool { return Sail::Instance->isEnabled; }, "Sail is Enabled" } },
```
Archez: This use of `Sail::` needs to be wrapped in an ifdef to prevent compile errors with networking off. Not sure if it makes sense to just ifdef out the two disable options all together, or just force the return value to false.

garrettjoecox: I think what I'm going to instead do is only #ifdef the SDL touch points within network.cpp, and leave the rest in place leaving it visually disabled without `BUILD_NETWORKING`

garrettjoecox:
This has been updated, and some patterns tweaked

PR #791, opened 2024-10-03
[Enhancement] Keep Express Mail in the same cycle

Archez:
This PR is an alternate solution for #732 that aims to just keep the Express Mail item in the player's inventory when trading it the first time to either Postman/Madame Aroma. This works by checking who the letter is being traded to, and if the opposite actors trade cycle flag is set or not. Only if it is the first trade will the Mail be kept, otherwise it will disappear.

This way the player is still expected to interact with both NPCs and the item gives are untouched, keeping the enhancement compatible with a future rando.

Leverages a new VB hook in the msg script event system to override item delete commands.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2019331618.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2019334269.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2019345782.zip)


Eblo:
I prefer this approach over what I did in #732 because this ensures the player actually gets to engage with the content from both routes.

garrettjoecox (approved):
Looks like you'll need to account for the header changes but otherwise LGTM

PR #794, opened 2024-10-04
Fix authentic overflowed textures

Archez:
This PR is similar to a fix that SoH has which attempts to fix some authentic overflowed/incorrect texture usages which impacts HD support friendliness.

The following items are fixed in this:
* Freezard's body texture
* Arrow tips shiny texture
* Iron Knuckles shiny texture

The graphics patching updates the corresponding DLs to use the correct format and/or size that matches the real texture data.

Opening as a draft right now until I decide if this should be a user facing toggle, or if it should be defaulted on.
In SoH we did a solution which extracted "secondary" textures that had the overflow accounted for in the data, but I'm not entirely sure that covers all these use cases "accurately".

Additionally, in 2ship we have already handled the deku stick texture overflow in a different way, so that probably should be aligned with this.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2031835192.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2031844627.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2031882375.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenuBar.cpp
+++ b/mm/2s2h/BenGui/BenMenuBar.cpp
@@ -8,6 +8,7 @@
 #include <unordered_map>
 #include <string>
 #include "2s2h/Enhancements/Enhancements.h"
+#include "2s2h/Enhancements/GfxPatcher/AuthenticGfxPatches.h"
 #include "2s2h/DeveloperTools/DeveloperTools.h"
 #include "HudEditor.h"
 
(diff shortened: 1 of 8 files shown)


garrettjoecox:
> Opening as a draft right now until I decide if this should be a user facing toggle, or if it should be defaulted on.

What's closer to hardware, visually? Unless it's causing crashes, in which case I'd default them to on.

If it's required for HD textures, can we do a `CVAR || resource->custom` check?

Archez:
Both options support HD textures. It just means that texture pack makers have to make replacements for both if they want both states of the toggles supported. One that is just the texture itself, and one that also has overflowed data on it.

The overflowed data is technically authentic to hardware, which would be toggle == off. I just can't seem to get things to actually look like emulator (unless emulator is incorrect..)

GhostlyDark:
The effects looking off is a general issue with SoH and 2Ship (which includes Rupees, Red Ice...).

It doesn't seem like https://github.com/HarbourMasters/Shipwright/issues/38 has been fully addressed.

GLideN64, rt64, Angrylion Plus and Parallel RDP can be used as a reference of how the effects are supposed to look like - they are all accurate to real hardware in this regard.

GhostlyDark:
Is the toggle supposed to work already? It seems to be always on, even after restarting 2Ship.

The Freezard and Iron Knuckle textures look significantly better this way, so I'm happy with the results.

Archez:
This is ready for review. Got confirmation for Ghostly that things worked for both toggle states regarding HD replacements.

garrettjoecox (approved):

PR #797, opened 2024-10-05
Use glob_recurse across all of 2s2h/

garrettjoecox:
Primarily this is so we don't have to touch cmake all the time when adding new subdirectories to `2s2h/`

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2019174325.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2019177151.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2019185912.zip)

### DIFF
--- a/mm/CMakeLists.txt
+++ b/mm/CMakeLists.txt
@@ -120,7 +120,12 @@ source_group("include" FILES ${Header_Files__include})
 # }}}
 
 file(GLOB_RECURSE ship__ RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} "2s2h/*.c" "2s2h/*.cpp" "2s2h/*.h" "2s2h/*.hpp")
-source_group("2s2h" FILES ${ship__})
+# Create source groups that match the real file directory paths
+foreach(source IN LISTS ship__)
+    get_filename_component(source_path "${source}" PATH)
+    string(REPLACE "/" "\\" source_path_adjusted "${source_path}")
+    source_group("${source_path_adjusted}" FILES "${source}")
+endforeach()
 
 file(GLOB_RECURSE libultra_headers RELATIVE ${CMAKE_CURRENT_SOURCE_DIR}
     "include/PR/*.h"
--- a/mm/CMakeLists.txt
+++ b/mm/CMakeLists.txt
@@ -151,13 +151,6 @@ else()
     )
 endif()
 
-# 2s2h/resource {{{
-source_group("2s2h\\resource\\type" REGULAR_EXPRESSION "2s2h/resource/type/*")
-source_group("2s2h\\resource\\type\\scenecommand" REGULAR_EXPRESSION "2s2h/resource/type/scenecommand/*")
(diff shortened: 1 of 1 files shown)


mm/CMakeLists.txt
```diff
@@ -119,9 +119,8 @@ list(REMOVE_ITEM Header_Files__include ${CMAKE_CURRENT_SOURCE_DIR}/include/ultra
 source_group("include" FILES ${Header_Files__include})
 # }}}
 
-# m (root)
-file(GLOB soh__ RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} "2s2h/*.c" "2s2h/*.cpp" "2s2h/*.h" "2s2h/*.hpp")
-source_group("2s2h" FILES ${soh__})
+file(GLOB_RECURSE ship__ RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} "2s2h/*.c" "2s2h/*.cpp" "2s2h/*.h" "2s2h/*.hpp")
+source_group("2s2h" FILES ${ship__})
```
Archez: If you replace this with the following, it will make it so that the source groups match 1:1 with directory structure on disk. This would be useful for IDEs like Xcode/Visual Studio to not have all the files show up under one folder.

```suggestion
# Create source groups that match the real file directory paths
foreach(source IN LISTS ship__)
    get_filename_component(source_path "${source}" PATH)
    string(REPLACE "/" "\\" source_path_adjusted "${source_path}")
    source_group("${source_path_adjusted}" FILES "${source}")
endforeach()
```

mm/CMakeLists.txt
```diff
@@ -131,106 +130,23 @@ if (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC")
...
     )
-# }}}
 endif()
 
 # 2s2h/resource {{{
-file(GLOB_RECURSE soh__Resource RELATIVE ${CMAKE_CURRENT_SOURCE_DIR} "2s2h/resource/*.cpp" "2s2h/resource/*.h")
-
 source_group("2s2h\\resource\\type" REGULAR_EXPRESSION "2s2h/resource/type/*")
 source_group("2s2h\\resource\\type\\scenecommand" REGULAR_EXPRESSION "2s2h/resource/type/scenecommand/*")
 source_group("2s2h\\resource\\importer" REGULAR_EXPRESSION "2s2h/resource/importer/*")
```
Archez: These source groups for `resource/` can be dropped in favor of the new loop above

Archez (approved):

mm/CMakeLists.txt
garrettjoecox: Cal is going to love you for that one

Caladius: REALLY? dude this is incredible! Finally I'll be able to see the real structure and not a lump sum using 877-CASH-NOW!

PR #799, opened 2024-10-05
Move GameInteractor & NameTag src up to root of 2s2h

garrettjoecox:
~~Relies on #797~~

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2019270194.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2019273251.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2019282561.zip)


Archez (approved):
LGTM, will be nice to not establish the pattern of "everything goes under `Enhancements/`"

Also thank you for fixing up all the incorrect header paths that weren't using `2s2h/` as the prefix.

PR #806, opened 2024-10-11
Back-port GC credits easter egg with our own twist

garrettjoecox:
Submitting as draft for now, as it still needs some iteration

Todo/Notes: 
- [ ] Currently XLU and HUD still render on top of rect (archez halp pls)
- [ ] Should we have a selection between our credits and the original GC ones?
- Had to remove X & Y buttons from the original combination
-  Credits list currently derived from `git shortlog -sn 2332f63f5a9ca4b949f029c11a9744d601d7155f..HEAD`, should we do something else? this might be the easiest to keep up to date automatically
- Incompatible with debug mode

https://github.com/user-attachments/assets/ff7eaf70-9e77-4621-a9ff-613189455eda

### DIFF
--- a/mm/2s2h/Enhancements/Restorations/HiddenCredits.cpp
+++ b/mm/2s2h/Enhancements/Restorations/HiddenCredits.cpp
@@ -1,0 +1,201 @@
+#include <libultraship/bridge.h>
+#include "2s2h/GameInteractor/GameInteractor.h"
+#include "2s2h/Enhancements/FrameInterpolation/FrameInterpolation.h"
+#include <spdlog/spdlog.h>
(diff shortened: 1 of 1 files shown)


garrettjoecox:
Any thoughts/suggestions on if we want to move forward with this or some iteration of this? Otherwise I'll close it out

Archez:
I think I'm interested, and I have ideas on how to fix the graphics part of it. The only thing I don't know an elegant solution for is maintaining the committer list. Maybe need to use a CMake solution to automate it.

garrettjoecox:
Closing in favor #1162

PR #822, opened 2024-11-06
Archive version checking and regeneration

Archez:
This brings over the archive version checking and regeneration logic from SoH, slightly tweaked to account for OTR vs O2R differences.

The game archive will be checked in order of O2R -> ZIP -> OTR.
If it is a zip extension, then after the regeneration we rename the new .o2r back to .zip
If it is a otr extension, we delete the archive and then leave the resulting .o2r (this is because OTRExporter only exports as O2R)

2ship.o2r version checking is currently commented out because of an error in OTRExporter where the version file is filled with garbage. An OTRExporter change will be required first.

Also brought over the zip checking for Extract.cpp and trimmed out old remnants of OOT crcs and vanilla/master quest info.

Lastly updated building docs to prefer users use the built-in extractor and make cmake copy the asset xmls for Visual Studio.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2154525314.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2154527361.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2154539347.zip)

### DIFF
--- a/docs/BUILDING.md
+++ b/docs/BUILDING.md
@@ -22,7 +22,7 @@ It is recommended that you install Python and Git standalone, the install proces
 
 _Note: Be sure to either clone with the ``--recursive`` flag or do ``git submodule update --init`` after cloning to pull in the libultraship submodule!_
 
-2. Place one or more [compatible](#compatible-roms) roms in the `OTRExporter` directory with namings of your choice
+2. After setup and initial build, use the built-in O2R extraction to make your mm.o2r file.
 
 _Note: Instructions assume using powershell_
 ```powershell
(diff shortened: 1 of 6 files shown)


garrettjoecox:
> 2ship.o2r version checking is currently commented out because of an error in OTRExporter where the version file is filled with garbage. An OTRExporter change will be required first.

Will this block a release? Or are we fine with this for now

Archez:
> > 2ship.o2r version checking is currently commented out because of an error in OTRExporter where the version file is filled with garbage. An OTRExporter change will be required first.
> 
> Will this block a release? Or are we fine with this for now

It doesn't have to. We provide the 2ship.o2r with the downloads, so the only issue would be if someone updated without replacing that file (only applicable on Windows), they would crash because of new fonts required by the modern menu.

OTRExporter already has changes merged regarding the Audio PR, so trying to fix it would require making a new branch for OTRExporter or merging the Audio PR.

garrettjoecox:
> they would crash because of new fonts required by the modern menu.

Seems likely, lets make sure we note this in the release notes if it's not already included in update instructions somewhere, as well as letting helpers know.

This is the crash message for this for future reference:
```
[09:08:10.907] [ResourceLoader.cpp:81] [error] Failed to load resource: File not loaded
[09:08:10.907] [ResourceManager.cpp:161] [error] Failed to load resource file at path fonts/Inconsolata-Regular.ttf
```

garrettjoecox (approved):
Verified working on Windows & Mac

PR #828, opened 2024-11-10
Fix two-handed sword magic crash

### LINKED ISSUE #827: Two-handed Sword Spin Attack crash - Linux
### Description of the crash:
When "Two-Handed Sword Spin Attack" is enabled on the Linux appimage of 2Ship, it causes a crash when attempting a spin attack.

- [ ] No, I cannot reproduce this crash
- [x] Yes, here are the steps to reproduce the crash:

### Video:
https://github.com/user-attachments/assets/74d47248-ef11-4f74-893c-1f883420f4e9

### Logs:
[crash_deets.txt](https://github.com/user-attachments/files/17690118/crash_deets.txt)
(Abridged log.)

### System Information:
- OS: Garuda Linux Bird of Prey x86_64 
- GFX Backend: OpenGL
- Game Version: Satoko Alfa 1.1.0
- Rom Version: NTSC-U Gamecube

Eblo:
It turns out that some environments may not like it if you pass small numbers through `va_arg`. This fixes #827 by doing away with an argument and using the global `Player` instead.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2167715873.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2167717508.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2167725564.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Equipment/TwoHandedSwordSpinAttack.cpp
+++ b/mm/2s2h/Enhancements/Equipment/TwoHandedSwordSpinAttack.cpp
@@ -34,8 +34,8 @@ void RegisterTwoHandedSwordSpinAttack() {
     REGISTER_VB_SHOULD(VB_MAGIC_SPIN_ATTACK_CHECK_FORM, {
         if (CVarGetInteger("gEnhancements.Equipment.TwoHandedSwordSpinAttack", 0)) {
             // Additionally allow the Fierce Deity form to use charged spin attacks
-            PlayerTransformation form = va_arg(args, PlayerTransformation);
-            if (form == PLAYER_FORM_FIERCE_DEITY) {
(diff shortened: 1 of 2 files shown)


Malkierian (approved):

Archez (approved):
Turns out the crash was on the .cpp side for linux with `PlayerTransformation` not being a proper size of `int` for `va_args`

> warning: ‘PlayerTransformation’ is promoted to ‘int’ when passed through ‘...’
> note: (so you should pass ‘int’ not ‘PlayerTransformation’ to ‘va_arg’)

Removing the arg all together in this case like you did also works 👍

PR #846, opened 2024-11-12
Fix skip intro and skip first cycle logic running when it shouldn't

Archez:
The Skip Intro and Skip First Cycle enhancements were running every time a save was loaded from File Select that previously used the enhancements. This is because in `Sram_OpenSave` the following check is performed which decides where the player should load in.

```c
if (gSaveContext.save.isFirstCycle) {
    gSaveContext.save.entrance = ENTRANCE(SOUTH_CLOCK_TOWN, 0);
    gSaveContext.save.day = 0;
    gSaveContext.save.time = CLOCK_TIME(6, 0) - 1;
} else {
    gSaveContext.save.entrance = ENTRANCE(CUTSCENE, 0);
    gSaveContext.nextCutsceneIndex = 0;
    gSaveContext.save.playerForm = PLAYER_FORM_HUMAN;
}
```

What was happening is that `save.isFirstCycle` was not properly set by our enhancements, which meant the intro cutscene entrance is set, triggering our enhancements over and over.

This `save.isFirstCycle` is poorly named, it doesn't mean "currently in the first cycle", rather it means "link has entered clock town at least once". Basically it is just set once the game auto saves after entering clock town for the first time.

I've reworked both enhancements to bail out early if this flag is already set, and made it so the Skip Intro portion correctly sets the flag, while removing an incorrect unsetting of this flag in the Skip First Cycle portion.

Also removed needless conditional checks for setting flags since the operations will only run once.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2174379479.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2174381442.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2174384294.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Cutscenes/SkipIntroSequence.cpp
+++ b/mm/2s2h/Enhancements/Cutscenes/SkipIntroSequence.cpp
@@ -12,8 +12,10 @@ void Flags_SetWeekEventReg(s32 flag);
 void RegisterSkipIntroSequence() {
     REGISTER_VB_SHOULD(VB_PLAY_TRANSITION_CS, {
         // Intro cutscene
-        if (!(gSaveContext.save.entrance == ENTRANCE(CUTSCENE, 0) && gSaveContext.save.cutsceneIndex == 0))
+        if (!(gSaveContext.save.entrance == ENTRANCE(CUTSCENE, 0) && gSaveContext.save.cutsceneIndex == 0) ||
+            gSaveContext.save.isFirstCycle) {
             return;
+        }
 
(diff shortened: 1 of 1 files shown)


balloondude2 (approved):

garrettjoecox (approved):
Good find. Also ran into some hiccups from that bit in sram_opensave yesterday lol

PR #851, opened 2024-11-13
Fix crashing with modern menu and UIWidgets with no tooltips

Archez:
Doing a couple things in here, mainly found a crash scenario with the modern menu if tooltips aren't provided. There is no current instance of a widget without a tooltip, just figured I get this in since I ran into it with some dummy toggles I was playing around with.

I also exposed the `save.isFirstCycle` flag in the save editor in case we need it to help trouble shoot broken saves after the update (I think the saves should automatically fix themselves with the previous changes, but just in case).

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2185246612.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2185246617.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2185254126.zip)

### DIFF
--- a/mm/2s2h/BenGui/UIWidgets.cpp
+++ b/mm/2s2h/BenGui/UIWidgets.cpp
@@ -6,13 +6,9 @@
 #include <string>
 #include <unordered_map>
 #include <libultraship/libultra/types.h>
+#include "2s2h/ShipUtils.h"
 
 namespace UIWidgets {
-
-bool IsCStringEmpty(const char* str) {
-    return str == NULL || str[0] == '\0';
-}
-
 // Automatically adds newlines to break up text longer than a specified number of characters
 // Manually included newlines will still be respected and reset the line length
 // If line is midword when it hits the limit, text should break at the last encountered space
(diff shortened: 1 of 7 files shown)


Malkierian (approved):

mm/2s2h/BenGui/UIWidgets.cpp
```diff
@@ -8,6 +8,11 @@
 #include <libultraship/libultra/types.h>
 
 namespace UIWidgets {
+
```
garrettjoecox: ShipUtils is probably a better home for this

garrettjoecox (approved):

mm/2s2h/BenGui/UIWidgets.cpp
Archez: Yup, that makes sense.

PR #861, opened 2024-11-16
[Enhancement] Receive oceanside spider house wallet reward on any day

lscholte:
The wallet reward for clearing the oceanside spider house is normally only obtainable on the first day. This enhancement will allow the wallet reward to be collected on any day.

I'm also not sure if this would better in a different section. Perhaps restorations because this enhancement does recreate MM3D behaviour?

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2196890159.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2196890653.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2196898084.zip)

### DIFF
--- a/mm/src/overlays/actors/ovl_En_Sth/z_en_sth.c
+++ b/mm/src/overlays/actors/ovl_En_Sth/z_en_sth.c
@@ -369,34 +369,34 @@ void EnSth_HandleOceansideSpiderHouseConversation(EnSth* this, PlayState* play) 
                         // This flag prevents multiple rewards, switching to secondary dialogue after
                         SET_WEEKEVENTREG(WEEKEVENTREG_OCEANSIDE_SPIDER_HOUSE_COLLECTED_REWARD);
 
-                        switch (day) {
-                            case 0: // first day
-                                if (CHECK_WEEKEVENTREG(WEEKEVENTREG_RECEIVED_OCEANSIDE_WALLET_UPGRADE)) {
+                        if (!CHECK_WEEKEVENTREG(WEEKEVENTREG_RECEIVED_OCEANSIDE_WALLET_UPGRADE) &&
+                            (day == 0 || CVarGetInteger("gEnhancements.Cheats.OceansideWalletAnyDay", 0))) {
+                            // This flag prevents getting two wallets from the same place.
+                            //   Instead, getting a rupee award below.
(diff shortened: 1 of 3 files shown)


mm/src/overlays/actors/ovl_En_Sth/z_en_sth.c
```diff
@@ -369,34 +369,34 @@ void EnSth_HandleOceansideSpiderHouseConversation(EnSth* this, PlayState* play)
                         // This flag prevents multiple rewards, switching to secondary dialogue after
                         SET_WEEKEVENTREG(WEEKEVENTREG_OCEANSIDE_SPIDER_HOUSE_COLLECTED_REWARD);
 
-                        switch (day) {
-                            case 0: // first day
-                                if (CHECK_WEEKEVENTREG(WEEKEVENTREG_RECEIVED_OCEANSIDE_WALLET_UPGRADE)) {
+                        if (!CHECK_WEEKEVENTREG(WEEKEVENTREG_RECEIVED_OCEANSIDE_WALLET_UPGRADE) &&
```
Eblo, line 372: Instead of directly using the CVar in the actor file, you should create a new VB, wrap the affected logic with a `GameInteractor_Should` call, and handle the custom logic in a new enhancement file. One example you could look at is your very own `VB_DEKU_GUARD_SHOW_SEARCH_BALLS`.

Eblo (commented):
Same deal as your other PRs, needs hookified

> I'm also not sure if this would better in a different section. Perhaps restorations because this enhancement does recreate MM3D behaviour?

"Time Cycle" makes sense to me.

balloondude2:
Looks like you have some outstanding feedback. Are you planning on coming back to this?

lscholte:
> Looks like you have some outstanding feedback. Are you planning on coming back to this?

Yeah I did forget about this. I still think this is a worthwhile enhancement to add, but I personally don't believe I would be in a position to add and test this for the next couple of months. But the good news is that @mckinlee has created https://github.com/HarbourMasters/2ship2harkinian/pull/1235 to move this along, so assuming that gets merged, then we should not need this pull request anymore.

Eblo:
Closing in favor of #1235.

PR #869, opened 2024-11-20
Update the labels of flag checkboxes to hex values

balloondude2:
This updates the flag checkbox labels for weekEventReg in the Save Editor to its mask in hex. This will make it easier to relate the Save Editor flags to the definitions in the code and entries from the Event Log.

hexMask

Also updates other flag checkboxes in the Save Editor > Flags tab and the Actor Viewer to display hex values, which I think would be more useful. 

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2219335566.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2219337567.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2219357865.zip)

### DIFF
--- a/mm/2s2h/BenGui/UIWidgets.cpp
+++ b/mm/2s2h/BenGui/UIWidgets.cpp
@@ -548,7 +548,6 @@ void DrawFlagArray32(const std::string& name, uint32_t& flags) {
     }
     ImGui::PopID();
 }
-// fmt::format("0x{:04X} ({})", (u16)actor->params, actor->params)
 void DrawFlagArray16(const std::string& name, uint16_t& flags) {
     ImGui::PushID(name.c_str());
     for (int16_t flagIndex = 0; flagIndex < 16; flagIndex++) {
--- a/mm/2s2h/BenGui/UIWidgets.cpp
+++ b/mm/2s2h/BenGui/UIWidgets.cpp
@@ -593,4 +593,27 @@ void DrawFlagArray8(const std::string& name, uint8_t& flags) {
     }
     ImGui::PopID();
 }
+
+void DrawFlagArray8Hex(const std::string& name, uint8_t& flags) {
+    ImGui::PushID(name.c_str());
+    for (int8_t flagIndex = 0; flagIndex < 8; flagIndex++) {
+        if ((flagIndex % 8) != 0) {
+            ImGui::SameLine();
+        }
+        ImGui::PushID(flagIndex);
+        uint8_t bitMask = 1 << flagIndex;
+        bool flag = (flags & bitMask) != 0;
+        std::string label = std::format("0x{:02x}", bitMask);
+        if (UIWidgets::Checkbox(label.c_str(), &flag,
+                                { .tooltip = label.c_str(), .labelPosition = LabelPosition::None })) {
+            if (flag) {
+                flags |= bitMask;
+            } else {
+                flags &= ~bitMask;
+            }
+        }
+        ImGui::PopID();
+    }
+    ImGui::PopID();
+}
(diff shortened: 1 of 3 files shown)


garrettjoecox (approved):

mm/2s2h/BenGui/UIWidgets.cpp
```diff
@@ -534,7 +534,7 @@ void DrawFlagArray32(const std::string& name, uint32_t& flags) {
         ImGui::PushID(flagIndex);
         uint32_t bitMask = 1 << flagIndex;
         bool flag = (flags & bitMask) != 0;
-        std::string label = std::to_string(flagIndex);
+        std::string label = std::format("0x{:02x}", flagIndex);
```
Archez: (build errors) Use `fmt::format` from `#include <spdlog/fmt/fmt.h>`, refer to actor viewer for an example. Also using capital X ensures the A-F parts are capital in the tooltip, I'd prefer that.

```suggestion
        std::string label = fmt::format("0x{:02X}", flagIndex);
```

mm/2s2h/BenGui/UIWidgets.hpp
```diff
@@ -417,6 +417,7 @@ namespace UIWidgets {
     void DrawFlagArray32(const std::string& name, uint32_t& flags);
     void DrawFlagArray16(const std::string& name, uint16_t& flags);
     void DrawFlagArray8(const std::string& name, uint8_t& flags);
+    void DrawFlagArray8Mask(const std::string& name, uint8_t& flags);
```
Archez, line 420: So what is the difference between `DrawFlagArray8Mask` and `DrawFlagArray8`? Seems like you modified both to do the same.

Archez (commented):
Should we consider displaying both values, Hex and Dec together?

mm/2s2h/BenGui/UIWidgets.hpp
balloondude2: `DrawFlagArray8Mask` will label like: `0x01`, `0x02`, `0x04`, `0x08`, `0x10`, etc.
`DrawFlagArray8` will label like: `0x01`, `0x02`, `0x03`, `0x04`, `0x05`, etc. Since this only goes up to 8, it's not really different from before. It's just adding a (kind of unnecessary) 0x0 prefix. I changed it to be consistent with the changes to `DrawFlagArray32` and `DrawFlagArray16`

Archez: Ah I see, I missed that distinction when glancing at the diff.

mm/2s2h/BenGui/UIWidgets.cpp
balloondude2: Pushed up a fix for this and added displaying both values.

mm/2s2h/BenGui/UIWidgets.cpp
```diff
@@ -547,7 +548,7 @@ void DrawFlagArray32(const std::string& name, uint32_t& flags) {
     }
     ImGui::PopID();
 }
-
+// fmt::format("0x{:04X} ({})", (u16)actor->params, actor->params)
```
Archez: Oops
```suggestion

```

balloondude2: I got the line added back in. I think the formatting should be good now 🫠

Archez (approved):

PR #870, opened 2024-11-23
[PoC] Mouse Gyro support (Primarily for steam deck)

garrettjoecox:
This injects the mouse delta as a gyro input, allowing the steam deck Gyro functionality to be used with our gyro functionality. Currently the window now captures the mouse unless the new modern menu is open. You would need to opt into this behavior.

The PR itself is hacky, I'd be interested in hearing what this may look like to get truly integrated into LUS cc @Archez @Kenix3 @briaguya-ai 

https://github.com/user-attachments/assets/74550a34-6d0a-4e5c-a644-8b25e76a1005

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2297158169.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2297161946.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2297168662.zip)

### DIFF
--- a/mm/2s2h/BenGui/Menu.cpp
+++ b/mm/2s2h/BenGui/Menu.cpp
@@ -134,8 +134,10 @@ bool ModernMenuHeaderEntry(std::string label) {
 
 void BenMenu::Draw() {
     if (!IsVisible()) {
+        Ship::Context::GetInstance()->GetWindow()->SetMouseCapture(true);
         return;
     }
+    Ship::Context::GetInstance()->GetWindow()->SetMouseCapture(false);
     DrawElement();
     // Sync up the IsVisible flag if it was changed by ImGui
     SyncVisibilityConsoleVariable();
--- a/mm/2s2h/BenPort.cpp
+++ b/mm/2s2h/BenPort.cpp
@@ -1766,3 +1766,9 @@ extern "C" int Controller_ShouldRumble(size_t slot) {
 
     return 0;
 }
+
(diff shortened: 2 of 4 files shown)


briaguya0:
i see there being a few options here:
* hackiest: keeping this as-is so only `ControlDeck` and `Controller` are modified - i don't think that's a good idea to merge
* kinda hacky: no new files, modify `ControllerGyro::UpdatePad` for reading the mouse data, not sure where setting relative mouse mode goes
* least hacky: make a new class that inherits from `ControllerGyroMapping` (possibly rename `SDLGyroMapping` to `SDLControllerGyroMapping` and make a new `SDLMouseGyroMapping`), not sure where setting relative mouse mode goes, would also need to update the input editor(s) to support setting mouse as gyro
  * with this in place we could check to see if we have an `SDLMouseGyroMapping` to determine if we need to run any slightly hacky mouse logic that needs to live in somewhat less than ideal places
  * another big question here is does SDL mouse work with dxgi? if not we'd either need to make another class for dxgi or we'd need to add logic to ensure none of it shows up/is run when we aren't in SDL land

garrettjoecox:
Closing this out, this should be possible now using our new mouse API interfaces from LUS, but I'll let someone else take that on. For now you can use the "Mouse Support" build under #2s2h-modding https://discord.com/channels/808039310850130000/1255290916311400563

PR #871, opened 2024-11-23
Implement interpolated skybox stars

### LINKED ISSUE #492: Stars are not interpolated
Stars use a method of transforms that we cannot easily interpolate. Someone else more knowledgeable can probably fill out more details here

Archez:
This adds an implementation to interpolate the skybox stars. The stars are originally drawn using texture rectangles, which are not possible to interpolate. Instead we have to draw the stars as quads against the projection matrix to take advantage of the matrix interpolation system.

The stars have "real" world positions which we can use to position a matrix, and then scale to match the shape of each star, then we apply the billboard matrix rotation so that they always face the camera. We have to perform additional math on the star positions to bring the stars "closer" to the camera, as some of them are beyond the Far clipping value of the projection matrix (causing them to not render).

The interpolated stars flow will only execute when a FPS setting above 20 is used, otherwise the original stars are rendered.

I did my best to calculate the math such that the interpolated stars are as close as possible in size/position to how the original stars were. Any remaining differences are largely just from texture rectangles "snapping" to whole integer values, where as quads have more precision.

https://github.com/user-attachments/assets/7cf2e24c-2c04-4c2b-9072-831621dc1344

There is an additional bug fix to correct missing alpha fading on the stars when they appear/disappear. This required a similar "fix" that the bombers notebook needed for some of its elements.

Closes #492 

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2229123888.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2229124641.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2229126539.zip)

### DIFF
--- a/mm/2s2h/BenPort.cpp
+++ b/mm/2s2h/BenPort.cpp
@@ -269,6 +269,10 @@ uint32_t OTRGlobals::GetInterpolationFPS() {
                               CVarGetInteger("gInterpolationFPS", 20));
 }
 
+extern "C" uint32_t Ship_GetInterpolationFPS() {
+    return OTRGlobals::Instance->GetInterpolationFPS();
+}
+
 struct ExtensionEntry {
     std::string path;
     std::string ext;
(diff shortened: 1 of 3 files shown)


Archez:
I could also see this worth being pointed to Satoko instead so that it gets release sooner.

Edit: On second thought, I think this should go into Satoko, as I plan to make a couple other quick fixes.

PR #876, opened 2024-11-25
Add better song of double time enhancement

garrettjoecox:
Allows you to advance to any time up until an hour prior to moon fall. In my testing some of the actor respawns look a little weird when in the frozen state, but snap into place when resuming time. Works with all 3 clock types.

todo: 
- [x] Render the analog stick graphic from shops, and maybe flash the clock to make it more obvious what's happening

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2283288496.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2283291743.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2283303178.zip)


mm/2s2h/Enhancements/Graphics/3DSClock.cpp
```diff
@@ -77,7 +77,7 @@ void Register3DSClock() {
 
                     static s32 sFinalHoursIntro = 0;
 
-                    sThreeDayClockAlpha = gPlayState->interfaceCtx.aAlpha;
+                    sThreeDayClockAlpha = gPlayState->interfaceCtx.bAlpha;
```
garrettjoecox, line 80: The original clock uses `bAlpha`, so I changed this to also use it

lscholte:
I was just trying this out because I was curious what it would look like. I like what you've done from a gameplay perspective, but one thing that might be awkward is getting to the exact time that I want as I find the analog stick is a bit too sensitive. So if I am say trying to travel to a specific time, I might overshoot and have to go backwards, but might also overshoot that as well. What you have now is great for doing large time jumps though.

I'm not sure how to make this better UX-wise. Perhaps a modifier button like Z to make it a bit less sensitive to the analog stick input?

garrettjoecox:
> Perhaps a modifier button like Z to make it a bit less sensitive to the analog stick input?

I like that idea, I have incorporated this. It's currently not obvious that is an option but I'll probably do another pass on the "UI" of this feature in the future and account for it.

PR #886, opened 2024-12-06
[Cosmetic Editor] Implements Initial Cosmetic Editor

Caladius:
Implements the Cosmetic Editor along with the following options available for modification:

- Normal Heart Fill
- Normal Beating Heart Fill
- Normal Heart Border
- Double Defense Heart Fill
- Double Defense Beating Heart Fill
- Double Defense Heart Border
- Small Key HUD Color
- Rupee Icon HUD Color
- Spin Slash Charge & Burst Colors
- Great Spin Slash Charge & Burst Colors
- Fire, Ice, Light Arrow Primary and Secondary Colors
- Item Button Colors (A, B, C Buttons, D-Pad, Start)
- File Select Window and Plate Colors

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2287487560.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2287488706.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2287495520.zip)


mm/src/overlays/actors/ovl_Arrow_Light/z_arrow_light.c
```diff
@@ -205,8 +206,10 @@ void ArrowLight_Draw(Actor* thisx, PlayState* play) {
 
         Gfx_SetupDL25_Xlu(play->state.gfxCtx);
 
-        gDPSetPrimColor(POLY_XLU_DISP++, 0x80, 0x80, 255, 255, 170, this->alpha);
-        gDPSetEnvColor(POLY_XLU_DISP++, 255, 255, 0, 128);
+        gDPSetPrimColorWithOverride(POLY_XLU_DISP++, 0x80, 0x80, 255, 255, 170, this->alpha, "Effects.LightArrowPrim");
+        gDPSetEnvColorWithOverride(POLY_XLU_DISP++, 255, 255, 0, 128, "Effects.LightArrowSec");
+        // gDPSetPrimColor(POLY_XLU_DISP++, 0x80, 0x80, 255, 255, 170, this->alpha);
+        // gDPSetEnvColor(POLY_XLU_DISP++, 255, 255, 0, 128);
```
garrettjoecox:
```suggestion
```

mm/src/overlays/actors/ovl_Arrow_Fire/z_arrow_fire.c
```diff
@@ -268,9 +269,8 @@ void ArrowFire_Draw(Actor* thisx, PlayState* play) {
             gSPDisplayList(POLY_XLU_DISP++, D_0E000000_TO_SEGMENTED(fillRect));
         }
         Gfx_SetupDL25_Xlu(play->state.gfxCtx);
-
-        gDPSetPrimColor(POLY_XLU_DISP++, 0x80, 0x80, 255, 200, 0, this->alpha);
-        gDPSetEnvColor(POLY_XLU_DISP++, 255, 0, 0, 128);
+        gDPSetPrimColorWithOverride(POLY_XLU_DISP++, 0x80, 0x80, 255, 200, 0, this->alpha, "Effects.FireArrowPrim");
```
garrettjoecox: (re-add whitespace)
```suggestion

        gDPSetPrimColorWithOverride(POLY_XLU_DISP++, 0x80, 0x80, 255, 200, 0, this->alpha, "Effects.FireArrowPrim");
```

mm/src/code/z_parameter.c
```diff
@@ -5970,7 +5981,8 @@ void Interface_DrawAButton(PlayState* play) {
     gDPPipeSync(OVERLAY_DISP++);
     Interface_SetPerspectiveView(play, 23 + R_A_BTN_Y_OFFSET, 68 + R_A_BTN_Y_OFFSET, 190, 235);
     gSPVertex(OVERLAY_DISP++, &interfaceCtx->actionVtx[0], 4, 0);
-    gDPSetPrimColor(OVERLAY_DISP++, 0, 0, 100, 200, 255, interfaceCtx->aAlpha);
+    gDPSetPrimColorWithOverride(OVERLAY_DISP++, 0, 0, 100, 200, 255, interfaceCtx->aAlpha, "Buttons.A");
+    // gDPSetPrimColor(OVERLAY_DISP++, 0, 0, 100, 200, 255, interfaceCtx->aAlpha);
```
garrettjoecox:
```suggestion
```

mm/src/code/z_parameter.c
```diff
@@ -19,9 +19,13 @@
 #include "BenPort.h"
 #include <string.h>
 #include "BenGui/HudEditor.h"
+#include "BenGui/CosmeticEditor.h"
 #include "2s2h_assets.h"
 #include "2s2h/GameInteractor/GameInteractor.h"
 
+// 2S2H [Port] Used for modifying Button Colors in CosmeticEditor.cpp
+Color_RGBA8 modifiedColor;
```
garrettjoecox: I don't see a lot of benefit from re-using this, but I won't fight it.

If we do keep it, you should make it `static` so if someone else tries to add their own `modifiedColor` somewhere else in the codebase it doesn't conflict with this

mm/2s2h/BenGui/HudEditor.cpp
```diff
@@ -148,6 +148,21 @@ extern "C" void HudEditor_ModifyDrawValues(s16* rectLeft, s16* rectTop, s16* rec
...
+//         Color_RGBA8 changedColor = CVarGetColor((cvarString + ".Color").c_str(), {});
+//         r = changedColor.r;
+//         g = changedColor.g;
+//         b = changedColor.b;
+//         a = changedColor.a;
+//     }
+//
+//     gDPSetPrimColor(pkt, m, l, r, g, b, a);
+// }
+
```
garrettjoecox:
```suggestion
```

mm/2s2h/BenGui/HudEditor.h
```diff
@@ -66,6 +66,7 @@ void HudEditor_ModifyKaleidoEquipAnimValues(s16* ulx, s16* uly, s16* shrinkRate)
 void HudEditor_ModifyDrawValuesFromBase(s16 baseX, s16 baseY, s16* rectLeft, s16* rectTop, s16* rectWidth,
                                         s16* rectHeight, s16* dsdx, s16* dtdy);
 void HudEditor_ModifyDrawValues(s16* rectLeft, s16* rectTop, s16* rectWidth, s16* rectHeight, s16* dsdx, s16* dtdy);
+// void gDPSetPrimColorWithOverride(Gfx* pkt, u8 m, u8 l, u8 r, u8 g, u8 b, u8 a, const char* cvar);
```
garrettjoecox:
```suggestion
```

mm/src/code/z_parameter.c
```diff
@@ -5270,44 +5275,50 @@ void Interface_DrawItemButtons(PlayState* play) {
...
     // #endregion
 
     // B Button Color & Texture
     HudEditor_SetActiveElement(HUD_EDITOR_ELEMENT_B);
+    modifiedColor = CosmeticEditor_getChangedColor(100, 255, 120, interfaceCtx->bAlpha, "Buttons.B");
     OVERLAY_DISP = Gfx_DrawTexRectIA8_DropShadow(
         OVERLAY_DISP, gButtonBackgroundTex, 0x20, 0x20, D_801BF9D4[EQUIP_SLOT_B], D_801BF9DC[EQUIP_SLOT_B],
         D_801BFAF4[EQUIP_SLOT_B], D_801BFAF4[EQUIP_SLOT_B], D_801BF9E4[EQUIP_SLOT_B] * 2, D_801BF9E4[EQUIP_SLOT_B] * 2,
-        100, 255, 120, interfaceCtx->bAlpha);
+        modifiedColor.r, modifiedColor.g, modifiedColor.b, interfaceCtx->bAlpha);
```
garrettjoecox: As I mentioned in voice, another way to handle these is to pass in a new `cvar` argument to `Gfx_DrawTexRectIA8_DropShadow`. Don't have a strong preference but it would prevent the need to declare modifiedColor and replace the vanilla values here.

garrettjoecox (approved):
noice

mm/src/code/z_parameter.c
Archez: Agreed, it doesn't look like we are using the assigned value from one function scope into another scope,

If it had to stay file level scope, then we should also prefix it with `s` -> `sModifiedColor`

Caladius: Added `static` and prefixed `sModifiedColor`

mm/src/code/z_parameter.c
Caladius: slightly daunting with this one, if you have a few minutes later today maybe we can run through it quickly? My solo attempt failed miserably haha.

Archez: Alternatively, instead of adding an arg to the existing func, you could make a wrapped func that deals with fetching color by cvar and then passing it to the original func. This way we aren't passing dummy arguments for things we don't want to change color for.

Caladius: modified, hopefully this was what you were after

Archez:
After sleeping on this, I think this pattern is the most ideal for us to move forward with. I haven't looked at any of the UI yet, but we can iterate on that as the menu stuff matures.

When I find some time, I'd like to update this PR first to replace the need to pass in the CVar name to the `XYZOverride` funcs and instead leverage the existing `CosmeticEditorElementID` enums that you made. These enums give us access to the elements properties anyways, similar to how the HUD Editor works. Avoids typos being a problem and such. 

e.g.
```c
gDPSetPrimColorWithOverride(OVERLAY_DISP++, 0, 0, interfaceCtx->heartsPrimR[0],
                            interfaceCtx->heartsPrimG[0], interfaceCtx->heartsPrimB[0],
                            interfaceCtx->healthAlpha, COSMETIC_ELEMENT_HEART_NORMAL);

Archez (approved):
Good to go as is for an introduction to cosmetic edits. I think the UI could use improvements, but that can be iterated on 👍

PR #889, opened 2024-12-08
Add extended culling handling for special actors

Archez:
Applies extended culling rules to various actors that implement their own culling/fading logic.

This includes the following:
* Grass units (the groups in Termina field etc)
* Keaton grass
* Liftable rocks and silver boulders
* Bronze boulder
* Bombable boulder
* Rock circles
* Butterfly/insect spawners

This works by creating and using the following methods: `Ship_ExtendedCullingActorAdjustProjectedZ`, `Ship_ExtendedCullingActorAdjustProjectedX`, and `Ship_ExtendedCullingActorRestoreProjectedPos`, which temporarily modify the actor project pos values so that we can "trick" specific distance checks to account for widescreen/far distance without having to directly modify a bunch of src lines of code.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2300430268.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2300437830.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2300455014.zip)

### DIFF
--- a/mm/2s2h/ShipUtils.cpp
+++ b/mm/2s2h/ShipUtils.cpp
@@ -3,13 +3,52 @@
 #include "assets/2s2h_assets.h"
 
 extern "C" {
+#include "z64.h"
+#include "functions.h"
 #include "macros.h"
+
+extern float OTRGetAspectRatio();
 
 extern f32 sNESFontWidths[160];
 extern const char* fontTbl[156];
 extern TexturePtr gItemIcons[131];
 extern TexturePtr gQuestIcons[14];
 extern TexturePtr gBombersNotebookPhotos[24];
+}
+
+constexpr f32 fourByThree = 4.0f / 3.0f;
+
+// Gets the additional ratio of the screen compared to the original 4:3 ratio, clamping to 1 if smaller
+extern "C" f32 Ship_GetExtendedAspectRatioMultiplier() {
+    f32 currentRatio = OTRGetAspectRatio();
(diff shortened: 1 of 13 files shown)


mm/2s2h/ShipUtils.cpp
```diff
@@ -3,15 +3,54 @@
...
+    if (CVarGetInteger("gEnhancements.Graphics.ActorCullingAccountsForWidescreen", 0)) {
+        f32 ratioAdjusted = Ship_GetExtendedAspectRatioMultiplier();
+        actor->projectedPos.x /= ratioAdjusted;
+    }
+}
+
+// Restores the projectedPos values on the actor after modifications from the Extended Culling hacks
+extern "C" void Ship_ExtendedCullingActorRestoreProjectedPos(PlayState* play, Actor* actor) {
+    f32 invW = 0.0f;
+    Actor_GetProjectedPos(play, &actor->world.pos, &actor->projectedPos, &invW);
```
garrettjoecox, line 51: I assume this originally called in `actor.c` prior to draw?

Archez: Not this one directly, but the same method inside it (the SkinMatrix_ one).
I could use that instead, just figured this one had a nicer name.

garrettjoecox (approved):
Noice

PR #890, opened 2024-12-10
[Enhancement] Persist entrance and respawn information in saves

Eblo:
Pause saves almost already worked well enough, but shared grottos and cutscene entrances made that a no-go. This PR addresses those hurdles.

- Add `RespawnData` to `ShipSaveInfo`, using it to persist `gSaveContext.respawn`. This affects multiple things. The most obvious is that it maintains respawn information, i.e. where a player will reload if they die. Less obvious is that this data is also used to handle shared grotto scenes, their chest contents, and their exit destinations. Persisting this allows pause saves to work from within grottos.
- To ensure that *only* a possible entrance cutscene gets skipped on load, I used `OnPassPlayerInputs` as the entry point for unregistering the entrance cutscene skip hook. This seemed like the best existing hook to say "the player has assumed control, so don't skip any more cutscenes now". I tested various combinations of loading an entrance cutscene save and trying other cutscenes, as well as loading a non-entrance cutscene save and trying other cutscenes. This seems to work as I intended.
- Add "Remember Save Location" setting that affects whether `SavingEnhancements_GetSaveEntrance` works as it did before (Clock Town, dungeon entrances) or uses the immediate last entrance.
- Added migration for old saves that do not have respawn information

This has worked remarkably well from testing, both via resets and cold boots. However, there are hundreds of entrances in the game, so I welcome any additional eyes on this.

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4675473476.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4675585254.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4676219685.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
+++ b/mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
@@ -14,31 +14,41 @@ static uint64_t lastSaveTimestamp = GetUnixTimestamp();
 
 static HOOK_ID autosaveGameStateUpdateHookId = 0;
 static HOOK_ID autosaveGameStateDrawFinishHookId = 0;
+static HOOK_ID skipEntranceCutsceneHookId = 0;
+static HOOK_ID killSkipEntranceCutsceneHookId = 0;
 
 // Used for saving through Autosaves and Pause Menu saves.
 extern "C" int SavingEnhancements_GetSaveEntrance() {
-    switch (gPlayState->sceneId) {
-        // Woodfall Temple + Odolwa
-        case SCENE_MITURIN:
-        case SCENE_MITURIN_BS:
-            return ENTRANCE(WOODFALL_TEMPLE, 0);
-        // Snowhead Temple + Goht
-        case SCENE_HAKUGIN:
-        case SCENE_HAKUGIN_BS:
-            return ENTRANCE(SNOWHEAD_TEMPLE, 0);
-        // Great Bay Temple + Gyorg
-        case SCENE_SEA:
-        case SCENE_SEA_BS:
-            return ENTRANCE(GREAT_BAY_TEMPLE, 0);
-        // Stone Tower Temple
-        case SCENE_INISIE_N:
-            return ENTRANCE(STONE_TOWER_TEMPLE, 0);
-        // Stone Tower Temple (inverted) + Twinmold
-        case SCENE_INISIE_R:
-        case SCENE_INISIE_BS:
-            return ENTRANCE(STONE_TOWER_TEMPLE_INVERTED, 0);
-        default:
-            return ENTRANCE(SOUTH_CLOCK_TOWN, 0);
+    if (CVarGetInteger("gEnhancements.Saving.RememberSaveLocation", 0)) {
+        // Maintain respawn information, used for grottos
+        for (int i = 0; i < RESPAWN_MODE_MAX; i++) {
+            gSaveContext.save.shipSaveInfo.respawn[i] = gSaveContext.respawn[i];
+        }
+        return gSaveContext.save.entrance;
+    } else {
+        switch (gPlayState->sceneId) {
+            // Woodfall Temple + Odolwa
+            case SCENE_MITURIN:
+            case SCENE_MITURIN_BS:
+                return ENTRANCE(WOODFALL_TEMPLE, 0);
+            // Snowhead Temple + Goht
+            case SCENE_HAKUGIN:
+            case SCENE_HAKUGIN_BS:
+                return ENTRANCE(SNOWHEAD_TEMPLE, 0);
+            // Great Bay Temple + Gyorg
+            case SCENE_SEA:
+            case SCENE_SEA_BS:
+                return ENTRANCE(GREAT_BAY_TEMPLE, 0);
+            // Stone Tower Temple
+            case SCENE_INISIE_N:
+                return ENTRANCE(STONE_TOWER_TEMPLE, 0);
+            // Stone Tower Temple (inverted) + Twinmold
+            case SCENE_INISIE_R:
+            case SCENE_INISIE_BS:
+                return ENTRANCE(STONE_TOWER_TEMPLE, 0);
+            default:
+                return ENTRANCE(SOUTH_CLOCK_TOWN, 0);
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


Eblo:
With rando merged into develop, I rebased this and gave it another once over. I think it's in a better state for review now.

Archez:
Haven't review yet, but I think it might be nice to retain the previous behavior (dungeons + clock town), and then have a toggle for the new behavior similar to the "remember save location" from SoH.

Eblo:
> Haven't review yet, but I think it might be nice to retain the previous behavior (dungeons + clock town), and then have a toggle for the new behavior similar to the "remember save location" from SoH.

Agreed, done.

l12436:
When I use this PR, the cut scene skipper seem cause the first intro scene being skip, Link will land in the scene that should not be to landed. the cut scene is successfully play when I remove this patch

Eblo:
Good catch. This code makes the assumption that any entrance cutscene immediately on load has already been seen and that the player saved after that scene, and thus skips it. I overlooked that a new game will open into a cutscene. I'll have to take another look at this. Thanks!

Eblo:

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
```diff
@@ -14,31 +14,41 @@ static uint64_t lastSaveTimestamp = GetUnixTimestamp();
...
+            case SCENE_SEA:
+            case SCENE_SEA_BS:
+                return ENTRANCE(GREAT_BAY_TEMPLE, 0);
+            // Stone Tower Temple
+            case SCENE_INISIE_N:
+                return ENTRANCE(STONE_TOWER_TEMPLE, 0);
+            // Stone Tower Temple (inverted) + Twinmold
+            case SCENE_INISIE_R:
+            case SCENE_INISIE_BS:
+                return ENTRANCE(STONE_TOWER_TEMPLE, 0);
```
mckinlee: This seems unintentional, if so revert it back to STONE_TOWER_TEMPLE_INVERTED.

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
```diff
@@ -149,6 +159,53 @@ void HandleAutoSave() {
...
+    // If there is a cutscene upon load (i.e. entrance cutscene), just skip it
+    if (!skipEntranceCutsceneHookId) {
+        skipEntranceCutsceneHookId = REGISTER_VB_SHOULD(VB_START_CUTSCENE, {
+            // Only skip normal cutscenes
+            if (gSaveContext.gameMode == GAMEMODE_NORMAL && gPlayState->sceneId != SCENE_SPOT00) {
+                *should = false;
+            }
+        });
+    }
+}
```
mckinlee, line 207: Not the biggest fan of having two hooks depend on each other. If for some reason OnSaveLoad fires multiple times, unlikely, it would get reregistered. `RegisterAutosave()` has a better clean up pattern imo. Here is what I'd recommend:

```suggestion
/*
 * Upon loading a save, skip any cutscenes that would play if the save is from a cutscene entrance (e.g. owl warps, Link
 * bowing at Mikau's grave, etc.). An OnPassPlayerInputs hook is used to detect when gameplay actually starts (all
 * entrance cutscenes are done), at which point the cutscene skip hook is unregistered. This handles cases where multiple
 * cutscenes play in succession.
 */
static void UnregisterEntranceCutsceneSkip() {
    if (skipEntranceCutsceneHookId) {
        GameInteractor::Instance->UnregisterGameHookForID<GameInteractor::ShouldVanillaBehavior>(
            skipEntranceCutsceneHookId);
        skipEntranceCutsceneHookId = 0;
    }
    if (gameplayStartHookId) {
        GameInteractor::Instance->UnregisterGameHook<GameInteractor::OnPassPlayerInputs>(gameplayStartHookId);
        gameplayStartHookId = 0;
    }
}

void skipEntranceCutsceneOnLoad(s16 fileNum) {
    // Clean up any existing hooks first
    UnregisterEntranceCutsceneSkip();

    // Register hook to skip entrance cutscenes - may skip multiple if they chain
    skipEntranceCutsceneHookId = REGISTER_VB_SHOULD(VB_START_CUTSCENE, {
        // Only skip normal cutscenes
        if (gSaveContext.gameMode == GAMEMODE_NORMAL && gPlayState != nullptr &&
            gPlayState->sceneId != SCENE_SPOT00) {
            *should = false;
        }
    });

    // Register hook to detect when gameplay starts (all cutscenes done)
    // OnPassPlayerInputs only fires during normal gameplay, not during cutscenes
    gameplayStartHookId = GameInteractor::Instance->RegisterGameHook<GameInteractor::OnPassPlayerInputs>(
        [](Input* input) {
            // Gameplay has started - all entrance cutscenes are done
            // Now unregister both hooks so normal cutscenes can play
            UnregisterEntranceCutsceneSkip();
        });
}
```

`killSkipEntranceCutsceneHookId` renamed to `gameplayStartHookId` for clarity.

mckinlee (commented):
Looking forward to this getting in. Made note of a issue we discussed in Discord and left a suggestion for the hook. Otherwise, everything else seems sound.

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
Eblo: Definitely a goof.

Eblo:
Addressed feedback, but found a new problem of my own: if the player pause saves immediately after leaving a grotto, then upon loading they will go to entrance 0 of whatever scene they exited into. Something else must be governing the target coordinates when exiting a grotto.

Eblo:
Pushed an answer to the above problem. The enhancement file now tracks the previous entrance on transition and, excepting the case of exiting a grotto, updates the entrance that will be saved to the file.

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
```diff
@@ -166,11 +234,25 @@ void RegisterSavingEnhancements() {
...
+        }
+        lastEntrance = gSaveContext.save.entrance;
+    });
+
     COND_HOOK(OnSaveLoad, true, [](s16 fileNum) {
         if (gSaveContext.save.shipSaveInfo.fileCreatedAt == 0) {
             gSaveContext.save.shipSaveInfo.fileCreatedAt = GetUnixTimestamp();
         }
         gSaveContext.shipSaveContext.lastTimeLog = GetUnixTimestamp();
+        lastEntrance = entranceToSave = gSaveContext.save.shipSaveInfo.pauseSaveEntrance;
```
mckinlee, line 243: Is this line intended to be ran even with the feature disabled? If so, I'd make this conditional on `CVAR_REMEMBER_SAVE_LOCATION`

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
```diff
@@ -191,6 +273,10 @@ void RegisterSavingEnhancements() {
     });
 
     GameInteractor::Instance->RegisterGameHook<GameInteractor::BeforeMoonCrashSaveReset>([]() { DeleteOwlSave(); });
+
+    GameInteractor::Instance->RegisterGameHook<GameInteractor::OnSaveLoad>(loadRespawnData);
+
+    GameInteractor::Instance->RegisterGameHook<GameInteractor::OnSaveLoad>(skipEntranceCutsceneOnLoad);
```
mckinlee: Same here. Should this be conditional on `CVAR_REMEMBER_SAVE_LOCATION`?

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
```diff
@@ -7,38 +7,53 @@ extern "C" {
...
 
+#define CVAR_REMEMBER_SAVE_LOCATION_NAME "gEnhancements.Saving.RememberSaveLocation"
+#define CVAR_REMEMBER_SAVE_LOCATION CVarGetInteger(CVAR_REMEMBER_SAVE_LOCATION_NAME, 0)
+
 static uint32_t autosaveInterval = 0;
 static uint32_t iconTimer = 0;
 static uint64_t currentTimestamp = 0;
 static uint64_t lastSaveTimestamp = GetUnixTimestamp();
+static uint32_t lastEntrance = -1;
+static uint32_t entranceToSave = -1;
```
mckinlee: Why `uint32_t`? `gSaveContext.save.entrance` is `s32`. `pauseSaveEntrance` is `s32`. `SavingEnhancements_GetSaveEntrance()` returns `int`.

mckinlee (commented):
Looking good. Just a few questions.

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
Eblo: Since it only matters at the point of pause saving, which is conditional, there's no harm in just letting this always run.

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
Eblo: Yes, this can and should be changed.

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
Eblo: Oops, yes this should be signed.

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
Eblo: This helped uncover that I wasn't using ShipInit here, so settings weren't updating past launch anyway. Fixed that.

mckinlee (approved):

PR #892, opened 2024-12-12
Handle Great Bay Coast for pause warp with index warp

Archez:
When using Index-Warp, after loading from a save or a scene change, the pause cursor for the pause menu defaults to an index of 0, which correlates to the "Great Bay Coast". This means that it is free to be able to warp to Great Bay with Index-Warp active.

This PR makes the pause owl warp feature just give the Greay Bay Coast warp point for free as a convenience, and updates the wording of the enhancement tooltip to explain the Index-Warp behavior better.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2314363285.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2314364796.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2314373611.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenuBar.cpp
+++ b/mm/2s2h/BenGui/BenMenuBar.cpp
@@ -770,8 +770,12 @@ void DrawEnhancementsMenu() {
             UIWidgets::CVarCheckbox("Skip Scarecrow Song", "gEnhancements.Playback.SkipScarecrowSong",
                                     { .tooltip = "Pierre appears when the Ocarina is pulled out." });
             UIWidgets::CVarCheckbox("Pause Owl Warp", "gEnhancements.Songs.PauseOwlWarp",
-                                    { .tooltip = "Allows the player to use the pause menu map to owl warp instead of "
-                                                 "having to play the Song of Soaring." });
+                                    { .tooltip =
+                                          "Allows warping to registered Owl Statues from the pause menu map screen. "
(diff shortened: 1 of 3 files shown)


mckinlee:
Love the tooltip changes. In vanilla, you don't get the great bay map point until you visit that location correct? I could be remembering wrong.  If so, my thought is to fix the enhancement to align with that and add a setting to unlock great bay.

Archez:
>  If so, my thought is to fix the enhancement to align with that and add a setting to unlock great bay.

I'm not sure I understand your question. Yes you don't get the GB map point until visiting GB. The way Index-Warp works is that the owl warp cursor position is the same value as the pause menu map cursor position. If you never cursor onto a map point, then the default value for the cursor is 0. This means in vanilla, with index warp active, you can warp to GB without having its map point so long as you don't move the cursor on the pause screen.

Because the pause menu owl warp enhancement requires that you move the cursor on the pause screen, and the enhancement also impacts the regular warp screen, it is basically not possible to "preserve" the default 0 value to get to GB early, so that is why I am giving it for "free" ("free" here is synonymous with how the enhancement "simplifies" the process of index-warp by not having to pre-select the cursor points).

garrettjoecox (approved):

PR #894, opened 2024-12-18
Adds even more interpolation fixes

Archez:
Adds the following interpolation fixes/changes:
* Interpolate Biggoron ice wind breath
* Fix interpolation on Keaton grass leaves vanishing swirl
* Interpolate Goron ice melting, snowball and dust effects
* Skip interpolation on pause cursor when moving it
* Skip interpolation on item drops once picked up by the player
* Interpolate wider angle changes to specifically handle Deku Link spinning and Deku Link diving into a Deku flower
  * Normally we have a heuristic to skip interpolating angle changes greater than 91º which I'm not sure how that was decided, so instead of increasing this across the board, I decided to just increase for Deku Link when in the two action states mentioned above.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2341077418.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2341079802.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2341089929.zip)

### DIFF
--- a/mm/src/code/z_actor.c
+++ b/mm/src/code/z_actor.c
@@ -2813,7 +2813,8 @@ void Actor_Draw(PlayState* play, Actor* actor) {
                    (actor->flags & (ACTOR_FLAG_10000000 | ACTOR_FLAG_400000)) ? NULL : &actor->world.pos, play);
     Lights_Draw(light, play->state.gfxCtx);
 
-    // If the player is performing a Deku spin or entering a Deku flower, set it so that angles
+    // If the player is performing a Deku spin or entering a Deku flower, set it so that interpolation allows for >90
+    // angle changes to be interpolated smoothly
     if (actor->id == ACTOR_PLAYER &&
         (((Player*)actor)->actionFunc == Player_Action_93 || ((Player*)actor)->actionFunc == Player_Action_95)) {
         FrameInterpolation_InterpolateWiderAngles();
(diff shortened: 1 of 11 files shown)


mm/src/code/z_actor.c
```diff
@@ -2809,6 +2813,12 @@ void Actor_Draw(PlayState* play, Actor* actor) {
                    (actor->flags & (ACTOR_FLAG_10000000 | ACTOR_FLAG_400000)) ? NULL : &actor->world.pos, play);
     Lights_Draw(light, play->state.gfxCtx);
 
+    // If the player is performing a Deku spin or entering a Deku flower, set it so that angles
```
garrettjoecox: > set it so that angles

?

garrettjoecox (approved):
Not the biggest fan of the solution for the item00, I assume there's no other field we can use to determine that state? I know one of them is lifetime but I think that might be multi-use.

Archez:
> Not the biggest fan of the solution for the item00, I assume there's no other field we can use to determine that state? I know one of them is lifetime but I think that might be multi-use.

I tried checking for the actionFunc originally, but the problem with that is I need to know that the actionFunc has been executed. Since Draw is performed after Update, the actionFunc is already assigned but never called yet by the time Draw is executed.

I might be able to get away with not adding a field to item00 struct by getting the player and seeing if the item00 pos.x and pos.z are identical to the player's.

mm/src/code/z_actor.c
Archez: Updated

Archez:
> I might be able to get away with not adding a field to item00 struct by getting the player and seeing if the item00 pos.x and pos.z are identical to the player's.

Updated PR to drop the item00 struct changes in favor of this.

PR #900, opened 2024-12-22
[Enhancement]: Add File Slot 3

Archez:
Adds support for a 3rd file slot that can be used for saves.

The File Choose system largely already supported this (from JP 1.0), but just had early escapes for rendering and menu navigation.
The Sram system needed more work in terms of the actual management of the writing/reading processes and owl save management.

Opening as a draft to show what it looks like and pose some questions on implementation. Currently everything is behind a user toggle that is defaulted on.

~~Open Questions:~~
* Should this be a user facing toggle? **— Edit: Stick with a toggle for now for those that want to remove it**
* If it is user facing, should it default on? **— Edit: Consensus is to default on**
* Should we create/maintain separate structs, functions, and code paths for Sram related functions based on the toggle, or should we modify the originals, and only have the toggle visually hide the file slot on the File Choose menus? **— Edit: Modify existing structs, but first made enums to replace magic values to reduce surface area of code touched**

image

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2409809814.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2409811532.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2409825757.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Saving/FileSlot3.cpp
+++ b/mm/2s2h/Enhancements/Saving/FileSlot3.cpp
@@ -1,0 +1,21 @@
+#include <libultraship/libultraship.h>
+#include "2s2h/ShipInit.hpp"
+
+extern "C" {
+#include "src/overlays/gamestates/ovl_file_choose/z_file_select.h"
+
+extern FileSelectState* gFileSelectState;
+}
+
+#define CVAR_NAME "gEnhancements.Saving.FileSlot3"
+#define CVAR CVarGetInteger(CVAR_NAME, true)
+
+void RegisterFileSlot3() {
+    // Reset the file select state to reload the save metadata
+    if (gFileSelectState != NULL) {
+        STOP_GAMESTATE(&gFileSelectState->state);
+        SET_NEXT_GAMESTATE(&gFileSelectState->state, FileSelect_Init, sizeof(FileSelectState));
+    }
+}
+
+static RegisterShipInitFunc initFunc(RegisterFileSlot3, { CVAR_NAME });
--- a/mm/2s2h/BenGui/BenMenuBar.cpp
+++ b/mm/2s2h/BenGui/BenMenuBar.cpp
@@ -504,6 +504,9 @@ void DrawEnhancementsMenu() {
         if (UIWidgets::BeginMenu("Saving / Time Cycle")) {
 
             ImGui::SeparatorText("Saving");
+            UIWidgets::CVarCheckbox(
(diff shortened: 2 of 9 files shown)


Archez:
> or should we modify the originals, and only have the toggle visually hide the file slot

Example of this PR if it were to modify original structs/funcs instead: https://github.com/HarbourMasters/2ship2harkinian/compare/develop...Archez:2ship2harkinian:file-slot-3-alt

sitton76:
Since you listed this as a open question I am gonna throw my 2 cents.

Unlike other enhancements I cannot see a sufficiently good reason why anyone would want to disable this. Aside from maybe the rare person who might be upset to see it "because it was not on my version of the game."

I personally think having this on by default is fine, and even having it always on could be fine.

Caladius:
I'm all for default on. No reason it can't be.

Inb4 "I opened mm but it's OoT, how do I load mm in 2ship?"

PurpleHato:
Same as above, I don't see any problems to have this by default, unless the File3 texture isn't present in the PAL version or its supported language
However, we're far beyond the point of even considering that possibility anyway :D

Archez:
Rebased after splitting out some changes for magic values to be converted to enums by #933. Toggle now strictly just impacts the UI of the file select screen and defaults to on so people get file slot 3 automatically.

Ready for review.

mm/2s2h/Enhancements/Saving/FileSlot3.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+extern FileSelectState* gFileSelectState;
+}
+
+#define CVAR_NAME "gEnhancements.Saving.FileSlot3"
+#define CVAR CVarGetInteger(CVAR_NAME, true)
+
+void RegisterFileSlot3() {
+    // Reset the file select state to reload the save metadata
+    if (gFileSelectState != NULL) {
```
garrettjoecox, line 15: fancy

garrettjoecox (approved):
Very nice

mm/2s2h/Enhancements/Saving/FileSlot3.cpp
Archez: Went ahead and applied this reset state pattern to the save import flow as there were some things with the UI that would get funky if you imported a save while being deeper in the file select menu.

PR #901, opened 2024-12-22
[Enhancement] Cycle Arrows

mckinlee:
This enhancement adds a quick arrow cycling feature that lets players switch between Normal, Fire, Ice, and Light Arrows by pressing L while holding the bow. It includes a button-flash effect for visual feedback, and takes care of magic usage by refunding magic if you switch away from a magic arrow before firing. Additional checks ensure arrow cycling won’t interfere with minigames or other in-game states. Minimap toggle is disabled while cycling.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2391157944.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2391170672.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2391172990.zip)


Archez:
> * Magic Use: Magic is only consumed when releasing a magic arrow, not while drawing it.
> * Fallback: If the player cannot afford a magic arrow, the game reverts to Normal Arrows.

Instead of changing the behavior of magic consumption, could we instead just "give back" the consumed magic if the arrow was switched without firing? Seems like that would simplify some of the handling, like auto switching to normal arrows, since the game naturally makes magic arrows fire as normal without the proper magic.

mm/2s2h/Enhancements/Equipment/ArrowCycle.cpp
```diff
@@ -0,0 +1,286 @@
...
+
+    Player_InitItemAction(play, player, static_cast<PlayerItemAction>(nextArrow));
+    UpdateEquippedBow(play, nextArrow);
+    Audio_PlaySfx(NA_SE_PL_CHANGE_ARMS);
+}
+
+// Registration and Hooks
+void RegisterArrowCycle() {
+    COND_VB_SHOULD(VB_MINIMAP_TOGGLE, CVAR, {
+        if (CVAR && gPlayState != NULL && CanCycleArrows()) {
```
garrettjoecox: From what I'm reading, gPlayState should never be null in this VB handler, and it's conditionally registered based on the cvar, so you can change this to just
```suggestion
        if (CanCycleArrows()) {
```

mm/2s2h/Enhancements/Equipment/ArrowCycle.cpp
```diff
@@ -0,0 +1,286 @@
...
+            Player* player = GET_PLAYER(gPlayState);
+            if (IsHoldingBow(player)) {
+                *should = false;
+            }
+        }
+    });
+
+    COND_HOOK(OnGameStateUpdate, CVAR, []() {
```
garrettjoecox, line 246: If you change this to an `OnActorUpdate, ACTOR_PLAYER` you won't need the playState check, and this hook won't run when the player isn't updating

garrettjoecox (approved):
Two tiny comments otherwise LGTM. nice job keeping this isolated

PR #905, opened 2024-12-24
Fix invalid menu name in rumble testing

lightmanLP:
```cpp
std::dynamic_pointer_cast<BenInputEditorWindow>(Ship::Context::GetInstance()->GetWindow()->GetGui()->GetGuiWindow("Input Editor"));
```
`Input Editor` -> `LUS::InputEditorWindow`
`2S2H Input Editor` -> `BenInputEditorWindow`

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2370088757.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2370089811.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2370090208.zip)

### DIFF
--- a/mm/2s2h/BenPort.cpp
+++ b/mm/2s2h/BenPort.cpp
@@ -1638,14 +1638,14 @@ extern "C" void OTRControllerCallback(uint8_t rumble) {
     if (controllerConfigWindow == nullptr) {
         controllerConfigWindow = std::dynamic_pointer_cast<BenInputEditorWindow>(
             Ship::Context::GetInstance()->GetWindow()->GetGui()->GetGuiWindow("2S2H Input Editor"));
-        // TODO: Add SoH Controller Config window rumble testing to upstream LUS config window
-        //       note: the current implementation may not be desired in LUS, as "true" rumble support
-        //             using osMotor calls is planned: https://github.com/Kenix3/libultraship/issues/9
-        //
-        // } else if (controllerConfigWindow->TestingRumble()) {
-        //     return;
+        // note: the current implementation may not be desired in LUS, as "true" rumble support
+        //    using osMotor calls is planned: https://github.com/Kenix3/libultraship/issues/9
+    }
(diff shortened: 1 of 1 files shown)


Archez:
As a fix, I think this would be better rebased and pointed to the `develop-Satoko` branch.

Archez (approved):

mm/2s2h/BenPort.cpp
```diff
@@ -1637,7 +1637,7 @@ extern "C" void OTRControllerCallback(uint8_t rumble) {
     static std::shared_ptr<BenInputEditorWindow> controllerConfigWindow = nullptr;
     if (controllerConfigWindow == nullptr) {
         controllerConfigWindow = std::dynamic_pointer_cast<BenInputEditorWindow>(
-            Ship::Context::GetInstance()->GetWindow()->GetGui()->GetGuiWindow("Input Editor"));
+            Ship::Context::GetInstance()->GetWindow()->GetGui()->GetGuiWindow("2S2H Input Editor"));
```
Archez: @briaguya-ai I just noticed the code that uses this window fetch is commented out (for bailing out of rumble if the input editor was actively testing rumble). SoH does not have this commented out.

If it is not needed at all then I think this code can just be removed all together. Especially considering that rumble could be tested on a port that isn't port 1.

briaguya0: i was quite confused as to how this got to be this way. it looks like it was commented out back when we were using the upstream LUS input editor window https://github.com/HarbourMasters/2ship2harkinian/commit/782c51ccd8bcc857d270527c9745f78cf7bc86b2 which *doesn't* have rumble testing support.

since the 2ship input editor is based off the ship input editor, we should be able to uncomment that part

PR #906, opened 2024-12-26
Add patch for fixing depth render on the transition wipe effect

Archez:
The transition DL turned on G_ZBUFFER and G_ZS_PRIM and uses gsDPSetPrimDepth to set the depth test value as 0. gsDPSetPrimDepth is unimplemented in LUS, which meant that the DL was just using the normal depth buffer as a test, leading to the effect being obscured behind the scene geometry.

In lieu of a proper prim depth implementation, we can temporarily remove G_ZBUFFER since the prim depth was being set to 0 anyways.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2364027797.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2364030105.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2364039215.zip)

### DIFF
--- a/mm/2s2h/Enhancements/GfxPatcher/AuthenticGfxPatches.cpp
+++ b/mm/2s2h/Enhancements/GfxPatcher/AuthenticGfxPatches.cpp
@@ -7,6 +7,7 @@ extern "C" {
 #include "objects/object_fz/object_fz.h"
 #include "objects/object_ik/object_ik.h"
 #include "overlays/ovl_En_Syateki_Okuta/ovl_En_Syateki_Okuta.h"
+#include "overlays/ovl_fbdemo_wipe1/ovl_fbdemo_wipe1.h"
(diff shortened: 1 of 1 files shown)


garrettjoecox:
Just to clarify, so on Ship it seems like we either just pulled this DL into the codebase ourselves, or that was how decomp had it, whereas it's in an asset on 2ship?

Archez:
Correct, oot decomp had it in-lined because of ZAPD issues or something. I don't know if that problem still exists for them, but it was never changed from what I can see. Not sure why mm didn't have that problem.

garrettjoecox (approved):

PR #913, opened 2024-12-31
Allow Better SoDT to work on Night of the Final Day

zodiac-ill:
Ordinarily, Song of Double Time does not work on Night of the Final Day in vanilla as you cannot advance any farther. This makes sense in vanilla, but Better SoDT users may want to advance time within Night of the Final Day.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2384204067.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2384212952.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2384220151.zip)


garrettjoecox (commented):
Nice! Been meaning to get to this, would you mind changing the VB to be more descriptive then this LGTM! (something like `VB_ALLOW_SONG_DOUBLE_TIME_ON_FINAL_NIGHT`)

zodiac-ill:
> Nice! Been meaning to get to this, would you mind changing the VB to be more descriptive then this LGTM! (something like `VB_ALLOW_SONG_DOUBLE_TIME_ON_FINAL_NIGHT`)

Got it!

mm/src/code/z_message.c
```diff
@@ -6027,7 +6027,8 @@ void Message_Update(PlayState* play) {
                     }
                 } else if (sLastPlayedSong == OCARINA_SONG_DOUBLE_TIME) {
                     if (interfaceCtx->restrictions.songOfDoubleTime == 0) {
-                        if ((CURRENT_DAY != 3) || (gSaveContext.save.isNight == 0)) {
+                        bool isNightOfFinalDay = (CURRENT_DAY == 3) && (gSaveContext.save.isNight);
```
Archez: I would prefer if the syntax of this conditional remained the same as the original source code. Or even yet, just inline the original conditional with the GI Should call, as it is not a large condition (the VB enum helps to give it readability as is).

zodiac-ill: Should be better now

Archez (approved):

PR #917, opened 2025-01-02
[Enhancement] Add Gibdo Trade Sequence Options

zodiac-ill:
Add options to the Gibdo trade sequence beneath the well in Ikana.

- Vanilla: Gibdo trade sequence works normally
- MM3D: Gibdos will only require (and take) one of the item they request, as in MM3D. Additionally, the Gibdo that requests a blue potion will also accept a red potion.
- No Trade: Gibdos will not require or take any items. They will disappear without requiring a trade.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2514051993.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2514059355.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2514065372.zip)


garrettjoecox (changes requested):
Solid job! Sorry it's taken a minute to get you some feedback.

So for the most part you nailed it with using `GameInteractor_Should` in places to branch into custom behavior, with the main exception that you should introduce that behavior outside of the original source.

So instead of
```cpp
if (GameInteractor_Should(VB_..., true)) {
    // Original Stuff
} else {
    // Custom Stuff
}
```

In your file, where you set `*should = false;` is where the custom stuff should belong.

The other part is I'd like to not add new values to the `EnTalkGibudRequestedItemIndex` enum and `sRequestedItemTable`. Let me know if you run into trouble with this and I can try and figure something out.

zodiac-ill:
When you get the chance, please review revised commit

mm/2s2h/Enhancements/DifficultyOptions/GibdoTradeSequenceOptions.cpp
```diff
@@ -0,0 +1,72 @@
...
+    COND_VB_SHOULD(VB_GIBDO_TRADE_SEQUENCE_SUFFICIENT_QUANTITY_PRESENTED, CVAR != GIBDO_TRADE_SEQUENCE_VANILLA, {
+        ItemId requestedItemId = va_arg(args, ItemId);
+        if (AMMO(requestedItemId) >= 1) {
+            *should = true;
+        }
+    });
+
+    COND_VB_SHOULD(VB_GIBDO_TRADE_SEQUENCE_ACCEPT_RED_POTION, CVAR == GIBDO_TRADE_SEQUENCE_MM3D, {
+        PlayerItemAction requestedItemAction = va_arg(args, PlayerItemAction);
+        PlayerItemAction presentedItemAction = va_arg(args, PlayerItemAction);
```
garrettjoecox: I believe this is required when passing things like enums through va_args, someone can correct me if I'm wrong
```suggestion
        PlayerItemAction requestedItemAction = (PlayerItemAction)va_arg(args, int);
        PlayerItemAction presentedItemAction = (PlayerItemAction)va_arg(args, int);
```

mm/2s2h/Enhancements/DifficultyOptions/GibdoTradeSequenceOptions.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+        EnTalkGibudRequestedItem** requestedItem = va_arg(args, EnTalkGibudRequestedItem**);
```
garrettjoecox, line 38: Does this need to be a nested pointer? I think you can get away with just the top level pointer and changing it

mm/2s2h/Enhancements/DifficultyOptions/GibdoTradeSequenceOptions.cpp
```diff
@@ -0,0 +1,72 @@
...
+        if (CVAR == GIBDO_TRADE_SEQUENCE_MM3D) {
+            Inventory_ChangeAmmo(requestedItem->item, -1);
+        }
+    });
+
+    COND_VB_SHOULD(VB_GIBDO_TRADE_SEQUENCE_DO_TRADE, CVAR == GIBDO_TRADE_SEQUENCE_NO_TRADE, {
+        *should = false;
+
+        EnTalkGibud* gibudCtx = va_arg(args, EnTalkGibud*);
+        PlayState* play = va_arg(args, PlayState*);
```
garrettjoecox: You don't have to change it, but PlayState is globally available with `gPlayState`

garrettjoecox (commented):
Some small things, and I'd like someone to thoroughly QA this, then we can get it in

mm/2s2h/Enhancements/DifficultyOptions/GibdoTradeSequenceOptions.cpp
Eblo: You are correct, see #840 for example

mm/src/overlays/actors/ovl_En_Talk_Gibud/z_en_talk_gibud.c
```diff
@@ -713,10 +713,13 @@ void EnTalkGibud_GetNextTextBoxId(EnTalkGibud* this, PlayState* play) {
...
     EnTalkGibudRequestedItem* requestedItem = &sRequestedItemTable[this->requestedItemIndex];
-
+    GameInteractor_Should(VB_GIBDO_TRADE_SEQUENCE_ACCEPT_RED_POTION, false, requestedItem->itemAction,
+                          presentedItemAction, &requestedItem);
     if (requestedItem->itemAction == presentedItemAction) {
         if (!requestedItem->isBottledItem) {
-            if (AMMO(requestedItem->item) >= requestedItem->amount) {
+            if ((AMMO(requestedItem->item) >= requestedItem->amount) ||
+                GameInteractor_Should(VB_GIBDO_TRADE_SEQUENCE_SUFFICIENT_QUANTITY_PRESENTED, false,
+                                      requestedItem->item)) {
```
Eblo:
```suggestion
            if (GameInteractor_Should(VB_GIBDO_TRADE_SEQUENCE_SUFFICIENT_QUANTITY_PRESENTED, 
                AMMO(requestedItem->item) >= requestedItem->amount,
                                      requestedItem->item) {
```

mm/2s2h/Enhancements/DifficultyOptions/GibdoTradeSequenceOptions.cpp
```diff
@@ -0,0 +1,72 @@
...
+    /* 0x8 */ s32 amount;
+    /* 0xC */ s16 isBottledItem;
+} EnTalkGibudRequestedItem; // size = 0x10
+}
+
+static EnTalkGibudRequestedItem redPotionRequestedItem = { PLAYER_IA_BOTTLE_POTION_RED, ITEM_POTION_RED, 1, true };
+
+void RegisterGibdoTradeSequenceOptions() {
+    COND_VB_SHOULD(VB_GIBDO_TRADE_SEQUENCE_SUFFICIENT_QUANTITY_PRESENTED, CVAR != GIBDO_TRADE_SEQUENCE_VANILLA, {
+        ItemId requestedItemId = va_arg(args, ItemId);
```
Eblo: Similar to the other suggestions for enums passed through va_args:
```suggestion
        ItemId requestedItemId = (ItemId)va_arg(args, int);
```

Eblo (changes requested):
One additional thought, which I don't consider a blocker for merging, is that the Gibdos' vanilla dialog does not reflect this setting. e.g. "at least five", "something blue which bestows health". Since the tooltip details the exact changes, I think that's acceptable.

mm/2s2h/Enhancements/DifficultyOptions/GibdoTradeSequenceOptions.cpp
zodiac-ill: The top-level pointer `requestedItem` points to the `sRequestedItemTable` entry that contains the blue potion. If I passed `requestedItem` by copy to the hook, that would leave with the ability to alter the vanilla table, which I definitely want to avoid. If I had altered the table, and the player switched back to Vanilla behavior, the gibdo would _only_ take a red potion, and not a blue one. So I definitely don't want to change what's already there, I found that the cleanest solution is to just pass a pointer to that pointer and change what it is pointing to, which would be our own entry.

zodiac-ill:
> One additional thought, which I don't consider a blocker for merging, is that the Gibdos' vanilla dialog does not reflect this setting. e.g. "at least five", "something blue which bestows health". Since the tooltip details the exact changes, I think that's acceptable.

I had also thought that the vanilla dialogue could possibly create confusion. However I wanted to avoid implementing a message that matches/resembles the MM3D dialogue, as that could be seen as including an asset from that game. Ultimately as you said I think the tooltip serves well enough.

Patrick12115:
Ran through it a few times in game and everything seemed to work as intended

garrettjoecox (approved):
Final change, let's move this outside of cheats, maybe gameplay or difficulty options

Patrick12115:
I believe this is under difficulty options in the actual menu.

garrettjoecox:
> I believe this is under difficulty options in the actual menu.

ah, so it is, I guess all of the difficulty options have the incorrect cvar path. We can adjust that later then

PR #929, opened 2025-01-08
Allow reset when in any game state

Archez:
There is no reason why we should only be allowed to reset when in a `PlayState`. One of my longest pet peeves across the projects.

Simply updates the reset command to use the master `GameState` instead 🙂 

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2403558051.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2403567114.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2403571587.zip)

### DIFF
--- a/mm/2s2h/DeveloperTools/DebugConsole.cpp
+++ b/mm/2s2h/DeveloperTools/DebugConsole.cpp
@@ -119,14 +119,13 @@ static bool SetPosHandler(std::shared_ptr<Ship::Console> Console, const std::vec
 }
 
 static bool ResetHandler(std::shared_ptr<Ship::Console> Console, std::vector<std::string> args, std::string* output) {
-    if (gPlayState == nullptr) {
-        ERROR_MESSAGE("PlayState == nullptr");
+    if (gGameState == nullptr) {
+        ERROR_MESSAGE("gGameState == nullptr");
         return 1;
     }
 
-    gPlayState->gameplayFrames = 0;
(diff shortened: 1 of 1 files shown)


mm/2s2h/DeveloperTools/DebugConsole.cpp
```diff
@@ -119,14 +119,13 @@ static bool SetPosHandler(std::shared_ptr<Ship::Console> Console, const std::vec
...
 
 static bool ResetHandler(std::shared_ptr<Ship::Console> Console, std::vector<std::string> args, std::string* output) {
-    if (gPlayState == nullptr) {
-        ERROR_MESSAGE("PlayState == nullptr");
+    if (gGameState == nullptr) {
+        ERROR_MESSAGE("gGameState == nullptr");
         return 1;
     }
 
-    gPlayState->gameplayFrames = 0;
```
Archez, line 127: This was copied from SoH, but isn't necessary in 2ship as the state variable is memset to 0 before init automatically from MM source.

garrettjoecox (approved):

inspectredc (approved):
Finally I can reset from the scene select screen 🙏🏼

PR #932, opened 2025-01-08
Add EasyFrameAdvance cheat

garrettjoecox:
Introduced a `OnGameStateMainStart`, because I'm augmenting input it needs to happen prior to update

I know a lot of people think this shouldn't be considered a cheat, but I still kinda do. Open to better placement suggestions

Also if it's not obvious, this behaves the same as Ship's, holding pause when unpausing will allow one frame to advance before pausing again

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2404461435.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2404466222.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2404471811.zip)


Archez (approved):
I was hoping that we would figure out the changes necessary to simulate unpause controller input lag but I think that would require more changes than I originally assumed.

In lieu of that, this works for now and is simple enough 👍

PR #936, opened 2025-01-10
Prevent Faster Song Playback During Cutscenes

zodiac-ill:
After discussing with Archez, it appears that this is a bug and not intended functionality of the enhancement.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2415697120.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2415699438.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2415710653.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Songs/FasterSongPlayback.cpp
+++ b/mm/2s2h/Enhancements/Songs/FasterSongPlayback.cpp
@@ -13,7 +13,8 @@ extern u8 sPlaybackState;
 
 void RegisterFasterSongPlayback() {
     COND_ID_HOOK(OnActorUpdate, ACTOR_PLAYER, CVAR, [](Actor* actor) {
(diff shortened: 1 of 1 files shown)


Archez:
Nice, simple enough. Assuming you tested all scenarios and there isn't any false positives/false negatives?

zodiac-ill:
> Nice, simple enough. Assuming you tested all scenarios and there isn't any false positives/false negatives?

Not quite everything yet, as it seemed like it would work consistently after testing it for a number of songs, but I will do more tests and update you when that is done.

zodiac-ill:
@Archez Alright, after checking everything I can think to check, this should be good to merge

Archez (approved):
Thanks for verifying all that! :shipit:

PR #938, opened 2025-01-11
Port difficulty options from SoH

garrettjoecox:
self explanatory.

For Permanent Heart Loss I would like to do what we do on SoH and store the amount of PoH/HC found and restore their health if they disable this, but not dealing with it rn

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2416522293.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2416522668.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2416525899.zip)


Archez (approved):
Just my two cents, but I think the file should be deleted almost immediately upon death, at least to prevent sneaky ctrl+r resets from bypassing the death timer.

Also an alternative to death -> fade to reset, we could also do death -> fade to moon crash -> then reset, for extra dramatic effect 🙂 

Just suggestions and not a deal breaker on merge, things can be tweaked later.

garrettjoecox:
> Just my two cents, but I think the file should be deleted almost immediately upon death, at least to prevent sneaky ctrl+r resets from bypassing the death timer.

Ooh good call, I didn't want an immediate reset because the player needs a moment to sulk, but an immediate file delete should be fine 👍

garrettjoecox:
@Archez did a quick pass on that if you want to sanity check it

PR #948, opened 2025-01-15
Modern menu v2

Malkierian:
This is the Dev UX rework PR. Basic overview of the changes:

- `SearchableMenuItems.h` is gone, with relevant things moved to either `Menu.cpp`, `Menu.h`, or `MenuTypes.h`, with a focus on keeping the headers includable in multiple places.
- `Menu` now holds all the base operations the menu will need to function. `BenMenu` is the 2ship extension of it that holds all the port-specific definitions. This model is mostly complete, but will need some fine-tuning for a few things that `Menu` needs access to that still need to be defined in the port derivation.
- Search results code has been unified into a single function that is called from wherever the search results are being displayed. The sidebar search button and text field have also been changed to match the menu theme.
- Menu theme options have been significantly expanded.
- Top bar search now always displays the "clear search" button.
- `widgetInfo` renamed to `WidgetInfo` to match struct naming conventions.
- Some `UIWidgets` changes to enable menu changes and enhancements.
- Menubar has been wiped (cleanup pending, it's all just commented out for now)
- The big thing is swapping widget definitions over to the builder method:
  - `WidgetInfo`and all widget Options structs have single-member functions that return references to the object being manipulated, so you can do function chains.
  - `WidgetInfo` now stores a `shared_ptr` of a base `WidgetOptions` type, which holds most members common to all widgets, but which can be cast to the derived Options structs to get widget-specific information.
  - Tooltip text has been moved to `WidgetOptions`, and almost all widget-specific members have been moved from `WidgetInfo` to the widget-specific structs (min/max/default, etc).
  - The `WidgetInfo::Options` function accepts an object of a variant type which includes all derived options structs so that the `shared_ptr` can be created automatically. This results in copying struct data, but since it only happens once, I deemed the simplification of code worth the overhead.
  - Menu and sidebar sections are now stored and referenced by name. This also allows the sidebar to be removed by name rather than arbitrary numerical index (though is still inserted based on preferred numeric index)
  - `AddWidget` added, which accepts a `Path` struct (which consists of header, sidebar, and column indicators), which also automatically adds the header and sidebar sections to the appropriate structures if they don't exist when first adding a widget to them, as well as simplifies the process of adding the widgets themselves to other structures.
- This has resulted in a *drastic* decrease in lines of code, and much clearer implications of what to use when.

As a note, `color` exists on both `WidgetOptions` and all derived structs because there's no way to make a default option for a base member in derived declaration when trying to maintain aggregate functionality, and thus, when color is changed, the functions in the derived structs need to change both the derived member and the base member. This may change in the future as more widgets in external windows are swapped over when I figure out how to have window widgets searchable but still easily display separately from the main menu.

Also, Options/WidgetType matching is currently not checked, but assumed. `static_pointer_cast` was used rather than dynamic because the compiler was giving me some grief over it.

A few things I still have planned that I will handle after this is merged (for the sake of reducing further conflicts, as the builder structure should remain mostly unchanged from here out):
- External/popout window widgets searchable and centrally located as well as easily adapting to the menu theme if it's desired, but easily displayed separately and not mistaken for standard menu display (low priority, would be opt-in on a per-window basis and have an option for search exlusion).
- Customizable maximum menu size (as function of percentage of main window size)
- Font scaling (not sure yet on manual or automatic, but leaning toward manual)
- Text wrapping for sidebar and widget entries
- ColorPicker widget

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2478883616.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2478884749.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2478894161.zip)


mm/2s2h/BenGui/Menu.cpp
```diff
@@ -132,7 +169,282 @@ bool ModernMenuHeaderEntry(std::string label) {
...
+                    ImGui::PushStyleColor(ImGuiCol_Text, widget.options->color);
+                }
+                ImGui::SeparatorText(widget.name.c_str());
+                if (widget.options->color != UIWidgets::Colors::NoColor) {
+                    ImGui::PopStyleColor();
+                }
+            } break;
+            case WIDGET_TEXT: {
+                if (widget.options->color != UIWidgets::Colors::NoColor) {
+                    ImGui::PushStyleColor(ImGuiCol_Text, widget.options->color);
```
Archez:
```suggestion
                    ImGui::PushStyleColor(ImGuiCol_Text, UIWidgets::ColorValues.at(widget.options->color));
```

mm/2s2h/BenGui/Menu.cpp
```diff
@@ -132,7 +169,282 @@ bool ModernMenuHeaderEntry(std::string label) {
...
+                    Ship::Context::GetInstance()->GetConfig()->Save();
+                    UpdateWindowBackendObjects();
+                }
+            } break;
+            case WIDGET_SEPARATOR: {
+                ImGui::Separator();
+            } break;
+            case WIDGET_SEPARATOR_TEXT: {
+                if (widget.options->color != UIWidgets::Colors::NoColor) {
+                    ImGui::PushStyleColor(ImGuiCol_Text, widget.options->color);
```
Archez:
```suggestion
                    ImGui::PushStyleColor(ImGuiCol_Text, UIWidgets::ColorValues.at(widget.options->color));
```

mm/2s2h/BenGui/Menu.cpp
```diff
@@ -1,19 +1,21 @@
...
 #include "DeveloperTools/ActorViewer.h"
 #include "DeveloperTools/CollisionViewer.h"
 #include "DeveloperTools/EventLog.h"
+#include "2s2h/Enhancements/GfxPatcher/AuthenticGfxPatches.h"
 #include "HudEditor.h"
-
-#include "SearchableMenuItems.h"
+#include "Notification.h"
+#include <variant>
+#include <fmt/format.h>
```
Archez:
```suggestion
#include <spdlog/fmt/fmt.h>
```

mm/2s2h/BenGui/BenMenu.cpp
```diff
@@ -0,0 +1,1312 @@
...
+#include "DeveloperTools/SaveEditor.h"
+#include "DeveloperTools/ActorViewer.h"
+#include "DeveloperTools/CollisionViewer.h"
+#include "DeveloperTools/EventLog.h"
+#include "2s2h/Enhancements/GfxPatcher/AuthenticGfxPatches.h"
+#include "HudEditor.h"
+#include "Notification.h"
+#include <variant>
+#include "StringHelper.h"
+#include <fmt/format.h>
```
Archez:
```suggestion
#include <spdlog/fmt/fmt.h>
```

mm/2s2h/BenGui/Menu.cpp
```diff
@@ -22,66 +24,101 @@ extern PlayState* gPlayState;
...
 
-void BenMenu::InitElement() {
+uint32_t GetVectorIndexOf(std::vector<std::string>& vector, std::string value) {
+    return std::distance(vector.begin(), std::find(vector.begin(), vector.end(), value));
+}
+
+void Menu::InsertSidebarSearch() {
+    menuEntries[0].sidebars.emplace("Search", searchSidebarEntry);
+    uint32_t curIndex = 0;
+    if (CVarGetString(menuEntries[0].sidebarCvar, "") != "") {
```
Archez: Direct comparison to c literal strings are undefined behavior, need to use a string compare func instead.
```suggestion
    if (!Ship_IsCStringEmpty(CVarGetString(menuEntries[0].sidebarCvar, "")) {
```

mm/2s2h/BenGui/Menu.cpp
```diff
@@ -132,7 +169,282 @@ bool ModernMenuHeaderEntry(std::string label) {
...
+                if (widget.options->color != UIWidgets::Colors::NoColor) {
+                    ImGui::PopStyleColor();
+                }
+            } break;
+            case WIDGET_TEXT: {
+                if (widget.options->color != UIWidgets::Colors::NoColor) {
+                    ImGui::PushStyleColor(ImGuiCol_Text, widget.options->color);
+                }
+                ImGui::AlignTextToFramePadding();
+                ImGui::Text(widget.name.c_str());
```
Archez:
```suggestion
                ImGui::Text("%s", widget.name.c_str());
```

mm/2s2h/BenGui/BenMenu.cpp
```diff
@@ -0,0 +1,1312 @@
...
+
+    // General Settings
+    WidgetPath path = { "Settings", "General", 1 };
+    settingsSidebar.emplace("General", SidebarEntry{ .columnCount = 3 });
+    settingsOrder.push_back("General");
+    AddWidget(path, "Menu Theme", WIDGET_CVAR_COMBOBOX)
+        .CVar("gSettings.MenuTheme")
+        .Options(ComboboxOptions().Tooltip("Changes the Theme of the Menu Widgets.")
+            .ComboMap(menuThemeOptions)
+            .DefaultIndex(Colors::DarkBlue));
```
Archez: This is subjective, but in my opinion I think the default/color should be `LightBlue`. That is what is the closest match from the "old indigo" and I think is easier on the eyes.

Should also update as the default within UIWidgets directly, and the reference in the save editor.

garrettjoecox: Indigo should be reverted to what it was prior, and made default everywhere as it was prior IMO

Malkierian: Indigo wasn't indigo, though, not according to like a dozen different color websites I looked at for it.

mm/2s2h/BenGui/BenMenu.cpp
```diff
@@ -0,0 +1,1312 @@
...
+//    return std::make_shared<FloatSliderOptions>(options);
+//}
+//
+//std::shared_ptr<IntSliderOptions> MakeIntSliderOptions(IntSliderOptions options) {
+//    return std::make_shared<IntSliderOptions>(options);
+//}
+//
+//std::shared_ptr<WidgetOptions> MakeWidgetOptions(WidgetOptions options) {
+//    return std::make_shared<WidgetOptions>(options);
+//}
```
garrettjoecox: I assume this can all go?
```suggestion
```

Malkierian: Yep.

mm/2s2h/BenGui/Menu.cpp
Malkierian: Why *have* you recommended this every time I've done it this way? It's not technically wrong, it's just that `fmt` has nothing to be added to it.

mm/2s2h/BenGui/Menu.cpp
Malkierian: This is because fmt/format doesn't exist on other platforms, right? I remember that kind of thing from before, I just can never remember what to actually include.

mm/2s2h/BenGui/Menu.cpp
garrettjoecox: IIRC doing it this way emits a warning or something

mm/2s2h/BenGui/Menu.cpp
garrettjoecox: Yes, I wasn't able to build as is

mm/2s2h/BenGui/BenMenu.cpp
garrettjoecox: It seems like a broad spectrum, but my issue is, I think it looks better than any of the other color options as a default lol. Closest maybe being dark green, light blue is too bright for contrast

Malkierian: Even just saying the color order, indigo is supposed to be between blue and violet, but what it was before was more like between light blue and blue. I wouldn't be opposed to readding it, though we'd have to have a different name for it, or replacing light blue with it.

Malkierian: We could also discuss all the colors generally, if we want to soften some of the brightness edges on all of them and gray them a bit (I think that's what the old indigo was, basically, to light blue currently).

mm/2s2h/BenGui/Menu.cpp
Archez: Yes, without the hardcoded format string, the passed in string could have format chars itself that aren't expected (think of injection concerns). So the warning says to specify our own format string to clarify behavior.

mm/2s2h/BenGui/BenMenu.cpp
garrettjoecox: Colors Feedback
- Dark Gray - just a smidge too dark, should be able to see the outline of buttons/checkboxes
- Light Gray - Not enough contrast, maybe just drop this? gray is plenty light enough
- Dark Brown, Brown, Orange - I think maybe drop Brown, change Dark Brown name to brown, and darken orange a bit, not enough contrast
- Violet - Not enough contrast, I think we can drop it personally in favor of purple
- Dark Blue - Should be shifted more towards gray or removed, hard on the eyes
- Light Blue - Not enough contrast
- Light Green - Not enough Contrast

garrettjoecox (approved):
Overall looks good to me, 

there are some things I don't love, like this little dance to define where you are at,
```cpp
// Difficulty Options
enhancementsSidebar.emplace("Difficulty Options", SidebarEntry{ .columnCount = 3 });
enhancementsOrder.push_back("Difficulty Options");
path = { "Enhancements", "Difficulty Options", 1 };
```

But it's much improved over the prior dev experience. Have you verified if you can successfully define a menu section in an external file and get it added to the menu?

Malkierian:
To be fair, it's not all defining where you're at, though I just made a function to encapsulate the vector and map insertions. I just haven't been able to figure out a way to do that automatically based on a path passed in to AddWidget because of the way column count is explicitly required (because that determines how many columns it divvies out space for, even if they're empty ones). The WidgetPath variable was just meant to be an easy way of passing the information without a bunch of redundant strings.

And no, I haven't verified that. Now that I added the function I just mentioned, it should be possible to add to existing sidebars, just can't define your own header/sidebar combo yet.

mm/2s2h/BenGui/BenMenu.cpp
Malkierian: Alright, I brightened Red, Green and Dark Gray just a bit, got rid of Light Green (because I was darkening it and it just became almost green), de-saturated Orange, Light Blue, Blue, and Violet a bit, tweaked Indigo to be a better in-between for blue and violet (I do still want to keep Violet and Purple), removed Light Gray from the theme selector, and darkened Brown and got rid of Dark Brown. Also changed default to Light Blue, as I think it's far better now, and closer to what Indigo was before.

Malkierian:
I also added text wrapping to Checkboxes (only those, because they're the only ones I could see, for now, that have overflow issues in the columns) to test it out. Not satisfied with the padding that happens, but not sure how to fix/unify it.

mm/2s2h/BenGui/Menu.cpp
```diff
@@ -426,7 +426,7 @@ void Menu::MenuDrawItem(WidgetInfo& widget) {
                 std::string menuSearchText(menuSearch.InputBuf);
 
                 if (menuSearchText == "") {
-                    ImGui::Text("Start typing to see results.");
+                    ImGui::Text("%s", "Start typing to see results.");
```
Archez: This `"%s"` is redundant, as the string itself is just hardcoded in place. The previous warning talk only concerns dynamic strings provided via variables.

mm/2s2h/BenGui/BenMenu.cpp
Malkierian: By the way, what do you think of the color tweaks?

mm/2s2h/BenGui/BenMenu.cpp
```diff
@@ -1256,12 +1269,22 @@ void BenMenu::InitElement() {
...
+            "Vertical Resolution Toggle Enabled" } },
+        { DISABLE_FOR_LOW_RES_MODE_ON,
+          { [](disabledInfo& info) -> bool { return CVarGetInteger(CVAR_LOW_RES_MODE, 0); },
+            "N64 Mode Enabled" } },
     };
 }
 
 void BenMenu::UpdateElement() {
     Ship::Menu::UpdateElement();
+    UpdateResolutionVars();
```
Archez: Now that `WIDGET_CUSTOM` is a thing with the custom func, I feel like this could just be done as a custom widget that renders nothing and just calls this func. That way we aren't establishing a pattern of sticking "update" funcs into here.

mm/2s2h/BenGui/ResolutionEditor.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
```
Archez: Replace `#pragma once` with `#ifndef _H #define _H #endif // _H` pattern.

mm/2s2h/BenGui/ResolutionEditor.h
```diff
@@ -0,0 +1,8 @@
...
+#include <libultraship/libultraship.h>
+
+namespace BenGui {
+bool IsDroppingFrames();
+void RegisterResolutionWidgets();
+void UpdateResolutionVars();
+} // namespace AdvancedResolutionSettings
```
Archez: This namespace comment is not accurate

mm/2s2h/BenGui/ResolutionEditor.cpp
```diff
@@ -0,0 +1,528 @@
...
+}
+
+bool IsDroppingFrames() {
+    // a rather imprecise way of checking for frame drops.
+    // but it's mostly there to inform the player of large drops.
+    const short targetFPS = CVarGetInteger("gInterpolationFPS", 20);
+    const float threshold = targetFPS / 20.0f + 4.1f;
+    return ImGui::GetIO().Framerate < targetFPS - threshold;
+}
+} // namespace AdvancedResolutionSettings
```
Archez: This namespace comment is not accurate

mm/2s2h/BenGui/ResolutionEditor.cpp
```diff
@@ -0,0 +1,528 @@
...
+static float aspectRatioY;
+static int32_t verticalPixelCount;
+// Additional settings
+static bool showHorizontalResField;
+static int32_t horizontalPixelCount;
+// Disabling flags
+static bool disabled_everything;
+static bool disabled_pixelCount;
+
+using namespace UIWidgets;
```
Archez, line 86: I think this `using` can be dropped, didn't see anything taking advantage of it.

mm/2s2h/BenGui/ResolutionEditor.cpp
Archez, line 1: Assuming this window is complete, there is a lot of commented code that can/should be removed.

mm/2s2h/BenGui/BenMenu.cpp
Malkierian: I want to setup some ShipInit stuff eventually that would handle this registration automatically, so I don't want to change this pattern to the idea of having a dummy update widget. I just don't know how to setup ShipInit type flows.

mm/2s2h/BenGui/ResolutionEditor.h
Malkierian: Is that just because it's not officially part of the C standard? I had to search the difference to know that there was one.

mm/2s2h/BenGui/ResolutionEditor.cpp
Malkierian: Yeah, I only pushed it so that Proxy could have access to the AddMenuEntry function and see how I was doing things with this for his work on properly implementing rando settings in the new menu.

mm/2s2h/BenGui/BenMenu.cpp
Archez: Looking at how this was originally used, this func was meant to be called before drawing the widgets. So in this particular case a dummy widget at the beginning makes more sense to me.

ShipInit could work too, you'd just need to specify the whole list of all CVars that could cause the need to calling the func again, but also feels overkill.

mm/2s2h/BenGui/ResolutionEditor.h
Archez: Correct. `#ifndef` is the standard pattern, and what is used across this code base as well (yes there are some stragglers remaining that need to be cleaned up).

mm/2s2h/BenGui/BenMenu.cpp
Malkierian: I meant to say ShipInit-type stuff from the start, as I envision just a static registration of update functions as necessary (so you don't have to insert the update function, or the registration function, anywhere specific) that get stored in a vector and called at the start of the draw widgets loop (though I think keeping it in the menu's update function would still serve the same purpose).

Archez: Ah I see what you mean. Well if you aren't willing to temporarily place it in a dummy widget, then it should at least be moved to the Draw func instead of `BenMenu::UpdateElement` since there is no reason to execute this func with the menu closed, or behind a visibility check.

(Also I still vote for dummy widget cause it also only _needs_ to be called if the advanced res settings are being drawn specifically).

Malkierian: Ah, I see what you mean about Draw vs Update for its placement. What if I setup the registration with a header/section combo that only ran the update functions registered when that section is open? I worry about having it in dummy widgets just because that puts the onus on the developer to make sure the widget is placed first before any of the other stuff that need its updating functionality, and I'd rather automate the order if I could.

mm/2s2h/BenGui/ResolutionEditor.cpp
Malkierian: For now, there is only one thing that uses it, but it's in the Apple block (UIWidgets::Colors). I know that there will be others, but I think it's not going to be nearly as much as I originally thought, so when I'm finished with it in the next PR, and the amount of usage doesn't justify it, I'll remove it and convert it back to full namespace pathing.

mm/2s2h/BenGui/UIWidgets.hpp
```diff
@@ -68,57 +105,335 @@ namespace UIWidgets {
...
+            return *this;
+        }
+    };
+
+    struct CheckboxOptions : WidgetOptions {
         bool defaultValue = false; // Only applicable to CVarCheckbox
         ComponentAlignment alignment = ComponentAlignment::Left;
         LabelPosition labelPosition = LabelPosition::Near;
-    };
+        Colors color = WidgetOptions::color = Colors::DarkBlue;
```
Archez: Lets make sure to update all the default color args to be `LightBlue` in UIWidgets.hpp and UIWdigets.cpp

mm/2s2h/BenGui/UIWidgets.hpp
```diff
@@ -68,57 +105,335 @@ namespace UIWidgets {
...
+            return *this;
+        }
+        FloatSliderOptions& Max(int32_t max_) {
+            max = max_;
+            return *this;
+        }
+        FloatSliderOptions& DefaultValue(int32_t defaultValue_) {
+            defaultValue = defaultValue_;
+            return *this;
+        }
```
Archez: Just noticed all these `FloatSliderOptions` are int32_t, and so things like `.Step(0.1f)` are getting truncated to `0` leaving the plus/minus buttons to do nothing, and default values de-synced with the actual usages.

Malkierian: Oops XD

Malkierian: Done.

PurpleHato:
After testing it on a fresh setup, the default theme appears to be green, but the combobox is set to the Dark Blue theme.
I suppose it's a tiny oopsies ^^
image

Edit: I've also noticed that if we have the Search Autofocus enabled, we cannot use comboboxes as they basically auto-cancel themselves

https://github.com/user-attachments/assets/b1d9d391-ebe9-4a43-bf7c-40cbe3dd74e7

Malkierian:
How odd. I'll look into it, thanks.

Malkierian:
Defaults handled.

Archez (approved):
Good improvements to see for the menu system!

PR #956, opened 2025-01-17
Tweak infinite deku hops to stop velocity loss while in the air

garrettjoecox:
~~I don't love the changes to the vanilla code on this one, but wasn't quite sure about the best way to handle it~~
Did another pass and am happy now

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2485252088.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2485252246.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2485255162.zip)


inspectredc:
other alternative would be maybe to multiply speedtarget by 2 in the hook?

garrettjoecox:
Did another pass, was able to re-use and rename a VB

mm/2s2h/Enhancements/Player/InfiniteDekuHopping.cpp
```diff
@@ -16,6 +16,16 @@ void RegisterInfiniteDekuHopping() {
...
         }
     });
+
+    COND_VB_SHOULD(VB_APPLY_AIR_CONTROL, CVAR, {
+        f32* dekuSpeedTargetMultiplier = va_arg(args, f32*);
+        Player* player = GET_PLAYER(gPlayState);
+
+        if (player->transformation == PLAYER_FORM_DEKU) {
+            // Deku Link's air control speedTarget gets halved, so we negate that by doubling it
+            *dekuSpeedTargetMultiplier *= 2.0f;
```
Archez, line 26: This multiplier is being applied to all mid air movement for deku link. We need to have a "currently deku hopping" check for this. Probably a state flag or checking to see what makes the HUD display the number on the A button.

Looks like maybe `this->stateFlags3 & PLAYER_STATE3_200000`

Archez (commented):
I think this still needs iteration. Although you've adjusted the velocity stuff, once the first "last" hop happens, all control over deku link is lost and he will just infinitely hop in the last direction with no way to override. For example jumping out into great bay, you'll get soft locked until you hit land or void.

mm/2s2h/Enhancements/Player/InfiniteDekuHopping.cpp
garrettjoecox: Made this suggestion

PR #958, opened 2025-01-20
[Minigame] Added Setting Beaver Race Rings Amount

Patrick12115:
Allows setting the amount of rings you need to complete the beaver races. The slider will affect both beavers and setting the slider to the max of 20 will set them back to their vanilla amounts of 20 for the first beaver and 25 for the second.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2486442251.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2486443177.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2486445567.zip)


garrettjoecox:
Needs updating to modern menu v2

PR #961, opened 2025-01-20
Fix Persistent Bunny Hood losing color with other masks

Archez:
The bunny hood DL does not set the required GeometryMode values, instead it works by inheriting the previous values from links head limb.

The persistent bunny hood draw behavior runs after the currently equipped mask is drawn, and in some cases those masks change the GeometryMode values which lead to the bunny hood textures just rendering the env fog color instead.

This adds a `gSPLoadGeometryMode` before the persistent bunny hood draw so reset the values back.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2458573914.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2458576776.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2458583657.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Masks/PersistentMasks.cpp
+++ b/mm/2s2h/Enhancements/Masks/PersistentMasks.cpp
@@ -65,11 +65,17 @@ void UpdatePersistentMasksState() {
             }
 
             OPEN_DISPS(gPlayState->state.gfxCtx);
+
+            // Set back geometry modes left over from player head DL, incase another mask changed the values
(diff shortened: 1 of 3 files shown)


garrettjoecox (approved):

Archez:
Updated to include a fix to prevent persistent bunny hood applying to kafei, since kafei runs the same player update code.

PR #963, opened 2025-01-22
Improve better song of double time experience

Archez:
This makes a few significant updates to the Better Song of Double Time enhancement:

* Selecting your time no longer actors the actors and scene on the fly, instead the final time will only be applied after selecting your time.
* Selecting your time will queue up the Song of Double Time cutscene (pressing A can skip it as usual)
* To ensure that actors aren't processed early on the new time, we force a respawn in-place to reload the scene with the new time applied. This covers use cases like trying to use BSoDT to "cheat" the Aliens minigame, but also other cases of actor quirks that we might not know of.
* The arrows/control stick indicators now animate just like they would in shops.
* Analog stick support for adjusting the time with more precision
* DPad support for adjusting the time in repeated amounts. Repeat behavior is similar to cursor movement on kaleido.
* Added holding R to speed up time adjustments along side the existing holding Z to slow down
* Added visual indicators for Z and R
* Added a HUD override-_override_ system so that we can temporarily force the clock into is vanilla position/scale when using BSoDT
* Added Before/After hooks for clock drawing to facilitate hijacking the time values so that all clock draw reflects the selected time without the game state changing.
* Added checks for 0th day and 4th to prevent using SoDT similar to vanilla (on vanilla SoDT softlocks), which reuses existing text IDs. This is a compromise, because although we could maybe "support" 0th/4th, there aren't good ways to visualize that, and we'd have to prevent going from 0th/4th to normal days.
* The A button is shown on HUD with DECIDE

https://github.com/user-attachments/assets/fe0f8075-a5dd-48fa-b82a-7111a86597ff

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2481588273.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2481594205.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2481604535.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -986,7 +986,8 @@ void BenMenu::AddEnhancements() {
         .CVar("gEnhancements.Songs.BetterSongOfDoubleTime")
         .Options(CheckboxOptions().Tooltip(
             "When playing the Song of Double Time, you can now choose the exact time you want to go "
-            "to, similar to the 3DS version."));
+            "to, similar to the 3DS version.\n\n"
+            "Holding Z allows decreasing the time adjustment factor, while holding R will increase the factor"));
     AddWidget(path, "Enable Sun's Song", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.Songs.EnableSunsSong")
         .Options(CheckboxOptions().Tooltip(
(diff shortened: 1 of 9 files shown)


SuperBro64:
> DPad support for adjusting the time in singular amounts

By that, do you mean one in-game minute at a time?

Archez:
> > DPad support for adjusting the time in singular amounts
> 
> By that, do you mean one in-game minute at a time?

DPad would be 30 minutes, same as maximum analog stick. Singular just means its not repeated if you held the DPad.
Holding Z makes that 5 minutes, and holding R makes that 1 hour.

I don't think having anything less than 5 minutes is really needed considering how fast a minute is in-game.

Archez:
Updated so that now the DPad movement can be repeated if held. It will first move once, then wait some time before moving once more, and then wait a shorter amount of time, on repeat. This is ripped from how kaleido does repeated cursor movement.

Also made it so the A button is shown on the HUD with DECIDE

garrettjoecox (approved):
As per our discussion, we should either get a SodT cutscene skip into this PR or a very quick follow up, as people who have gotten used to the cutscene skip on the current iteration won't love randomly starting to see it again

PR #965, opened 2025-01-24
Add crash protection for action swap

Archez:
This adds some crash prevention around the behavior of action swap. Specifically there are two instances of OOB/UB access that both regard playing SFX.

The first instance happens when drawing an arrow with action swap. This OOB access on console leads to a value of `1` which isn't a valid SFX entry, but seems to be gracefully handled. For some windows users however, a different value is read from memory and could cause a hard crash in 2ship. Since this console behavior is that nothing happens, I've opted to address this situation by playing `NA_SE_NONE`.

The second instance happens when releasing the bow without having any bow ammo. This OOB access on console leads to a value of `58104`, which is not a valid SFX entry either, however this causes a crash on console. In this case I've opted to create a new handler `Ship_HandleConsoleCrashAsReset` which does as it suggests, which is to soft reset 2ship to avoid hard crashing. A notification is emitted to let the user know that the crash was prevented.

This handler can be placed in other places of known crashes from glitches, like index warping. In a follow up PR I might explore a "safer glitches" toggle that can be set by players to skip the soft reset and just keep the game playing like normal.

I'm not sure yet if I like this handler being placed in `BenPort` but I also don't know where else it makes sense (maybe ShipUtils?)

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2489257243.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2489264547.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2489269304.zip)

### DIFF
--- a/mm/2s2h/BenPort.cpp
+++ b/mm/2s2h/BenPort.cpp
@@ -1766,7 +1766,12 @@ extern "C" int Controller_ShouldRumble(size_t slot) {
 }
 
 // Helper to redirect the user to the boot screen in place of known console crash scenarios, and emits a notification
-extern "C" void Ship_HandleConsoleCrashAsReset() {
+extern "C" bool Ship_HandleConsoleCrashAsReset() {
+    // If fix crashes is on, return false and let fallback handling process in source
+    if (CVarGetInteger("gEnhancements.Fixes.ConsoleCrashes", 1)) {
+        return false;
+    }
+
     std::reinterpret_pointer_cast<Ship::ConsoleWindow>(
         Ship::Context::GetInstance()->GetWindow()->GetGui()->GetGuiWindow("Console"))
         ->Dispatch("reset");
@@ -1776,4 +1781,6 @@ extern "C" void Ship_HandleConsoleCrashAsReset() {
         .message = "Crash prevented!",
         .remainingTime = 10.0f,
     });
+
+    return true;
 }
--- a/mm/2s2h/BenPort.cpp
+++ b/mm/2s2h/BenPort.cpp
@@ -55,6 +55,7 @@ CrowdControl* CrowdControl::Instance;
 #include "2s2h/DeveloperTools/DebugConsole.h"
 #include "2s2h/DeveloperTools/DeveloperTools.h"
 #include "2s2h/SaveManager/SaveManager.h"
+#include "2s2h/BenGui/Notification.h"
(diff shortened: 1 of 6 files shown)


garrettjoecox (commented):
I know we've had some back and forth discussion, sorry if I've been confusing lol, but here's where I'm at regarding crashes:

In the menu, under the fixes category we have one checkbox labeled `"Fix console crashes"`, that defaults to on. The tooltip should state `"Fixes crashes that would typically happen on Console. Disabling this option will simply soft reset 2Ship when encountering these crashes instead of actually crash the program."`

This option should replace the current usage of `gEnhancements.Fixes.HessCrash` (and it's item in the menu)

And this action swap fix in this PR should check for this cvar, if it's disabled soft reset, if it's enabled handle it gracefully as you did above, playing NA_SE_NONE

How do you feel about that?

Archez:
> I know we've had some back and forth discussion, sorry if I've been confusing lol, but here's where I'm at regarding crashes:
> 
> In the menu, under the fixes category we have one checkbox labeled `"Fix console crashes"`, that defaults to on. The tooltip should state `"Fixes crashes that would typically happen on Console. Disabling this option will simply soft reset 2Ship when encountering these crashes instead of actually crash the program."`
> 
> This option should replace the current usage of `gEnhancements.Fixes.HessCrash` (and it's item in the menu)
> 
> And this action swap fix in this PR should check for this cvar, if it's disabled soft reset, if it's enabled handle it gracefully as you did above, playing NA_SE_NONE
> 
> How do you feel about that?

This is exactly where I was gonna go with it, just was waiting on the menu PR.

mm/2s2h/BenPort.cpp
```diff
@@ -1763,3 +1764,23 @@ extern "C" int Controller_ShouldRumble(size_t slot) {
 
     return 0;
 }
+
+// Helper to redirect the user to the boot screen in place of known console crash scenarios, and emits a notification
+extern "C" bool Ship_HandleConsoleCrashAsReset() {
+    // If fix crashes is on, return false and let fallback handling process in source
```
garrettjoecox, line 1770: I like this 👍

garrettjoecox (approved):

PR #966, opened 2025-01-24
Remove caching from CVar prefix defines

Archez:
These should not be cached, it could lead to weird behavior for local dev.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2482732326.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2482739577.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2482744281.zip)


Malkierian:
Do they still work as-is with the rest of it, though? As I recalled, I had trouble with it as anything but cached. The way to reset them in local dev is just with a --fresh in the setup command.

Malkierian:
The real solution to this, now that I know more about CMake, might just be setting it up with the Option flow instead, but that would also require a change to LUS to match.

Archez:
Yeah that is the issue here. They should not be cache vars on LUS side, but instead `option` with defaults, then ports use `set` like I have here in the PR.

From my testing (confirmed on mac and windows), my cache entry has the "old/bad" values, but with this PR the real values are still used. So this should be good to merged, but in general a PR to LUS is also preferred as well.

Per CMake docs:
> Note: The content of the cache variable will not be directly accessible if a normal variable of the same name already exists (see [rules of variable evaluation](https://cmake.org/cmake/help/latest/manual/cmake-language.7.html#cmake-language-variables)).

PR #968, opened 2025-01-26
Menu Modularize & MenuInit

Malkierian:
MenuEntries now uses unordered_map for storage, and vector for draw order.
AddMenuEntry is now used for main menu entries (settings, enhancements, etc).
MenuInit (based off of ShipInit) added to be able to register update and init functions (init can handle widgets, new sidebars in existing menu entries, or entirely new menu entries). Fixed top search modification to headerWidth to restore proper scrollbar functionality. Advanced resolution options now use MenuInit to handle update and init (options still hidden in this PR because of framebuffer issues).

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2995383774.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2995391022.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2995453839.zip)


mm/2s2h/BenGui/MenuTypes.h
```diff
@@ -200,7 +206,7 @@ struct WidgetInfo {
 struct WidgetPath {
     std::string sectionName;
     std::string sidebarName;
-    uint8_t column;
+    SectionColumns column;
```
Archez, line 207: I'm not sure if I agree with this becoming an enum. I don't really see what this buys us, at least with regards to how it is used currently in this PR (not sure if you have future plans for it). The numbers previously weren't "magical" in the normal sense of "magic numbers", they were just basic numbers. This enum also limits a menu section to at-most 3 columns, when before there was no limit (technically 255 for u8)/

If we wanted to restrict how many columns can be added to a particular menu section, I feel like that should be handled in a different way.

Archez (approved):
For the most part, this looks good. Just have one main comment below, but I wont let it block this PR since I know new menu work is happening on SoH that I haven't looked at yet.

mm/2s2h/BenGui/MenuTypes.h
Malkierian: I made it an enum because people were already using `1` in the functions for column 1, but that would actually put it in column 2. The enum isn't for restricting column counts, but for designating which column a widget gets added to.

Malkierian: Also, 3 columns was the physical limit as far as I could tell. Any more and the column widths made displaying widgets ineffectual.

Archez: Right, but before you had logic which would take the column value and subtract 1 to normalize it to a 0-based index. Why not just keep it that way and make sure anything else that wasn't already normalizing is updated to handle that? That kind of implementation detail should be internalized to the menu system itself anyways.

garrettjoecox: Seems like there's still some discussion here, but wasn't blocking, should we get these conflicts resolved and merge?

Malkierian: Having had some time to think about the enum situation, my feelings are that this makes it absolutely clear at the top level what you will get. Having the int parameter, if you weren't already familiar with the system, you'd have to go several layers deep in order to see what it was expecting and how it was applying.

Malkierian: And to put a bow on the column count thing, I don't know that you could just arbitrarily decide to put more than 3 in a section and have it just work. I didn't program it or test for that possibility. All that being said, I'm updating this now and should have it ready to go shortly. After this gets merged, I'll start working on porting all the extra changes and fixes from SoH.

libultraship
Archez, line 1: Looks like a rogue LUS reference change from merge conflict

Malkierian: argh. I swear I took care of that

Malkierian: Done.

PR #970, opened 2025-01-27
[Enhancement] Add Moon Crash Cutscene Skip to Story Cutscenes time savers

JoshSanch:
Getting a game over on the first cycle without saving after using Skip Intro Sequence creates a weird behavior that makes this cutscene skip seem strange - but it also occurs when the cutscene skip is not used. Seems to be an issue with the Skip Intro Sequence which is in use in all video examples here - I have added links here to view the behaviors for reference.

Link to view implementation behavior: https://www.youtube.com/watch?v=37N_wjt-yB4

Links to view quirky behavior when Moon Crash occurs on the first cycle without saving (Skip Misc Interactions enabled/disabled):
https://youtu.be/kvy4NSR1s9w (Moon Crash CS Enabled)
https://youtu.be/nB7_fK1Rmrc (Moon Crash CS Disabled)

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4283695047.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4283707057.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4283751436.zip)


mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipMoonCrash.cpp
```diff
@@ -0,0 +1,43 @@
...
+}
+
+#define CVAR_NAME "gEnhancements.Cutscenes.SkipMiscInteractions"
+#define CVAR CVarGetInteger(CVAR_NAME, 0)
+
+// Didn't end up using these few cutscene values, but they are captured here as documentation in case they
+// are relevant in the future
+#define TERMINA_FIELD_MOON_CRASH_CS_ENTRANCE 0x54C0
+#define VOID_FIRE_WALL_CS_ENTRANCE 0x1C00
+#define VOID_FIRE_WALL_CS_INDEX 0xFFF8
```
garrettjoecox: I think leaving them here is a smell, and might just lead to confusion as people glaze over the comment
```suggestion
#define TERMINA_FIELD_MOON_CRASH_CS_ENTRANCE 0x54C0
```

mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipMoonCrash.cpp
```diff
@@ -0,0 +1,43 @@
+#include <libultraship/bridge.h>
+#include <libultraship/luslog.h>
+#include "2s2h/GameInteractor/GameInteractor.h"
+#include "2s2h/ShipInit.hpp"
+
+extern "C" {
+#include "variables.h"
+}
+
+#define CVAR_NAME "gEnhancements.Cutscenes.SkipMiscInteractions"
```
garrettjoecox: This might be better for StoryCutscenes

garrettjoecox (commented):
The code mostly looks good, but I'd like if we could get some thorough QA on this to ensure it doesn't break any vanilla behavior

mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipMoonCrash.cpp
JoshSanch: Updated

Eblo:
tl;dr I found many moon crash and saving issues, but none of them seem logically related to this enhancement, nor does anything in this enhancement's code seem like it could cause what I've seen

There were a number of things I wanted to verify still work correctly with this enhancement (whether Razor sword, bottles, stolen gear, and consumables honor the player's SoT reset configuration), but I can't because of other issues that are probably unrelated to this enhancement. I'm questioning how much I understand about the moon crash now since I don't think I've ever battle tested that aspect of this game.

> Getting a game over on the first cycle without saving after using Skip Intro Sequence creates a weird behavior that makes this cutscene skip seem strange - but it also occurs when the cutscene skip is not used. Seems to be an issue with the Skip Intro Sequence which is in use in all video examples here

image

I seem to be getting these issues no matter what my time saver settings are, and even if I have actually saved. This includes playing through the entire intro up to the first cycle, or playing in subsequent cycles. This happens even with "Skip Story Cutscenes" disabled, and therefore this enhancement does not run, so that much is probably unrelated.

- Various saves are just throwing me back to the start of the intro when I load them. Happens with first cycle pause saves (which may be unforeseen consequences of saving during a part of the game that wasn't built for it), but also happens with a legitimate Song of Time save from playing atop the Clock Tower. Since the moon crash isn't a factor here, again I think it's unrelated to this enhancement, but a big issue nonetheless.
- If I moon crash during first cycle, Link will always be human in the Clock Tower interior cutscene. The Happy Mask Salesman will be invisible for most of it as well. My inventory is wiped, but maybe that's a first cycle thing?
- If I moon crash after first cycle, the Clock Tower interior plays the "wow you got your instrument!" cutscene. Link will always be Deku as expected for that scene. I can do this multiple times in a row; it never seems to change what Happy Mask Salesman cutscene I get. This wipes some things from my inventory, including Deku Mask, Razor Sword, bottles, and other masks.

mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipMoonCrash.cpp
```diff
@@ -0,0 +1,39 @@
...
+#include "2s2h/ShipInit.hpp"
+
+extern "C" {
+#include "variables.h"
+}
+
+#define CVAR_NAME "gEnhancements.Cutscenes.SkipStoryCutscenes"
+#define CVAR CVarGetInteger(CVAR_NAME, 0)
+
+#define TERMINA_FIELD_MOON_CRASH_CS_ENTRANCE 0x54C0
```
Archez: I'm not a fan of these defines, instead we should just compare against the entrance values using the entrance macro.

If you _really_ want the define, then it should contain the `ENTRANCE` value instead of these hardcoded numbers

In this case its `ENTRANCE(TERMINA_FIELD, 12)`

mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipMoonCrash.cpp
```diff
@@ -0,0 +1,39 @@
...
+
+#define MOON_CRASH_NEW_CYCLE_CS_ID 13
+#define MOON_CRASH_NEW_CYCLE_ENTRANCE_ID 0xC030
```
Archez: Same as above, but this value is `ENTRANCE(CLOCK_TOWER_INTERIOR, 3)`

mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipMoonCrash.cpp
```diff
@@ -0,0 +1,39 @@
...
+void RegisterSkipMoonCrash() {
+    // Skips initial cutscene in Termina Field where the moon falls, goes straight to clock tower cutscene
+    COND_VB_SHOULD(VB_PLAY_TRANSITION_CS, CVAR, {
+        if (gSaveContext.save.cutsceneIndex == 0x0 &&
+            gSaveContext.save.entrance == TERMINA_FIELD_MOON_CRASH_CS_ENTRANCE) {
+            // This cutscene command would otherwise be run as part of the Fire Wall cutscene that gets skipped.
+            // IT MUST BE CALLED HERE IF THAT CUTSCENE IS SKIPPED OR ELSE THE GAME STATE WILL NOT RESET CORRECTLY!
+            Sram_ResetSaveFromMoonCrash(&gPlayState->sramCtx);
+
+            gSaveContext.save.cutsceneIndex = MOON_CRASH_NEW_CYCLE_CS_ID;
```
Archez: `gSaveContext.save.cutsceneIndex` should remain at `0` for the clock tower interior entrance.

This value `MOON_CRASH_NEW_CYCLE_CS_ID` being 13 appears to be a misunderstanding with the cutscene ID used by the salesman when the salesman inits. His logic handles choosing the right cs ID based on the players inventory.

garrettjoecox:
@JoshSanch looks like you've got some outstanding feedback here

balloondude2:
#1202 likely fixed the weird behavior you were seeing earlier. @JoshSanch do you plan on addressing any of the outstanding feedback on this?

JoshSanch:
I'll take some time tomorrow evening to update this and get it ready to merge.

mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipMoonCrash.cpp
```diff
@@ -0,0 +1,35 @@
+#include <libultraship/bridge.h>
+#include <libultraship/luslog.h>
```
Eblo: Change the bridge import and remove the unused luslog one.
```suggestion
#include "public/bridge/consolevariablebridge.h"
```

Eblo (changes requested):
Checks out to me playtesting-wise. Just needs some header cleanup.

- Moon crash in rando
- Moon crash in first cycle
- Moon crash after first cycle, before SoH
- Moon crash after SoH

Eblo (approved):

mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipMoonCrash.cpp
balloondude2: To add to this, because the skip only executes when `gSaveContext.save.cutsceneIndex` is `0` and `gSaveContext.save.cutsceneIndex` does not change in `Sram_ResetFromMoonCrash()`, I don't think this assignment or consequently the `#define MOON_CRASH_NEW_CYCLE_CS_ID 0` is necessary.

mm/2s2h/Enhancements/Cutscenes/StoryCutscenes/SkipMoonCrash.cpp
```diff
@@ -0,0 +1,34 @@
+#include "public/bridge/consolevariablebridge.h"
+#include "2s2h/GameInteractor/GameInteractor.h"
+#include "2s2h/ShipInit.hpp"
+
+extern "C" {
+#include "variables.h"
+}
+
+#define CVAR_NAME "gEnhancements.Cutscenes.SkipStoryCutscenes"
```
balloondude2, line 9: Can you move the enhancement file to the StoryCutscenes folder to match the enhancement classification?

balloondude2 (commented):
Thanks for the updates! Just a couple of very minor nitpicks.

Eblo:
Pinging again on this. I had approved this PR, but I agree with the feedback that came after my approval.

mckinlee (approved):

Eblo (approved):

PR #973, opened 2025-01-29
[TimeSaver] Fast Dampe Flame Digging

Patrick12115:
Only requires digging up one flame to spawn the big poe.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2523096157.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2523096443.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2523100723.zip)


Patrick12115:
An issue with toggling the enhancement on after already revealing a fire has been fixed and code has been simplified.

mm/2s2h/Enhancements/Timesavers/DampeDiggingSkip.cpp
```diff
@@ -18,18 +18,24 @@ void RegisterDampeDiggingSkip() {
...
         if (firePo->actionFunc == EnBigpo_RevealedFireIdle) {
-            EnBigpo_SetupSpawnCutscene((EnBigpo*)firePo->actor.parent);
+            dampeCutsceneTriggered = false;
 
-            Actor* dampeActor = SubS_FindActor(gPlayState, NULL, ACTORCAT_NPC, ACTOR_EN_TK);
-            if (dampeActor != NULL) {
-                EnTk* dampe = (EnTk*)dampeActor;
-                if (dampe->unk_2CA) {
-                    func_80AEDBEC(dampe, gPlayState);
+            if (!dampeCutsceneTriggered) {
```
Archez: It doesn't look like `dampeCutsceneTriggered` is actually doing anything in this hook. It is always set to `false` on every run, and so the `if (!dampeCutsceneTriggered) {}` will always execute, then at the end `dampeCutsceneTriggered` is set to `true`, but nothing ever uses it after.

These could probably all be dropped entirely, unless there was something else you were intending to use it for.

mm/2s2h/Enhancements/Timesavers/DampeDiggingSkip.cpp
```diff
@@ -18,18 +18,24 @@ void RegisterDampeDiggingSkip() {
...
+                EnBigpo_SetupSpawnCutscene((EnBigpo*)firePo->actor.parent);
```
Archez: This logic is running multiple times, rather than once. There doesn't seem to be any negative side effects from it, but I'd prefer to only execute once.

And the amount of times it executes is doubled if two fires were revealed and then the enhancement was turned on.

mm/2s2h/Enhancements/Timesavers/DampeDiggingSkip.cpp
```diff
@@ -0,0 +1,41 @@
...
+void EnBigpo_SetupSpawnCutscene(EnBigpo* thisx);
+void EnBigpo_RevealedFireIdle(EnBigpo* thisx, PlayState* play);
+}
+
+#define CVAR_NAME "gEnhancements.Timesavers.DampeDiggingSkip"
+#define CVAR CVarGetInteger(CVAR_NAME, 0)
+
+void RegisterDampeDiggingSkip() {
+
+    COND_ID_HOOK(ShouldActorDraw, ACTOR_EN_BIGPO, CVAR, [](Actor* actor, bool* should) {
```
Archez: This hook probably should not be using `ShouldActorDraw` unless it needed to influence something about the draw behavior. An Update hook instead would be preferred.

Archez (commented):
Based on a couple points from below, I'd like to propose a slightly tweaked version of this enhancement if that is fine with you. I can push directly to your branch.

mm/2s2h/Enhancements/Timesavers/DampeDiggingSkip.cpp
Patrick12115: I haven't gotten around to getting my stuff updated since the latest changes to test this real quick, but a big issue I was having was I was getting a camera lock after the big poe spawned and never actually triggering the Dampe cutscene if I turned the enhancement on after a flame was already dug up. It would work fine if on before I dug any up though. So as long as that isn't happening with any of these tweaks, do and push whatever you want haha

mm/2s2h/Enhancements/Timesavers/DampeDiggingSkip.cpp
Patrick12115: No, update is fine, I just didn't know which one would be prefered

Archez (approved):
Went ahead and pushed the small tweaks to this branch.
The main callouts are:
* Brought back the Dampe update hook you had and simplified it so that upon digging, it auto increments Dampe's flame counter to 3 so that it runs his logic for setting his cutscene and closing the dig textbox.
* Changes the Big Poe hook to look for the real bigpoe actor, instead of one of the flames, and performs a loop over the flames to see if any are revealed. I also kept the additional Dampe handling in this hook to retain the "toggling the enhancement after digging a flame"

This is good to merge from my end now 👍

PR #969, opened 2025-01-26
Fix Persistent Mask Crash on Talking to Postman

zodiac-ill:
Persistent Masks implementer forgot to include Player pointer as argument for GI SHOULD call at `z_en_mm3.c` line 470. Also edits `PersistentMasks.cpp` to prevent null deref.

To reproduce this crash, simply turn on Persistent Bunny Hood and talk to the postman while he is on bed.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2487822166.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2487823954.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2487825571.zip)

PR #979, opened 2025-02-01
Persistent Mask fixes

Eblo:
#969 made the Postman and his minigame act as if the bunny hood is *always* equipped, even when it is not. ~~I don't fully understand why, but if `GET_PLAYER(gPlayState)` is used instead of passing a `Player*` through the hook, it works as intended. This PR removes `Player*` as an argument for all uses of `VB_CONSIDER_BUNNY_HOOD_EQUIPPED` and just uses the global reference.~~

The other fix is to make it so that the active persistent mask cannot be offered as an item action, matching the vanilla behavior of the currently equipped mask. In effect, this also prevents the issue where the player could continue to use the persistent mask after giving up that mask's item on the Moon.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2521888998.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2521889360.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2521894192.zip)


zodiac-ill:
> https://github.com/HarbourMasters/2ship2harkinian/pull/874 made the Postman and his minigame act as if the bunny hood is always equipped, even when it is not. I don't fully understand why, but if GET_PLAYER(gPlayState) is used instead of passing a Player* through the hook, it works as intended.

This is because, for whatever reason, `GET_PLAYER(play)` is called from the postman code, it will sometimes return `nullpltr`. I do not know why this is. I noticed this when I was fixing a crash that happened here as a mistake caused the pointer not to be passed, resulting in a garbage pointer being dereferenced. Strangely when fixing this I noticed that `nullpltr` was now being dereferenced, so I just threw in a check to prevent null deref and left it at that.

Eblo:
> This is because, for whatever reason, `GET_PLAYER(play)` is called from the postman code, it will sometimes return `nullpltr`. I do not know why this is. I noticed this when I was fixing a crash that happened here as a mistake caused the pointer not to be passed, resulting in a garbage pointer being dereferenced. Strangely when fixing this I noticed that `nullpltr` was now being dereferenced, so I just threw in a check to prevent null deref and left it at that.

Oops, I quoted the wrong PR. That explains a little bit.

Archez:
Removing player is not allowed. The reason player was passed in is because this hook is used in the common player code and therefore also used by kafei actor. We _do not_ want kafei being sped up by the persistent bunny hood.

What is the issue here?

Archez:
The issue I think is that there is still a bad parenthesis order happening and one of the player args in the postman code is accidentally outside the Should call.

GET_PLAYER should _never_ return null. If we are getting NULL in the hook side, then there is a mistake in the usage.

Eblo:
> What is the issue here?

Postman always reacts as if Link is wearing Bunny Hood, and always displays the timer, even if the player does not have the Bunny Hood. Screenshots from a fresh first cycle save, no Bunny Hood:

image
image

>The issue I think is that there is still a bad parenthesis order happening and one of the player args in the postman code is accidentally outside the Should call.

I'll take another look at this, thanks for the Kafei context.

Eblo:
Yep, that's all it was. I reverted my changes so that it uses the `Player*` arg as it should.

Archez (approved):

PR #985, opened 2025-02-04
Fix button id conflicts on input editor

Archez:
A couple of elements needed the `##` pattern in the input editor to prevent imgui collisions

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2536409800.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2536413732.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2536427230.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenInputEditorWindow.cpp
+++ b/mm/2s2h/BenGui/BenInputEditorWindow.cpp
@@ -630,10 +630,13 @@ void BenInputEditorWindow::DrawStickSection(uint8_t port, uint8_t stick, int32_t
 
     ImGui::SameLine();
     ImGui::BeginGroup();
-    DrawStickDirectionLine(ICON_FA_ARROW_UP, port, stick, Ship::UP, color);
-    DrawStickDirectionLine(ICON_FA_ARROW_DOWN, port, stick, Ship::DOWN, color);
-    DrawStickDirectionLine(ICON_FA_ARROW_LEFT, port, stick, Ship::LEFT, color);
-    DrawStickDirectionLine(ICON_FA_ARROW_RIGHT, port, stick, Ship::RIGHT, color);
+    DrawStickDirectionLine(StringHelper::Sprintf("%s##%d", ICON_FA_ARROW_UP, id).c_str(), port, stick, Ship::UP, color);
+    DrawStickDirectionLine(StringHelper::Sprintf("%s##%d", ICON_FA_ARROW_DOWN, id).c_str(), port, stick, Ship::DOWN,
+                           color);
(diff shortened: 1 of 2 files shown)


mm/2s2h/BenGui/BenInputEditorWindow.cpp
```diff
@@ -630,10 +630,13 @@ void BenInputEditorWindow::DrawStickSection(uint8_t port, uint8_t stick, int32_t
...
-    DrawStickDirectionLine(ICON_FA_ARROW_DOWN, port, stick, Ship::DOWN, color);
-    DrawStickDirectionLine(ICON_FA_ARROW_LEFT, port, stick, Ship::LEFT, color);
-    DrawStickDirectionLine(ICON_FA_ARROW_RIGHT, port, stick, Ship::RIGHT, color);
+    DrawStickDirectionLine(StringHelper::Sprintf("%s##%d", ICON_FA_ARROW_UP, id).c_str(), port, stick, Ship::UP, color);
+    DrawStickDirectionLine(StringHelper::Sprintf("%s##%d", ICON_FA_ARROW_DOWN, id).c_str(), port, stick, Ship::DOWN,
+                           color);
+    DrawStickDirectionLine(StringHelper::Sprintf("%s##%d", ICON_FA_ARROW_LEFT, id).c_str(), port, stick, Ship::LEFT,
+                           color);
+    DrawStickDirectionLine(StringHelper::Sprintf("%s##%d", ICON_FA_ARROW_RIGHT, id).c_str(), port, stick, Ship::RIGHT,
+                           color);
```
briaguya0: does this fix both left and right stick being open at the same time?

briaguya0: i guess it probably should because it's using `id`, just figured i'd ask

Archez: Yeah, `id` is either 0 or 1 for the specific sticks.
(`id` also could just be dropped in favor of using `stick` instead, which also is just 0 or 1 for the stick enum, not sure if there was a reason to have both values explicitly)

briaguya0: i don't remember the exact reasoning, i did my best to keep the imgui code in these files decently clean but there's a lot going on so it makes sense some things slipped through

briaguya0 (approved):

PR #995, opened 2025-02-13
Add missing flag to keep monkeys from spawning again

balloondude2:
This skip wasn't setting a switch flag that normally gets set when the skip isn't used, causing the monkeys to keep spawning in.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2590363062.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2590364499.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2590385449.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipBothersomeMonkey.cpp
+++ b/mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipBothersomeMonkey.cpp
@@ -17,6 +17,7 @@ void RegisterSkipBothersomeMonkey() {
         EnMnk* monkey = (EnMnk*)va_arg(args, EnMnk*);
         *should = false;
         SET_EVENTINF(EVENTINF_25);
+        Flags_SetSwitch(gPlayState, 0x02);
         monkey->actionFunc = EnMnk_Monkey_WaitToRun;
     });
 }
--- a/mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipBothersomeMonkey.cpp
+++ b/mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipBothersomeMonkey.cpp
@@ -32,7 +32,6 @@ void RegisterSkipBothersomeMonkey() {
             }
             monkey->actionFunc = EnMnk_Monkey_SetupRunAfterTalk;
         }
-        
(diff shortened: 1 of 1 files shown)


Archez:
Based on where `VB_MONKEY_WAIT_TO_TALK_AFTER_APPROACH` is used, it seems like there could be at least two possible sets of monkeys that trigger it: `MONKEY_OUTSIDECHAMBER` and `MONKEY_OUTSIDEWOODS`. We need to decide if we should actually make the hook work for both sets of monkeys, or filter for the oustide woods type.

And based on the src code, I think it would be preferable to set the flag the same way as it is in vanilla (and would support both monkey sets):
```c
if (MONKEY_GET_SWITCH_FLAG(&this->picto.actor) != 0x7F) {
    Flags_SetSwitch(play, MONKEY_GET_SWITCH_FLAG(&this->picto.actor));
}
```

And another thing worth calling out, when the monkey sets the `EVENTINF_25` it is also setting `WEEKEVENTREG_79_02`, which is missing from the skip hook.

balloondude2:
I conflated `MONKEY_OUTSIDEPALACE` and `MONKEY_OUTSIDECHAMBER`, so when I tested they still seemed fine, so I'll look into that some more. I agree that the hook should work for both cases or have an explicit filter for `MONKEY_OUTSIDEWOODS`. 

> And another thing worth calling out, when the monkey sets the `EVENTINF_25` it is also setting `WEEKEVENTREG_79_02`, which is missing from the skip hook.

I looked at setting `WEEKEVENTREG_79_02`, but noticed that Proxy had removed that when he last refactored and assumed he had some reason for that. Looking at his refactor more closely, he also removed a `Flags_SetSwitch(play, MONKEY_GET_SWITCH_FLAG(&monkey->picto.actor));`, so I'll check with him on his reasoning for removing those initially.

balloondude2:
The skip now accounts for both `MONKEY_OUTSIDEWOODS` and `MONKEY_OUTSIDECHAMBER` and sets the appropriate flags.

Archez (approved):
LGTM thanks for making those changes!

PR #778, opened 2024-09-24
Custom Audio Support

louist103:
Adds custom audio support.
 - Adds support for the classic custom audio using custom .aseq files and meta files (sequenced audio).
 - Adds support for mp3, wav, flac, etc audio sequences (streamed audio).
 - Exports all audio assets, sequences, sound fonts, and samples as XML for easier modifications.
 - Names samples that were copied from OOT.
 - 
 Needs:
- [ ] General cleanups
- [x] Fix GC XMLs
- [ ] Audio editor is still a little gitchy
- [x] https://github.com/louist103/ZAPDTR/pull/20 Merged
- [x] https://github.com/louist103/OTRExporter/pull/31 and https://github.com/louist103/OTRExporter/pull/29 Merged

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2679102873.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2679136110.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2679209792.zip)

PR #999, opened 2025-02-19
[SFX] Mute Low HP Alarm, Carpenters Sounds, and Crying Goron Child

Patrick12115:
Adds SFX options to mute Low HP Alarm, the carpenter sounds heard in South and East Clock Town from the tower being built, and the crying Goron Child inside the Goron Shrine. He will still cry during certain text when outside and inside the Goron Racetrack, as these are easily skippable and don't last long.

Added an SFX section in Gameplay below modes, until #778 is merged in and can be relocated.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2633850052.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2633857051.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Sfx/MuteCarpenterSfx.cpp
+++ b/mm/2s2h/Enhancements/Sfx/MuteCarpenterSfx.cpp
@@ -1,0 +1,28 @@
+#include <libultraship/bridge.h>
+#include "2s2h/GameInteractor/GameInteractor.h"
+#include "2s2h/ShipInit.hpp"
+
+extern "C" {
+#include "overlays/actors/ovl_Obj_Sound/z_obj_sound.h"
+}
+
+#define CVAR_NAME "gEnhancements.Sfx.MuteCarpenterSfx"
+#define CVAR CVarGetInteger(CVAR_NAME, 0)
+
+void RegisterMuteCarpenterSfx() {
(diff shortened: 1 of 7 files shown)


Patrick12115:
Added Crying Goron Child

mm/2s2h/Enhancements/Sfx/MuteCarpenterSfx.cpp
```diff
@@ -0,0 +1,28 @@
...
+        ObjSound* objSound = (ObjSound*)actor;
+
+        // Returning if not in South or East Clocktown just as a safety precaution, incase these actor params are used
+        // elsewhere
+        if (gPlayState->sceneId != SCENE_CLOCKTOWER && gPlayState->sceneId != SCENE_TOWN) {
+            return;
+        }
+
+        if ((objSound->actor.params & OBJ_SOUND_ID_MASK) == 16) {
+            Actor_Kill(&objSound->actor);
```
Archez: [nit]: minor comment about the behavior of the `ShouldActorInit` hook, setting `*should = false` will prevent the actor from spawning (by calling `Actor_Kill()` for you), but it also ensures that the actual actors init func is never executed.

In this case, nothing significant is happening in obj_sound init func, but it would be good practice to set the should result instead.

```suggestion
            *should = false;
```

Archez (approved):
Approving, but left a suggestion below

PR #1001, opened 2025-02-21
Add Hook Debugger Window

Archez:
Ports over https://github.com/HarbourMasters/Shipwright/pull/4323 from SoH, with some slight tweaks

Tweaks include:
* Removing user parent directories from hook info
* Better default table formatting/sizing to work out of the box with the modern menu
* Using `map` for the hook info so that the debugger has a stable order

image

I haven't measure this, so I don't know if this introduces any significant performance impacts, but if it did I could see argument about putting the debugging structs behind a compile time variable so devs would have to explicitly enable it.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2633099140.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2633103053.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2633114521.zip)

### DIFF
--- a/mm/2s2h/GameInteractor/GameInteractor.h
+++ b/mm/2s2h/GameInteractor/GameInteractor.h
@@ -178,11 +178,6 @@ struct HookRegisteringInfo {
 struct HookInfo {
     uint32_t calls;
     HookRegisteringInfo registering;
-
-    HookInfo() : calls(0), registering(HookRegisteringInfo{}) {
-    }
-    HookInfo(HookRegisteringInfo _registering) : calls(0), registering(_registering) {
-    }
 };
 
 #ifdef __cpp_lib_source_location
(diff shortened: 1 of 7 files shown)


mm/2s2h/GameInteractor/GameInteractor.h
```diff
@@ -131,17 +131,66 @@ typedef enum {
...
+
+struct HookInfo {
+    uint32_t calls;
+    HookRegisteringInfo registering;
+
+    HookInfo() : calls(0), registering(HookRegisteringInfo{}) {
+    }
+    HookInfo(HookRegisteringInfo _registering) : calls(0), registering(_registering) {
     }
+};
```
garrettjoecox: This is just personal preference, but I find these struct definitions difficult to parse, and not really worth it. Are we getting anything out of it? Seems like the difference at the caller would be
```cpp
HookInfo{ GET_CURRENT_REGISTERING_INFO(HOOK_TYPE_NORMAL) }
vs
HookInfo{ 0, GET_CURRENT_REGISTERING_INFO(HOOK_TYPE_NORMAL) }
```

garrettjoecox (approved):
small opinion based thing but otherwise LGTM

mm/2s2h/GameInteractor/GameInteractor.h
Archez: Mainly just copy and paste from soh. They are simple constructors. I can drop these, but I think the ones for `HookRegisteringInfo` are worth having (cause there is more logic going on in construction).

PR #1004, opened 2025-02-22
Implement ActorExtension functionality

garrettjoecox:
Inspired by a new pattern introduced in recomp, this system allows us to store arbitrary data associated with each actor, outside of their actor struct. This will be preferred going forward instead of modifying the original actor's struct or hijacking unused variables like we have done in the past.

This also includes the ability to fetch a given actor's ActorListIndex (the index at which this actor was stored in the scene data), while I have marked this as an "example", it will be immensely helpful in identifying actors going forward for things like Rando.

Any and all feedback is welcome!

This is how I foresee using the ActorExtension in rando:
```cpp
ActorExtensionId randoCheckIdActorExt;

void SetActorRandoCheckId(Actor* actor, RandoCheckId randoCheckId) {
    RandoCheckId* randoCheckId = (RandoCheckId*)ActorExtension_Get(actor, randoCheckIdActorExt);
    if (randoCheckId != NULL) {
        *randoCheckId = randoCheckId;
    }
}

RandoCheckId GetActorRandoCheckId(Actor* actor) {
    RandoCheckId* randoCheckId = (RandoCheckId*)ActorExtension_Get(actor, randoCheckIdActorExt);
    if (randoCheckId != NULL) {
        return *randoCheckId;
    }

    return RC_UNKNOWN;
}


static RegisterShipInitFunc initFunc([]() {
    if (randoCheckIdActorExt == 0) {
        randoCheckIdActorExt = ActorExtension_CreateForAll(sizeof(RandoCheckId));
    }

    COND_ID_HOOK(ShouldActorInit, ACTOR_OBJ_TSUBO, IS_RANDO, [](Actor* actor, bool* should) {
        SetRandoCheckId(actor, IdentifyPotRC(actor));
    });

    COND_VB_SHOULD(VB_POT_DROP_ITEM, IS_RANDO, {
        RandoCheckId randoCheckId = GetRandoCheckId(actor);

        if (randoCheckId == RC_UNKNOWN) return;

        *should = false;
        // SpawnItem etc
    });
}, { "IS_RANDO" });
```

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2676135420.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2676150958.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2676212156.zip)

### DIFF
--- a/mm/2s2h/ActorExtension/ActorExtension.cpp
+++ b/mm/2s2h/ActorExtension/ActorExtension.cpp
@@ -1,0 +1,79 @@
+
+#include "ActorExtension.h"
+
+ActorExtensionId sNextActorExtensionId = 0;
+
+struct ActorExtensionKeyHash {
+    std::size_t operator()(const std::pair<Actor*, ActorExtensionId>& key) const {
+        return std::hash<Actor*>{}(key.first) ^ (std::hash<ActorExtensionId>{}(key.second) << 1);
+    }
+};
+
+std::unordered_map<ActorExtensionId, size_t> sGlobalActorExtensionSizes;
+std::unordered_map<s16, std::unordered_map<ActorExtensionId, size_t>> sActorExtensionSizes;
+std::unordered_map<std::pair<Actor*, ActorExtensionId>, void*, ActorExtensionKeyHash> sActorExtensionData;
+
+void* ActorExtension_Get(Actor* actor, ActorExtensionId id) {
+    if (actor == NULL) {
+        return NULL;
+    }
+
+    auto it = sActorExtensionData.find(std::make_pair(actor, id));
+    if (it == sActorExtensionData.end()) {
+        return NULL;
+    }
+
+    return it->second;
+}
+
+ActorExtensionId ActorExtension_CreateForId(s16 actorId, size_t size) {
+    ActorExtensionId id = ++sNextActorExtensionId;
+    sActorExtensionSizes[actorId][id] = size;
+    return id;
+}
+
+ActorExtensionId ActorExtension_CreateForAll(size_t size) {
+    ActorExtensionId id = ++sNextActorExtensionId;
+    sGlobalActorExtensionSizes[id] = size;
+    return id;
+}
+
+void ActorExtension_Alloc(Actor* actor, s16 actorId) {
+    if (actor == NULL) {
+        return;
+    }
+
+    for (auto& [id, size] : sGlobalActorExtensionSizes) {
+        void* data = malloc(size);
+        memset(data, 0, size);
+        sActorExtensionData[std::make_pair(actor, id)] = data;
+    }
+
+    for (auto& [id, size] : sActorExtensionSizes[actorId]) {
+        void* data = malloc(size);
+        memset(data, 0, size);
+        sActorExtensionData[std::make_pair(actor, id)] = data;
+    }
+}
+
+void ActorExtension_Free(Actor* actor) {
+    if (actor == NULL) {
+        return;
+    }
+
+    for (auto& [id, size] : sGlobalActorExtensionSizes) {
+        auto it = sActorExtensionData.find(std::make_pair(actor, id));
+        if (it != sActorExtensionData.end()) {
+            free(it->second);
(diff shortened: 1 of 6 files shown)


mm/2s2h/ActorExtension/ActorExtension.cpp
```diff
@@ -0,0 +1,106 @@
...
+ActorExtensionId actorListIndexActorExt;
+s16 currentActorListIndex = 0;
+
+static RegisterShipInitFunc initFunc(
+    []() {
+        HOOK(OnPortInit, []() { // Only ever called once
+            actorListIndexActorExt = ActorExtension_CreateForAll(sizeof(s16));
+        });
+    },
+    {});
```
Archez: This feels redundant. Why create a `OnPortInit` when `ShipInit::InitAll` is called just before this anyways? It leaves two long living callbacks registered, instead of one.

Also the `HOOK()` does not unregister, so if `ShipInit::InitAll` is executed again, that causes there to be multiple+ hooks registered (even if nothing will ever trigger them).

Regardless, this should have a static var involved in some fashion to prevent multiple registers, or even just reuse the `currentActorListIndex` id itself.

My suggestion would be to drop `OnPortInit` entirely and just write the ShipInit like this:

```suggestion
ActorExtensionId actorListIndexActorExt = 0;
s16 currentActorListIndex = 0;

static RegisterShipInitFunc initFunc(
    []() {
        if (actorListIndexActorExt == 0) {
            actorListIndexActorExt = ActorExtension_CreateForAll(sizeof(s16));
        }
    },
    {});
```

mm/2s2h/ActorExtension/ActorExtension.cpp
```diff
@@ -0,0 +1,106 @@
...
+
+// The memory in the actor extension is zeroed out by default, we make use of this to indicate nothing has been stored,
+// so this actor was spawned outside of an ActorList command. In return, when we store the actual ActorList index, we
+// store it + 1, then undo that offset here.
```
Archez: Could you just change the `currentActorListIndex` to default/reset to `-1`? I think that should be simple enough to drop this `+/- 1` logic off the stored index value.

mm/src/code/z_actor.c
```diff
@@ -3552,6 +3558,13 @@ Actor* Actor_SpawnAsChildAndCutscene(ActorContext* actorCtx, PlayState* play, s1
         return NULL;
     }
 
+    // #region 2S2H [ActorExtension]
+    ActorExtension_Alloc(actor, actorInit->id);
+    s16* actorListIndex = ActorExtension_Get(actor, actorListIndexActorExt);
```
Archez: This doesn't have null check safety. Probably consider making and exporting a `SetActorListIndex` for setting on the .cpp side.

mm/2s2h/ActorExtension/ActorExtension.cpp
```diff
@@ -0,0 +1,106 @@
...
+    for (auto& [id, size] : sActorExtensionSizes[actor->id]) {
+        auto it = sActorExtensionData.find(std::make_pair(actor, id));
+        if (it != sActorExtensionData.end()) {
+            free(it->second);
+            sActorExtensionData.erase(it);
+        }
+    }
+}
+
+// Example usage, the storing of the actorListIndex happens in Actor_SpawnAsChildAndCutscene:
```
Archez: Maybe ActorList stuff should split to a self contained file and to isolate header definitions?

Archez (commented):
Overall I think this is in a good direction. I'd still like to try out what it would be like to malloc/free using the Zelda Arena instead, but thats an implementation detail that can checked later and shouldn't have any impact to consumers.

Left some questions and suggestions below.

garrettjoecox:
> Left some questions and suggestions below.

Feedback applied

Archez (approved):

PR #1029, opened 2025-02-26
Port "Copy Assets with "If Different" Variants, CMake bump" from SoH

balloondude2:
Ports https://github.com/HarbourMasters/Shipwright/pull/4714 to reduce build times. Also bumps minimum required CMake version to 3.26

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2653557344.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2653599989.zip)

### DIFF
--- a/mm/CMakeLists.txt
+++ b/mm/CMakeLists.txt
@@ -1,4 +1,4 @@
-﻿cmake_minimum_required(VERSION 3.16.0 FATAL_ERROR)
+﻿cmake_minimum_required(VERSION 3.26.0 FATAL_ERROR)
 
 set(CMAKE_SYSTEM_VERSION 10.0 CACHE STRING "" FORCE)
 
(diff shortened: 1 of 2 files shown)


mm/CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-﻿cmake_minimum_required(VERSION 3.16.0 FATAL_ERROR)
+﻿cmake_minimum_required(VERSION 3.26.0 FATAL_ERROR)
```
Archez, line 1: Mind also up-syncing the version number in the root CMakeLists.txt too for consistency

Archez (approved):

mm/CMakeLists.txt
balloondude2: Yep, got that pushed up

PR #1031, opened 2025-02-27
[Rando] Adds new Dungeon Access Options

Patrick12115:
Changes the combo box to two check boxes where you can select Require Tranformation, Require Song, Require Both or Require Neither. Selecting both is the Vanilla behavior of needing both to access the dungeon and selecting neither will make all 3 dungeons accessible with no requirements, so Woodfall Temple will already be raised, Biggoron will be gone, and the turtle will already be spawned.

Edit: This also fixes the bug where you can continue to play ocarina notes after activating the dungeon access

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2674170108.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2674189539.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2674374612.zip)


garrettjoecox (approved):
We collaborated on this, the code is a little messy but it seemed to work. Would like some thorough QA (I know pat has done plenty, but always good to have a fresh set of eyes)

balloondude2:
Notes from my testing: 

1. `Requires Only Transformation` or `Requires Transformation or Song` requires you to pull out an instrument to open the dungeon. This doesn't seem to be accounted for in the logic, where `hasFormAccess` is set to `CAN_BE_transformation`. 
2. In `Requires Only Transformation` or `Requires Transformation or Song` pulling out an instrument in the correct form at Woodfall and Snowhead automatically ended playing and opened the dungeon. For Great Bay, I had to manually exit out of playing. Not blocking for me, but I'd prefer it to be similar for all three places. 
3. This is more of an observation, but on `Open`, if Misc Skips aren't turned on, the Tatl text at Snowfall doesn't make a lot of sense.

Patrick12115:
> Notes from my testing:
> 
> 1. `Requires Only Transformation` or `Requires Transformation or Song` requires you to pull out an instrument to open the dungeon. This doesn't seem to be accounted for in the logic, where `hasFormAccess` is set to `CAN_BE_transformation`.
> 2. In `Requires Only Transformation` or `Requires Transformation or Song` pulling out an instrument in the correct form at Woodfall and Snowhead automatically ended playing and opened the dungeon. For Great Bay, I had to manually exit out of playing. Not blocking for me, but I'd prefer it to be similar for all three places.
> 3. This is more of an observation, but on `Open`, if Misc Skips aren't turned on, the Tatl text at Snowfall doesn't make a lot of sense.



Fixed number 1

Can't seem to replicate number 2 on my machine

I think I will leave number 3 to the decision of others as to whether this should be handled here or just leave it

Eblo:
From my testing:

1. `Requires Transformation and Song` works just like vanilla as it should.
2. `Requires Transformation or Song` seems to work as described. Either any form must play the relevant song, or the relevant form must pull out the instrument.
3. `Requires only Transformation` works as soon as the relevant form pulls out the instrument. Nothing happens if a different form plays the relevant song.
4. `Requires only Song` works if the correct song was played, regardless of form.
5. `Open` works as described on all three temples. Things are available upon scene load

> In `Requires Only Transformation` or `Requires Transformation or Song` pulling out an instrument in the correct form at Woodfall and Snowhead automatically ended playing and opened the dungeon. For Great Bay, I had to manually exit out of playing. Not blocking for me, but I'd prefer it to be similar for all three places.

I also cannot replicate this.

PR #1028, opened 2025-02-25
[Rando] Shuffle Snowballs

Caladius:
Continuing my exhibition of Shuffling Termina with Snowballs this time.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2650157317.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2650161338.zip)

PR #1034, opened 2025-03-02
Shuffle Snowball Drops

garrettjoecox:
Reupload of #1028 

Needs a final QA then can be merged in

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2676611627.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2676612070.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2676628415.zip)


mm/2s2h/Rando/ActorBehavior/ObjSnowball.cpp
```diff
@@ -0,0 +1,202 @@
...
+        Actor* actor = va_arg(args, Actor*);
+
+        RandoCheckId randoCheckId = GetActorRandoCheckId(actor);
+
+        if (randoCheckId == RC_UNKNOWN) {
+            return;
+        }
+
+        if (!RANDO_SAVE_CHECKS[randoCheckId].cycleObtained) {
+            SpawnSnowballDrop(actor->home.pos, randoCheckId);
```
balloondude2, line 185: Probably want `actor->world.pos` here, and also the other `SpawnSnowballDrop()` call on line:199

sitton76:
Noticed two issues while testing.

1. Throwing the snowball makes the item spawn in the spawn location of the snowball rather then where it landed.
2. For the Twin Island area of the mountains checks from the big snow boulders only seem to work on the first day prob due to the actor layout being slightly different since the Elder moves around to a different spot each day(Twin islands in two spots day 1 and 2, and Mountain Village in day 3) but to that I suspect this might also apply to the Mountain Village too but have not verified yet.

PR #1035, opened 2025-03-02
Add Enemy Health Bars using actor extension

Archez:
Adds enemy health bars leveraging the Actor Extension system.

Actors upon init will have the collider health value tracked as their "maximum" health, which is used to compute how much health they have left for the health bars.

Only a handful of actors were identified as needing special casing, otherwise everything appears to "just work" (I haven't hand verified all enemies, but bosses looked good from my testing).

Adds an `OnInterfaceDrawStart()` hook so that I can render the health bars before the rest of the HUD, this way the HUD displays on top of the bars.

Existing health bar options like anchoring, size, offset, and coloring are not yet implemented from SoH, I think I want to wait to see how the cosmetic editor and HUD editor evolve first before adding special options.

image

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2676631126.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2676638451.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2676694197.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Graphics/EnemyHealthBars.cpp
+++ b/mm/2s2h/Enhancements/Graphics/EnemyHealthBars.cpp
@@ -1,0 +1,212 @@
+
+#include "2s2h/ActorExtension/ActorExtension.h"
+#include "2s2h/GameInteractor/GameInteractor.h"
+#include "2s2h/Enhancements/FrameInterpolation/FrameInterpolation.h"
+#include "2s2h/ShipInit.hpp"
+#include "2s2h/ShipUtils.h"
+
+extern "C" {
+#include "assets/interface/parameter_static/parameter_static.h"
+#include "overlays/kaleido_scope/ovl_kaleido_scope/z_kaleido_scope.h"
+#include "overlays/actors/ovl_Boss_02/z_boss_02.h"
+#include "overlays/actors/ovl_Boss_07/z_boss_07.h"
+#include "functions.h"
+#include "macros.h"
+#include "variables.h"
+}
+
+#define CVAR_NAME "gEnhancements.Graphics.EnemyHealthBars"
+#define CVAR CVarGetInteger(CVAR_NAME, 0)
+
+typedef enum {
+    ENEMYHEALTH_ANCHOR_ACTOR,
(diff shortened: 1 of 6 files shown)


mm/2s2h/Enhancements/Graphics/EnemyHealthBars.cpp
```diff
@@ -0,0 +1,212 @@
...
+// Draws an enemy health bar using the magic bar textures and positions it in a similar way to Z-Targeting
+void Interface_DrawEnemyHealthBar(TargetContext* targetCtx, PlayState* play) {
+    InterfaceContext* interfaceCtx = &play->interfaceCtx;
+    PauseContext* pauseCtx = &play->pauseCtx;
+    Player* player = GET_PLAYER(play);
+    Actor* actor = targetCtx->lockOnActor;
+    Actor* healthActor = actor;
+
+    if (actor == NULL || (actor->category != ACTORCAT_ENEMY && actor->category != ACTORCAT_BOSS) ||
+        targetCtx->lockOnAlpha == 0 || ((s32)pauseCtx->state > PAUSE_STATE_OPENING_2)) {
```
garrettjoecox, line 50: What is the significance of `pauseCtx->state > PAUSE_STATE_OPENING_2`?

garrettjoecox (approved):

mm/2s2h/Enhancements/Graphics/EnemyHealthBars.cpp
Archez: Don't want health bars displaying over kaleido, since they are rendered on OVERLAY instead of OPA.

This is the same logic that stops minimap from displaying when paused.

garrettjoecox: Gotcha, yeah just wasn't sure why not just `pauseCtx->state != PAUSE_STATE_OFF`

PR #1038, opened 2025-03-03
[Enhancement] Skip Enemy Cutscenes

Eblo:
Some boss cutscene skips existed, but this expands the scope to include most bosses and mini-bosses in the game.

* Add new VB, `VB_ENEMY_CUTSCENE_ACTION`, that exists to be used within actor files. In most cases, it is a simple set to false to skip vanilla behavior. Other uses involve setting state information as part of the phase change cutscene skip, or a death cutscene skip. These uses were not covered by the existing cutscene hooks; some actors use their own timers and camera control for instance.
* Renamed boss skips to enemy skips, since the scope increased over what was there before
* Separated from story cutscene skips into new setting specifically for enemy cutscene skips

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3677073965.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3677085911.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3677166842.zip)


mm/src/overlays/actors/ovl_Boss_06/z_boss_06.c
```diff
@@ -182,9 +183,11 @@ void func_809F23CC(Boss06* this) {
 }
 
 void func_809F24A8(Boss06* this) {
-    this->actionFunc = func_809F24C8;
-    this->unk_A28 = 0.0f;
-    this->unk_1AC = 0.0f;
+    if (GameInteractor_Should(VB_ENEMY_CUTSCENE_ACTION, true, this)) {
```
Archez: For functions where the entire body is wrapped with the GameInteractor, I would prefer to instead have an early return with the result inverted, this way we don't have to indent all the original src code.

```c
if (!GameInteractor_Should(VB_ENEMY_CUTSCENE_ACTION, true, this)) {
    return;
}
```

mm/src/overlays/actors/ovl_En_Bigslime/z_en_bigslime.c
```diff
@@ -819,24 +820,28 @@ void EnBigslime_JerkCameraPlayerHit(EnBigslime* this, PlayState* play) {
  * then zooms into the Gekko until the Gekko calls the minislimes down from the ceiling
  */
 void EnBigslime_UpdateCameraIntroCs(EnBigslime* this, PlayState* play, s32 noticeTimer) {
-    Camera* subCam = Play_GetCamera(play, this->subCamId);
-    Vec3f subCamEye;
-    f32 zoom = (noticeTimer * 19.0f) + 67.0f;
-    s16 yawOffset = this->actor.yawTowardsPlayer + (noticeTimer * 0x31);
+    if (GameInteractor_Should(VB_ENEMY_CUTSCENE_ACTION, true, this)) {
```
Archez: Same here

mm/src/overlays/actors/ovl_En_Bigslime/z_en_bigslime.c
```diff
@@ -819,24 +820,28 @@ void EnBigslime_JerkCameraPlayerHit(EnBigslime* this, PlayState* play) {
...
+    }
 }
 
 /**
  * Takes the camera and makes the focus point (at) point at bigslime, who is on the
  * center of the roof. This is used when the minislimes merges into bigslime.
  */
 void EnBigslime_UpdateCameraFormingBigslime(EnBigslime* this, PlayState* play) {
-    Play_SetCameraAtEye(play, this->subCamId, &this->actor.focus.pos, &Play_GetCamera(play, this->subCamId)->eye);
+    if (GameInteractor_Should(VB_ENEMY_CUTSCENE_ACTION, true, this)) {
```
Archez: Same here

mm/src/overlays/actors/ovl_En_Pametfrog/z_en_pametfrog.c
```diff
@@ -324,26 +326,30 @@ s32 func_8086A2CC(EnPametfrog* this, CollisionPoly* floorPoly) {
...
-    Camera* subCam = Play_GetCamera(play, this->subCamId);
-    s16 subCamYaw;
-    Vec3f subCamEye;
-
-    subCamYaw = BINANG_ROT180(Camera_GetCamDirYaw(subCam));
-    subCamEye.x = (Math_SinS(subCamYaw) * magShakeXZ) + subCam->at.x;
-    subCamEye.y = subCam->at.y + magShakeY;
-    subCamEye.z = (Math_CosS(subCamYaw) * magShakeXZ) + subCam->at.z;
-    Play_SetCameraAtEye(play, this->subCamId, &subCam->at, &subCamEye);
+    if (GameInteractor_Should(VB_ENEMY_CUTSCENE_ACTION, true, this)) {
```
Archez: Same here

mm/src/overlays/actors/ovl_En_Pametfrog/z_en_pametfrog.c
```diff
@@ -324,26 +326,30 @@ s32 func_8086A2CC(EnPametfrog* this, CollisionPoly* floorPoly) {
...
+        subCamEye.x = (Math_SinS(subCamYaw) * magShakeXZ) + subCam->at.x;
+        subCamEye.y = subCam->at.y + magShakeY;
+        subCamEye.z = (Math_CosS(subCamYaw) * magShakeXZ) + subCam->at.z;
+        Play_SetCameraAtEye(play, this->subCamId, &subCam->at, &subCamEye);
+    }
 }
 
 void EnPametfrog_StopCutscene(EnPametfrog* this, PlayState* play) {
-    Camera* subCam;
+    if (GameInteractor_Should(VB_ENEMY_CUTSCENE_ACTION, true, this)) {
```
Archez: Same here

mm/src/overlays/actors/ovl_En_Wiz/z_en_wiz.c
```diff
@@ -832,20 +833,22 @@ void EnWiz_Dance(EnWiz* this, PlayState* play) {
 }
 
 void EnWiz_SetupSecondPhaseCutscene(EnWiz* this, PlayState* play) {
-    s16 secondPhaseCsId = CutsceneManager_GetAdditionalCsId(this->actor.csId);
+    if (GameInteractor_Should(VB_ENEMY_CUTSCENE_ACTION, true, this)) {
```
Archez: Same here

mm/src/overlays/actors/ovl_En_Bigslime/z_en_bigslime.c
```diff
@@ -2368,11 +2377,13 @@ void EnBigslime_CutsceneDefeat(EnBigslime* this, PlayState* play) {
         EnBigslime_SetupGekkoDespawn(this, play);
     } else {
         // Continue for the camera to follow Gekko as it spins in defeat
-        subCam = Play_GetCamera(play, this->subCamId);
-        subCamAt.x = this->actor.world.pos.x;
-        subCamAt.y = this->actor.world.pos.y + 40.0f;
-        subCamAt.z = this->actor.world.pos.z;
-        Play_SetCameraAtEye(play, this->subCamId, &subCamAt, &subCam->eye);
+        if (GameInteractor_Should(VB_ENEMY_CUTSCENE_ACTION, true, this)) {
```
Archez: Similarly, since there is no more code after this, this could also be turned into an early return (inverted result).

mm/src/overlays/actors/ovl_En_Bigslime/z_en_bigslime.c
```diff
@@ -2407,12 +2418,14 @@ void EnBigslime_GekkoDespawn(EnBigslime* this, PlayState* play) {
     if (this->despawnTimer == 0) {
         EnBigslime_SetupFrogSpawn(this, play);
     } else {
-        subCam = Play_GetCamera(play, this->subCamId);
-        Math_Vec3f_Copy(&subCamAt, &subCam->at);
-        Math_Vec3f_Diff(&subCam->eye, &this->subCamDistToFrog, &subCamEye);
-        subCamEye.y -= 1.8f;
-        subCamAt.y -= 1.7f;
-        Play_SetCameraAtEye(play, this->subCamId, &subCamAt, &subCamEye);
+        if (GameInteractor_Should(VB_ENEMY_CUTSCENE_ACTION, true, this)) {
```
Archez: Same here (pt. 2)

Archez (commented):
Left a little spam of comments lol, mainly trying to think about how to minimize impact to vanilla src (even if just whitesapcing related)

Eblo:
Applied the 8 early return suggestions. The diff is a little cleaner now.

garrettjoecox (commented):
At a high level, this feels more invasive than I'd like on the source code for z_en_pametfrog, z_en_bigslime, and z_en_kaizoku.

If these actors aren't often in a scene, it might be worth trying to use `ShouldActorUpdate`, and conditionally overriding their actionFunc with a custom one that cuts out the undesired behavior before the update is run.

Eblo:
The actors using `VB_ENEMY_CUTSCENE_ACTION` are indeed uncommon. I think what you're suggesting is feasible for the actor behavior changes, but camera behavior may not be so easy. I initially wanted to hook in at the lower level in the camera functions for exactly the reason of having to touch less of the source code as possible, but actor information is not accessible at that point. I'll see what I can do.

Eblo:
> If these actors aren't often in a scene, it might be worth trying to use `ShouldActorUpdate`, and conditionally overriding their actionFunc with a custom one that cuts out the undesired behavior before the update is run.

I pushed an update that does this just for EnPametfrog. Given how extensive the change in approach is, I wanted to do just one enemy to submit for consideration before doing all of them.

The footprint on `z_en_pametfrog.c` has now been eliminated, and in its place is `SkipGekkoSnapperCutscenes.cpp`. Much of the code there is duplicated actor functions, with references to camera controls removed. Since some of the camera code is nested in functions indirectly called by actionFuncs, this redundancy was unavoidable.

If this example is preferred over `GameInteractor_Should` calls, I can proceed with the other enemies. I think the ones you've named, including this one I've already done, would be the most extensive changes. The others should be simple to convert. Then the new VB can just be removed.

Eblo:
After a discussion, I took Archez's suggestion to instead hook into the more common shared camera functions and use a static variable combined with actor lifecycle hooks to remove the need for ad-hoc modifications to individual actor source code. This flag is set to true when a relevant actor is initialized and false when the actor is destroyed. Considering that these cutscene-driven enemies don't share rooms with each other, that should be fine.

* `VB_PLAYER_CUTSCENE_ACTION` is used to stop actors from controlling/locking Link during cutscenes that are now skipped
* `VB_SET_CAMERA_AT_EYE` is used to prevent actors from controlling the camera for cutscenes that are now skipped
* `VB_SET_CAMERA_FOV` similar to the above, but only for Eyegore
* I also added some static variables to the enhancement file for player rotation. These are only used for the Fighter Pirate actor, which forces Link to face them during the now-skipped death cutscene. I could use ActorExtension if that's preferred, but that feels like a bit much overhead for something so brief and minor.

mm/2s2h/Enhancements/Cutscenes/SkipEnemyCutscenes.cpp
```diff
@@ -0,0 +1,407 @@
...
+
+    COND_VB_SHOULD(VB_SET_CAMERA_FOV, CVAR, {
+        if (skipSetCamera) {
+            *should = false;
+        }
+    });
+
+    // Camera control pair functions. Ensure skipSetCamera is reset to false when actor is destroyed
+    COND_ID_HOOK(OnActorInit, ACTOR_EN_PAMETFROG, CVAR, [](Actor* actor) { skipSetCamera = true; });
+    COND_ID_HOOK(OnActorDestroy, ACTOR_EN_PAMETFROG, CVAR, [](Actor* actor) { skipSetCamera = false; });
```
Archez: For these camera control pairs and the static var, using `OnActorInit`/`OnActorDestroy` means that the camera funcs are basically stopped for the life of the actor, rather than only when the actor is updating.

My original thought was to use `ShouldActorUpdate`/`OnActorUpdate` which is basically a before update and after update hook. This way the static var is only set around the boundaries of the specific actor updating, rather than held through the life of multiple gamestate frames (potentially blocking other non-enemy related camera changes).

To be even more specific, you could check the actionFunc within the `ShouldActorUpdate` to only set during the specific updates that need to be handled.

Archez (commented):
Loving the latest updates! The footprint has been significantly reduced which is nice.
Mainly had one bit of feedback below that I think should be worth considering.

mm/2s2h/Enhancements/Cutscenes/SkipEnemyCutscenes.cpp
Eblo: Applied this change.

> (potentially blocking other non-enemy related camera changes)

This helped me uncover a bug that made it easy to test this change. ISoT uses camera funcs to zoom away from the player. With the init/destroy approach, that would not happen if these affected actors were present in a scene. With the update hook approach, the ISoT camera zooms as intended regardless of the actors' presence.

mm/2s2h/Enhancements/Cutscenes/SkipEnemyCutscenes.cpp
```diff
@@ -0,0 +1,405 @@
...
+        if (enKnight->actionFunc == func_809B6764) {
+            // This is Igos, and the two lackeys have been slain. Skip cutscene, start next phase
+            if (D_809BEFD4->actor.draw == NULL && D_809BEFD8->actor.draw == NULL) {
+                enKnight->unk17A = enKnight->unk178 = enKnight->unk176 = enKnight->unk174 = 0;
+                enKnight->actionFunc = func_809B71DC;
+                enKnight->unk684 = 0;
+                enKnight->unk6A4 = 0.0f;
+                enKnight->unk46C = enKnight->unk470 = 1.0f; // Show sword and shield
+                enKnight->unk688 = 0;
+                func_809B3E9C(enKnight, gPlayState);
```
balloondude2: `func_809B3E9C` sets an actionFunc, so I don't think `enKnight->actionFunc = func_809B71DC;` ends up doing anything or is necessary here.

mm/src/code/z_actor.c
```diff
@@ -1540,6 +1540,10 @@ void Actor_SpawnHorse(PlayState* play, Player* player) {
  * Player must leave the cutscene action state and enter it again before halting actors can be toggled.
  */
 s32 Player_SetCsAction(PlayState* play, Actor* csActor, u8 csAction) {
+    if (!GameInteractor_Should(VB_PLAYER_CUTSCENE_ACTION, true, csActor, &csAction)) {
+        return false;
+    }
```
balloondude2: `&csAction` doesn't seem to be used over in the .cpp file.

The other VBs also seem to have unused va_args.

balloondude2 (approved):
Left a few minor nitpicks. It'd be nice to get some more QA on this.

PR #1042, opened 2025-03-05
[Rando] Fix Dungeon Access Logic

Patrick12115:
Didn't handle the default logic correctly when adding the new dungeon access options. This should fix that.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2700909450.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2700993789.zip)

### DIFF
--- a/mm/2s2h/Rando/Logic/Logic.h
+++ b/mm/2s2h/Rando/Logic/Logic.h
@@ -153,8 +153,9 @@ inline bool CanAccessDungeon(DungeonIndex dungeonIndex) {
         case RO_ACCESS_DUNGEONS_SONG_ONLY:
             return hasSongAccess;
         case RO_ACCESS_DUNGEONS_OPEN:
-        default:
             return true;
+        default:
+            return hasSongAccess && hasFormAccess;
     }
 }
 
--- a/mm/2s2h/Rando/Logic/Logic.h
+++ b/mm/2s2h/Rando/Logic/Logic.h
@@ -154,6 +154,7 @@ inline bool CanAccessDungeon(DungeonIndex dungeonIndex) {
             return hasSongAccess;
         case RO_ACCESS_DUNGEONS_OPEN:
             return true;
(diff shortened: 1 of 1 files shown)


Archez (approved):

mm/2s2h/Rando/Logic/Logic.h
```diff
@@ -153,8 +153,9 @@ inline bool CanAccessDungeon(DungeonIndex dungeonIndex) {
         case RO_ACCESS_DUNGEONS_SONG_ONLY:
             return hasSongAccess;
         case RO_ACCESS_DUNGEONS_OPEN:
-        default:
             return true;
+        default:
+            return hasSongAccess && hasFormAccess;
```
balloondude2, line 159: I see how this default will fix the issue with `RO_ACCESS_DUNGEONS_FORM_AND_SONG`, but is there a reason we don't want an explicit case for `RO_ACCESS_DUNGEONS_FORM_AND_SONG` and are just having it fall to `default`?

Patrick12115: I barely know how any of this works, so I have no preference. Just the first thing I did and it worked.

balloondude2: I'm not sure if it's a big deal (I also don't claim to really know what's going on either :Shrek:) but I would prefer an explicit case for each option. If others are okay as is, that's fine with me though.

Archez: Having an explicit case would be better, if at least just for readability.

Patrick12115: Alright, done. Hopefully the way you guys meant.

Archez: How you did it is good 👍

balloondude2 (approved):

PR #1044, opened 2025-03-06
[Difficulty Options] Adds a timer for the Jinxed effect

Patrick12115:
Adds a slider to set the jinxed effect from anywhere between 0-60. If set to 0 the jinxed effect will not apply at all.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2734656858.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2734663709.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2734685736.zip)

### DIFF
--- a/mm/2s2h/Enhancements/DifficultyOptions/JinxedTimer.cpp
+++ b/mm/2s2h/Enhancements/DifficultyOptions/JinxedTimer.cpp
@@ -18,8 +18,6 @@ void RegisterJinxedTimer() {
                 gSaveContext.jinxTimer = CVAR * 20;
             }
         }
-        // Log the timer value for debugging
-        printf("jinxTimer after modification: %d\n", gSaveContext.jinxTimer);
     });
 }
 
--- a/mm/2s2h/Enhancements/DifficultyOptions/JinxedTimer.cpp
+++ b/mm/2s2h/Enhancements/DifficultyOptions/JinxedTimer.cpp
@@ -1,0 +1,23 @@
+#include <libultraship/bridge.h>
+#include "2s2h/GameInteractor/GameInteractor.h"
+#include "2s2h/ShipInit.hpp"
+
+#define CVAR_NAME "gEnhancements.DifficultyOptions.JinxedTimer"
+#define CVAR CVarGetInteger(CVAR_NAME, 60)
+
+void RegisterJinxedTimer() {
+    COND_VB_SHOULD(VB_MODIFY_JINX_TIMER, CVAR < 60, {
+        s32* timer = va_arg(args, s32*);
+
+        if (CVAR == 0) {
+            *timer = 0;
+        } else {
+            // Prevent the timer from exceeding CVAR * 20 and resetting to 1200 if hit again while jinxed.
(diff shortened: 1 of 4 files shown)


mm/2s2h/Enhancements/DifficultyOptions/JinxedTimer.cpp
```diff
@@ -0,0 +1,23 @@
+#include <libultraship/bridge.h>
+#include "2s2h/GameInteractor/GameInteractor.h"
+#include "2s2h/ShipInit.hpp"
+
+#define CVAR_NAME "gEnhancements.DifficultyOptions.JinxedTimer"
+#define CVAR CVarGetInteger(CVAR_NAME, 60)
+
+void RegisterJinxedTimer() {
+    COND_VB_SHOULD(VB_MODIFY_JINX_TIMER, CVAR < 60, {
+        s32* timer = va_arg(args, s32*);
```
Archez: `gSaveContext.jinxTimer` is a `u16` and you are passing it as a pointer, so this arg type and var type should be `u16*`. The default arg promotion of int types smaller than 32 only applies to standard types, not pointer types.

However, `gSaveContext` can be accessed directly without needing to use the `va_arg` pattern, I would prefer to just set the jinxTimer directly in here instead of passing a pointer to the save.

mm/src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -12138,6 +12138,8 @@ void Player_UpdateCommon(Player* this, PlayState* play, Input* input) {
 
     if (gSaveContext.jinxTimer != 0) {
         gSaveContext.jinxTimer--;
+
+        GameInteractor_Should(VB_MODIFY_JINX_TIMER, true, &gSaveContext.jinxTimer);
```
Archez: Seems to me that we could just use a `OnPlayerUpdate` hook instead of adding a new `VB_` to execute your logic.

Patrick12115:
Applied feedback

mm/2s2h/Enhancements/DifficultyOptions/JinxedTimer.cpp
```diff
@@ -0,0 +1,24 @@
...
+            return;
+        }
+
+        if (CVAR == 0) {
+            gSaveContext.jinxTimer = 0;
+        } else {
+            if (gSaveContext.jinxTimer > CVAR * 20) {
+                gSaveContext.jinxTimer = CVAR * 20;
+            }
+        }
```
garrettjoecox: Unless I'm mistaken, your final condition should work fine with 0, so you can just replace this whole block with
```suggestion
        if (gSaveContext.jinxTimer > CVAR * 20) {
            gSaveContext.jinxTimer = CVAR * 20;
        }
```

Archez (approved):

PR #1048, opened 2025-03-08
 Settings: Basic Speedrun Cutscene Behavior Customization

JoshSanch:
Adds a basic sidebar menu and CVar conventions for configuring cutscene customizations requested by speedrunners.

Namely, this PR allows speedrunners to customize what items are granted by the Skip Intro Sequence/Skip First Cycle Time Saver enhancements.

420559570-2820f114-7bf2-4fa9-abc4-878fc9cb6092

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2715239399.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2715241228.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -423,7 +423,9 @@ void BenMenu::AddSettings() {
     path.column = 1;
     path.sidebarName = "Speedrunning";
     AddSidebarEntry("Settings", "Speedrunning", 1);
-    AddWidget(path, "Speedrunning", WIDGET_CUSTOM).CustomFunction([](WidgetInfo& info) { SpeedrunCustomization_Draw(); });
+    AddWidget(path, "Speedrunning", WIDGET_CUSTOM).CustomFunction([](WidgetInfo& info) {
+        SpeedrunCustomization_Draw();
+    });
 }
 int32_t motionBlurStrength;
 
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -10,6 +10,7 @@
 #include "DeveloperTools/EventLog.h"
 #include "2s2h/Enhancements/GfxPatcher/AuthenticGfxPatches.h"
 #include "2s2h/PresetManager/PresetManager.h"
+#include "2s2h/SpeedrunCustomization/SpeedrunCustomization.h"
 #include "HudEditor.h"
 #include "Notification.h"
 #include <variant>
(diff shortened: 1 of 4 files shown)


Archez:
I'm a bit hesitant on accepting this. I think it would set a precedent on us actively creating changes to specifically curate speedrunning experiences/runs and putting us in a position to act on runners opinions, beyond just the standard options that the port provides. In my opinion, I think it should be the other way around. The speedrunning community should be designing their experience around what the port provides.

For the changes in this PR specifically, I mentioned before that I would prefer to just remove granting Deku Nuts from the Intro Skip, since they are largely unusable and get tossed after first cycle anyways. It's easier to just assume the bare minimum, and then leave it up to speed runners to decide their routes from what is available. Right now you just have nuts and bomb bag, but I could see bank rupees being the next thing people would start to ask about, but all these things depend on how people route those sections, which a skip essentially hand waves away.

An alternate answer to this could just be curated save files the speedrunning community can provide based on the categories, similar to how console has curated saves. This way the port itself doesn't have to keep "adapting" to the speedrunning community.

---

Anyways, sorry for the big wall of text. This is definitely just my opinion currently, and I would like to hear from other maintainers and members their thoughts.

JoshSanch:
Let me offer an alternative perspective and path for this PR, in this case, since this is on me for packaging it as specifically a PR for speedrunners.

This PR in its current state is a targeted enhancement of the project's current Cutscene Skip functionality.
Options like:
- Skip Intro Sequence
- Skip First Cycle
- Skip Story Cutscenes
- Skip Misc Interactions

These Cutscene Skips offer no customization at present for players in general who might, for example, want to skip the Goron Lullaby cutscenes but really like watching Skull Kid taunt the player on the Clock Tower at the beginning of their playthrough.

Do you think that there is a version of this PR that is, instead of an isolated enhancement targeted for speedrunners, a general improvement to the ability to customize the cutscene skip system?

JoshSanch:
@Archez bump on this thread

garrettjoecox:
@JoshSanch In general I agree with Archez' sentiment, but additionally I'm trying to be cautious when it comes to introducing _more_ menus, and _more_ options. While these may appeal to a very small number of people who would understand and utilize them, to the general audience it's just more stuff to overwhelm them, causing them to not look at the menus at all. 

Regarding cutscene skip granularity, I could imagine having a menu that is hidden by default where you can choose each individual cutscene skip, but then you have to weigh the maintenance burden that introduces with every single skip we add going forward, and how many users will actually use that sub-menu over just clicking the checkbox that marks them all (probably 99%)

Going to close this PR for now, but I am open to further discussion on this in the discord

PR #1052, opened 2025-03-09
Rando Check: Termina Field Gossip Stone Grottos Heart Piece

JoshSanch:
Adds check for the Heart Piece granted by the Gossip Stones in the Termina Field grottos which change color when you play a Dungeon Access song for them in the corresponding forms (Deku/Sonata, Zora/New Wave Bossa Nova, Goron/Goron Lullaby).

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2725260368.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2725428136.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2725442526.zip)


mm/2s2h/Rando/ActorBehavior/EnGs.cpp
```diff
@@ -159,4 +159,23 @@ void Rando::ActorBehavior::InitEnGsBehavior() {
...
+        GetItemId* item = va_arg(args, GetItemId*);
+        Actor* refActor = va_arg(args, Actor*);
+        Player* player = GET_PLAYER(gPlayState);
+
+        if (refActor->id != ACTOR_EN_GS || *item != GI_HEART_PIECE) {
+            return;
+        }
+
+        if (!RANDO_SAVE_CHECKS[RC_TERMINA_FIELD_GOSSIP_STONE_GROTTO].cycleObtained) {
+            RANDO_SAVE_CHECKS[RC_TERMINA_FIELD_GOSSIP_STONE_GROTTO].eligible = true;
```
balloondude2: In Checks.cpp, you list `WEEKEVENTREG_90_10` as the check's flag. That flag getting set should add the Rando item to the queue, so I'm not sure if manually setting `eligible` here is needed.

JoshSanch: Wasn't aware that was how the checks with weekevent flags worked - so basically checks that use weekevent flags manage their own check grants? If so, I'll remove the manual grant.

Eblo: `FLAG_WEEK_EVENT_REG` runs `OnFlagSet` hooks when `SET_WEEKEVENTREG` gets called. Rando has the following OnFlagSet hook:
https://github.com/HarbourMasters/2ship2harkinian/blob/00c491d328f33a76ed897e75cf84ab6b1e75fb24/mm/2s2h/Rando/MiscBehavior/OnFlagSet.cpp#L7-L17

Ergo, yes, whenever a rando check has a flag configured, it will automatically queue the RI when that flag gets set.

balloondude2: Yeah, most if not all of the checks with flags (not just weekEventReg) are handled with a generic OnFlagSet/OnSceneFlagSet hook.

JoshSanch: Updated.

balloondude2 (approved):
Thanks for making that update. LGTM and works from my testing. 

My only concern is that the logic won't work as written for future entrance randomizer. Leaving my approval as I'm fine if that's taken care of when entrance rando is realized. But one option would be to use events, where each gossip stone grotto sets a unique event (similar to the frog choir check).

PR #1060, opened 2025-03-13
Rando: Medigoron Behavioral Fixes

### LINKED ISSUE #1014: Weirdness surrounding the Powder Keg test and associated Check.
When this check was implemented, it was primarily tested with the "Skip Powder Keg" enhancement enabled, causing us to implement it in a way that doesn't work well without it on. For instance one quirk is that you receive the check when blowing up the boulder in twin isles.

There are also various scenarios that just aren't handled correctly, if you have more information about this issue please comment below.

JoshSanch:
Fixes #1014 

Updates the Rando behavior of Medigoron to separate the ability to shop from the certification test and subsequent reward.
Also grants the player the check reward when talking to Medigoron for the first time each cycle after they've:

- Completed the certification test OR
- Acquired the Powder Keg

Also includes logic fixes to account for the ability to unfreeze the ice outside of his shop by clearing Snowhead Temple.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2742879582.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2742882700.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2742900783.zip)


garrettjoecox (commented):
Needs some thorough QAing! nice work thanks for tackling this beast

Eblo:
As far as I can tell from my testing, this seems to work as described. Automatically grants check if the enhancement is enabled or the player already has the powder keg inventory slot unlocked. Otherwise, the player must complete the test.

Eblo:
Extra comment for documentation's sake: the Powder Keg Blast Mask cheat has some minor weirdness but doesn't ultimately block anything. If the player has not unlocked the powder keg, blows the raceway rock using the superpowered Blast Mask, and talks to the Medigoron, he acts as if the player hasn't done the test. However, the player can just take the test from there, let that powder keg explode without needing to actually take it anywhere, and then the Medigoron recognizes that the player passed the test. Since it's still doable, and it comes from a cheat anyway, I think that's fine.

PR #1065, opened 2025-03-15
[Rando] Implement Treasure Chest Game Checks

balloondude2:
This adds the Treasure Chest Game checks, one for each main form. 

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2757023506.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2757029013.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2757067515.zip)


mm/2s2h/Rando/MiscBehavior/OnCycleSave.cpp
```diff
@@ -70,6 +70,24 @@ void Rando::MiscBehavior::AfterEndOfCycleSave() {
                 WEEKEVENTREG((randoStaticCheck.flag) >> 8) =
                     GET_WEEKEVENTREG((randoStaticCheck.flag) >> 8) & (u8) ~((randoStaticCheck.flag) & 0xFF);
                 break;
+            case FLAG_CYCL_SCENE_SWITCH:
+                // Clear the flag without triggering hook
```
Eblo, line 74: Was this really necessary? Vanilla already ensures the first-time rewards can only be obtained once, so those relevant flags must persist between cycles. Or is this doing something else?

Edit: Oops, this is for *un*-setting flags without triggering hooks. In which case, why was this needed now when other chests didn't call for this?

Eblo (commented):
Works from my testing, but left one comment asking for clarity.

I quite like that these chests' GI animations play out like normal instead of being queued like with other checks. Looks to be the intended behavior in that that each non-FD form yields the shuffled check the first time, while repeats yield the standard rupee rewards.

mm/2s2h/Rando/MiscBehavior/OnCycleSave.cpp
balloondude2: So these checks don't look at chest flags, but rather switch flags (which is why it's `case FLAG_CYCL_SCENE_SWTICH:`). The chest flags don't need them because there is only one chest flag that is persistent in the game, which happens to be for the Laundry Pool scene where there isn't actually a chest. (Edit: It's possible this flag is used for something, but we don't use it in Checks.cpp atm)

This was added solely to allow the Goron form check to be repeatable. Otherwise, the switch flag that determines if the game will give the initial (heart piece) or subsequent reward will persist across cycles. 

One potential side effect will be any other, if any, checks that use persistent switch flags now being repeatable as well.

Eblo: I'm wary of having a sweeping change like this to account for a single check. There could be unknown side effects elsewhere. Is it a big deal if the game just continues to give the subsequent reward cycle after cycle? With this change, the first repeat reward in a subsequent cycle is just going to be junk, while without, it would be the next reward of rupees (effectively junk). The player is probably unlikely to repeat this minigame for a given form in subsequent cycles as it is.

Eblo (approved):
Upon further discussion, the scene flag reset change probably won't hurt anything else since it only affects this and the Stray Fairy checks at the moment. I'm fine with this as is.

PR #1069, opened 2025-03-17
[Rando] Excluding Checks from Seed

Caladius:
Adds functionality to exclude Checks from seed generation.

- [ ] Logic Fixes for Glitchless (non No Logic).

image

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4110452547.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4110462804.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4110608862.zip)


mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp
```diff
@@ -114,6 +117,27 @@ void Rando::MiscBehavior::OnFileCreate(s16 fileNum) {
...
+
+                            if (excludedChecksList != "") {
+                                std::string word;
+                                std::istringstream stream(excludedChecksList);
+                                while (std::getline(stream, word, ',')) {
+                                    excludedChecks.push_back((RandoCheckId)std::stoi(word));
+                                }
+                            }
+                            initExcludedChecks = true;
+                        }
```
balloondude2: I think this can be moved outside of both for loops, and then you won't need `initExcludedChecks`.

mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp
```diff
@@ -114,6 +117,27 @@ void Rando::MiscBehavior::OnFileCreate(s16 fileNum) {
...
+
+                        auto it = std::find(excludedChecks.begin(), excludedChecks.end(), randoCheckId);
+                        if (it != excludedChecks.end()) {
+                            RANDO_SAVE_CHECKS[randoCheckId].randoItemId = RI_JUNK;
+                            RANDO_SAVE_CHECKS[randoCheckId].skipped = true;
+                            continue;
+                        }
```
balloondude2: This is causing the check's item to not get added to the itemPool, so it won't be placed anywhere. This normally happens at the end of the loop with `itemPool.push_back(randoStaticCheck.randoItemId);`. You could add another `itemPool.push_back(randoStaticCheck.randoItemId);` here to make sure the item is included, but I'm not sure if that's the best solution.

balloondude2 (commented):
Didn't look at the menu code, but I liked the UX.

mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp
```diff
@@ -203,6 +212,17 @@ void Rando::MiscBehavior::OnFileCreate(s16 fileNum) {
                             }
                         }
 
+                        // Skip checks that have been excluded in the Locations menu and add their vanilla item to the
+                        // pool
+                        auto it = std::find(excludedChecks.begin(), excludedChecks.end(), randoCheckId);
+                        if (it != excludedChecks.end()) {
+                            itemPool.push_back(Rando::StaticData::Checks[randoCheckId].randoItemId);
+                            randoStaticCheck.randoItemId = RI_JUNK;
```
balloondude2: You currently aren't able to generate two seeds in a row without closing the application in between (which I think is related to the boot to warp issues). I think it's because this line is changing the static data, which essentially messes up the data for subsequent generations.

Caladius: Fixed this and also made a change to correct what is displayed in the Rando tab of the save editor. Should be good now, Glitchless generates successfully multiple times per run and also tested No Logic.

mm/2s2h/Rando/Logic/GlitchlessLogic.cpp
```diff
@@ -96,6 +96,10 @@ void ApplyGlitchlessLogicToSaveContext(std::unordered_map<RandoCheckId, bool>& c
 
                     RandoItemId randoItemId;
 
+                    if (randoCheckId == RC_ANCIENT_CASTLE_OF_IKANA_BOSS) {
+                        int hi = 0;
+                    }
```
Eblo: Forget to remove some test code?

mm/2s2h/Rando/Menu.cpp
```diff
@@ -402,6 +430,112 @@ static void DrawItemsTab() {
     ImGui::EndChild();
 }
 
+static void DrawLocationsTab() {
+    auto menuThemeColor = UIWidgets::Colors(CVarGetInteger("gSettings.Menu.Theme", LightBlue));
```
Eblo, line 440:
```suggestion
    auto menuThemeColor = UIWidgets::Colors(CVarGetInteger("gSettings.Menu.Theme", defaultThemeIndex));
```

mm/2s2h/Rando/Menu.cpp
```diff
@@ -402,6 +430,112 @@ static void DrawItemsTab() {
...
+    }
+    ImGui::EndChild();
+    ImGui::SameLine();
+    ImGui::BeginChild("randoExcludedChecks", ImVec2(columnWidth, ImGui::GetContentRegionAvail().y));
+    ImGui::SeparatorText("Forced Junk Checks");
+
+    static ImGuiTextFilter excludedFilter;
+    UIWidgets::PushStyleCombobox(menuThemeColor);
+    excludedFilter.Draw("##filter", ImGui::GetContentRegionAvail().x - ImGui::CalcTextSize("Clear All").x - 30.0f);
+    // UIWidgets::PopStyleCombobox();
```
Eblo:
```suggestion
```

mm/2s2h/Rando/Logic/GlitchlessLogic.cpp
```diff
@@ -108,7 +112,11 @@ void ApplyGlitchlessLogicToSaveContext(std::unordered_map<RandoCheckId, bool>& c
                         SPDLOG_TRACE("Check: {}:{}", Rando::StaticData::Checks[randoCheckId].name,
                                      Rando::StaticData::Items[randoItemId].spoilerName);
                     } else {
-                        randoItemId = Rando::StaticData::Checks[randoCheckId].randoItemId;
+                        if (RANDO_SAVE_CHECKS[randoCheckId].skipped) {
+                            randoItemId = RI_JUNK;
+                        } else {
+                            randoItemId = Rando::StaticData::Checks[randoCheckId].randoItemId;
+                        }
```
Eblo: Is this block necessary for glitchless logic? In OnFileCreate.cpp you already do:
```c++
RANDO_SAVE_CHECKS[randoCheckId].randoItemId = RI_JUNK;
```
That should affect any logic setting, which in effect means that the original assignment here should be `RI_JUNK` if the check was excluded. If this is indeed redundant, the entire `skipped` flag may be as well.

Edit: Or, given my other comment about the check tracker, maybe the `skipped` flag could mark a check as ignored in the tracker?

mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp
```diff
@@ -203,6 +212,20 @@ void Rando::MiscBehavior::OnFileCreate(s16 fileNum) {
...
+                        // Skip checks that have been excluded in the Locations menu and add their vanilla item to the
+                        // pool
+                        auto it = std::find(excludedChecks.begin(), excludedChecks.end(), randoCheckId);
+                        if (it != excludedChecks.end()) {
+                            RandoItemId vanillaItem = Rando::StaticData::Checks[randoCheckId].randoItemId;
+                            itemPool.push_back(vanillaItem);
+
+                            RANDO_SAVE_CHECKS[randoCheckId].randoItemId = RI_JUNK;
+                            RANDO_SAVE_CHECKS[randoCheckId].skipped = true;
+                            RANDO_SAVE_CHECKS[randoCheckId].shuffled = false;
```
Eblo: Setting `shuffled` to false may be causing some unintended behavior. If I exclude a freestanding Heart Piece check, the Clock Tower platform for example, it will end up being a non-RI Heart Piece instead of junk. I think that might mean a redundant Heart Piece is in the world, and this might also be impacting other ways that rando uses the `shuffled` flag.

Is there any harm in allowing the check to remain "shuffled" even if it's known junk? Was this done to make the check tracker automatically ignore skipped checks?

Eblo (commented):
Left some feedback. I second the UX; it's straight and to the point.

mm/2s2h/Rando/Menu.cpp
Caladius: defaultThemeIndex Doesnt exist here currently.

mm/2s2h/Rando/Logic/GlitchlessLogic.cpp
Caladius: I could not get this to operate as expected without adding that chunk there. I welcome any alternative code suggestions, this was annoying to figure out.

mm/2s2h/Rando/Menu.cpp
Malkierian: This is why I made the THEME_COLOR macro on SoH. It lives in `SohGui`.

mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp
Caladius: Remaining Shuffled didnt seem to have any harm (that I could tell), you are correct this was done to automate the Check Tracker ignoring it.

mm/2s2h/Rando/Logic/GlitchlessLogic.cpp
Caladius: Sure did! lmao

mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp
Eblo: Could you try leaving `shuffled` alone, and instead updating the tracker to take `skipped` into account when ignoring checks? As is, setting `shuffled` to false does cause strange behavior.

mm/2s2h/Rando/Logic/GlitchlessLogic.cpp
```diff
@@ -108,7 +108,11 @@ void ApplyGlitchlessLogicToSaveContext(std::unordered_map<RandoCheckId, bool>& c
...
```
Eblo:
```suggestion
                        if (!RANDO_SAVE_CHECKS[randoCheckId].skipped) {
                            randoItemId = Rando::StaticData::Checks[randoCheckId].randoItemId;
                        }
```

mm/2s2h/Rando/Logic/GlitchlessLogic.cpp
```diff
@@ -96,7 +96,13 @@ void ApplyGlitchlessLogicToSaveContext(std::unordered_map<RandoCheckId, bool>& c
 
                     RandoItemId randoItemId;
 
-                    if (isShuffled) {
+                    if (RANDO_SAVE_CHECKS[randoCheckId].skipped) {
+                        randoItemId = RI_JUNK;
+                        auto findIt = std::find(itemPool.begin(), itemPool.end(), RI_JUNK);
+                        if (findIt != itemPool.end()) {
+                            itemPool.erase(findIt);
+                        }
```
balloondude2:
```suggestion
                        for (auto it = itemPool.begin(); it != itemPool.end(); ++it) {
                            if (it != itemPool.end()) {
                                if (Rando::StaticData::Items[*it].randoItemType == RITYPE_JUNK) {
                                    randoItemId = *it;
                                    itemPool.erase(it);
                                    break;
                                }
                            }
                        }
```

Because this is looking specifically for RI_JUNK, seeds are likely to fail to generate without many shuffle options selected. The base check pool only has one instance of RI_JUNK for example. This suggestion will allow any item with RITYPE_JUNK to be selected, allowing for more checks to be excluded.

Eblo (approved):

balloondude2 (approved):

PR #1072, opened 2025-03-19
Cleanup includes

louist103:
This removes all instances of `#include "libultraship/libultraship.h"` and `#include "libultraship/bridge.h"`
These headers are both very large and contain most of LUS' headers. Editing any header in LUS would cause most of 2ship to rebuild. 
On top of making rebuilds faster when modifying LUS, this decreases the build time from just over 2 minutes, to 1:50 on a 25 thread build. This will probably translate to several minutes on actions builds assuming no cache.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2804301447.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2804308111.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2804331752.zip)


mm/src/code/z_parameter.c
```diff
@@ -21,7 +21,9 @@
 #include "2s2h/BenGui/HudEditor.h"
 #include "2s2h/BenGui/CosmeticEditor.h"
 #include "2s2h_assets.h"
+#include "gfxbridge.h"
```
Archez: Use full path like the other bridges
```suggestion
#include "public/bridge/gfxbridge.h"
```

mm/src/code/z_camera.c
```diff
@@ -93,6 +94,11 @@ Vec3f D_801EDDF0;
 #define CAM_CHANGE_SETTING_2 (1 << 2)
 #define CAM_CHANGE_SETTING_3 (1 << 3)
 
+#define CHECK_BTN_DPAD(input)                                                                                   \
+    (CVarGetInteger("gEnhancements.Dpad.DpadEquips", 0) &&                                                      \
+     (CHECK_BTN_ALL(input, BTN_DRIGHT) || CHECK_BTN_ALL(input, BTN_DLEFT) || CHECK_BTN_ALL(input, BTN_DDOWN) || \
+      CHECK_BTN_ALL(input, BTN_DUP)))
+
```
Archez: This should return back to where it was in the header file, so that its side-by-side with the other DPad macros and the original non-dpad macros.

It having a CVar check in the macro definition does not require us to have the LUS header in the decomp header, it will only be required in the files that use the macro (like this file).

mm/src/overlays/gamestates/ovl_select/z_select.c
```diff
@@ -7,10 +7,8 @@
 #include "z_select.h"
 #include "z64shrink_window.h"
 #include "z64view.h"
-#include "libc/alloca.h"
```
Archez, line 10: I would prefer to not remove any decomp includes (regardless if we are using them or not)

mm/src/code/graph.c
```diff
@@ -3,7 +3,10 @@
 #include "regs.h"
 #include "functions.h"
 #include "fault.h"
-#include <libultraship/bridge.h>
+#include "public/bridge/gfxdebuggerbridge.h"
+#include "public/bridge/consolevariablebridge.h"
+#include "public/bridge/windowbridge.h"
+#include "gfx.h"
```
Archez: Did you mean to add `gfx.h` here? Wouldn't this already be covered by the vanilla include for `z64.h`?

mm/include/functions.h
```diff
@@ -7,9 +7,6 @@ extern "C" {
 #endif
 
 #include "z64.h"
-#include "libultraship/luslog.h"
-
-#include <public/bridge/consolevariablebridge.h>
```
Archez, line 12: RIP @garrettjoecox, but yeah I think compilation time/units improvements outweighs the convienience of having these be in a "global" header (originally added to have `CVar*` and `LUSLOG*` readily available)

Perhaps we can think of a conditional compile pattern to make this nicer for those that want that (at the cost of longer compile times), without impacting CI/CD and default clone experiences.

mm/2s2h/Rando/Logic/Logic.h
```diff
@@ -1,13 +1,13 @@
...
 #define RANDO_LOGIC_H
 
-#include <libultraship/bridge.h>
 #include "Rando/Rando.h"
 #include "2s2h/GameInteractor/GameInteractor.h"
 #include "2s2h/ShipUtils.h"
 
 #include <unordered_map>
 #include <set>
+#include <cassert>
```
Archez, line 10: Does it make sense for this to be included in the `.h` or should it be in the `.cpp` instead?

mm/2s2h/Enhancements/Minigames/SkipBalladOfWindfish.cpp
```diff
@@ -1,10 +1,11 @@
-#include <libultraship/bridge.h>
+#include "public/bridge/consolevariablebridge.h"
 #include "2s2h/GameInteractor/GameInteractor.h"
-#include "2s2h/Enhancements/Enhancements.h"
 #include "2s2h/ShipInit.hpp"
 
 extern "C" {
 #include "functions.h"
+#include "variables.h"
+         "
```
Archez: go home `"`, your drunk
```suggestion
```

mm/2s2h/Enhancements/GfxPatcher/AuthenticGfxPatches.cpp
```diff
@@ -1,7 +1,6 @@
 #include "AuthenticGfxPatches.h"
-#include "libultraship/libultraship.h"
-#include <cstring>
-
+#include "public/bridge/consolevariablebridge.h"
+#include "gfx.h"
 extern "C" {
```
Archez: Lets move the vanilla source headers into the extern block following our other patterns
```suggestion

extern "C" {
#include "gfx.h"
```

mm/2s2h/Enhancements/Cutscenes/SkipEntranceCutscenes.cpp
```diff
@@ -1,6 +1,7 @@
-#include <libultraship/libultraship.h>
+#include "public/bridge/consolevariablebridge.h"
 #include "2s2h/GameInteractor/GameInteractor.h"
 #include "2s2h/ShipInit.hpp"
+#include "variables.h"
```
Archez: Here and a couple of these other enhancement files where you added `variables.h`, could you place them inside `extern "C" { }` blocks if they aren't already, to match our existing patterns
```suggestion
#include "2s2h/ShipInit.hpp"

extern "C" {
#include "variables.h"
}
```

mm/2s2h/Rando/Logic/Logic.h
louist103: Since the assert is used in a function I think it should stay here.

Archez: Ah, I missed that. I'm only used to us having functions in headers in `.hpp` files, so that confused me a little.

louist103: Can't say I'm a fan but its out of scope to change it here and I'd rather not touch rando until after a release.

.github/workflows/macports-deps.txt
```diff
@@ -1 +1 @@
-libsdl2 +universal libsdl2_net +universal libpng +universal glew +universal libzip +universal nlohmann-json +universal tinyxml2 +universal libogg +universal libopus +universal opusfile +universal libvorbis +universal
+libsdl2 +universal libsdl2_net +universal libpng +universal glew +universal libzip +universal nlohmann-json +universal tinyxml2 +universal libogg +universal libopus +universal opusfile +universal libvorbis +universal libfmt10 +universal
```
Archez: Since you fixed the spdlog/fmt header include, can you remove this lib now as fmt is pulled through spdlog

```suggestion
libsdl2 +universal libsdl2_net +universal libpng +universal glew +universal libzip +universal nlohmann-json +universal tinyxml2 +universal libogg +universal libopus +universal opusfile +universal libvorbis +universal
```

PR #1062, opened 2025-03-14
Texture Scrolling Interpolation

NEstelami:
**Note: This should not be merged until #1061 is.**

This implements interpolation for various scrolling textures. Things like water, Zora's magic barrier, waterfalls, etc.

https://github.com/user-attachments/assets/61ba7e83-612f-4df5-9e10-fa4983fa9861

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2803774206.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2803783293.zip)

PR #1076, opened 2025-03-20
Improved Matrix Interpolation

NEstelami:
**Note: This should not be merged until https://github.com/HarbourMasters/2ship2harkinian/pull/1062 is.**

This PR addresses artifacts present with matrix interpolation by implementing a system of matrix deconstruction and reconstruction. This prevents scaling issues as well as instances where characters "pancake".

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2804031452.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2804039892.zip)


Eblo:
The pancaking fix is very nice indeed. I have however caught some issues. At a first glance I noticed that this seems to break closed chests, actor name tags, and some HUD elements when interpolation is on.

Flashing visuals warning:

https://github.com/user-attachments/assets/2833c893-f7f9-4071-8190-99ab378d20bd

This video was recorded in 60 FPS, so the 144 FPS issues look more obvious here than on a 144 Hz display. The HUD flashing is steady and easy to overlook.

FWIW I did not notice these issues in the related PR builds for horse and scrolling texture interpolation.

PR #1077, opened 2025-03-20
[Enhancement] Add N64 Weird Frames restoration

Sirius902:
* Adds an enhancement to restore N64 Weird Frames which fixes weirdshots.
* ~~Fixes an `opus` include to match others in the project (I personally couldn't compile on NixOS without this change).~~ Fixed by a separate PR.

I do have two questions though:
* ~~There is an extra weirdshot frame (and maybe more) you can get by releasing target too late that would crash N64 but doesn't crash 2ship and instead of going underground you go up into the air. Not sure if people playing rando would care but if there's 2ship speedruns I'm not sure if this would be allowed. I could have multiple options where one would only allow you to get weirdshot frames that don't crash N64. Do people care enough to have that or is it fine as is?~~ This frame triggers the console crash handler with the "fix crashes" setting disabled so you can just disable that if you don't want N64 unobtainable frames.
* ~~Is there any sort of "rando profile" like in ship that I could have this auto-enabled for? I know in ship rando my N64 Weirdshots restoration is enabled by default so I wonder if I could do that somehow here or if that would make sense to do.~~ Added to the "curated" preset.

EDIT: Added a debug log for weird frames.

EDIT: ~~Drafting until I can figure out how to dump the weirdshot frames from ROM in the extraction step.~~ I figured out how to build it dynamically without that! :D

EDIT: ~~Drafted until I bring over some cleanup changes from https://github.com/HarbourMasters/Shipwright/pull/5195.~~

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2973010817.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2973020254.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2973026782.zip)

### DIFF
--- a/mm/src/code/z_skelanime.c
+++ b/mm/src/code/z_skelanime.c
@@ -3,6 +3,8 @@
 #include <string.h>
 #include <stdio.h>
 
+#include "2s2h/GameInteractor/GameInteractor.h"
+
 #define ANIM_INTERP 1
 
 s32 PlayerAnimation_Loop(PlayState* play, SkelAnime* skelAnime);
@@ -1017,7 +1019,8 @@ void AnimationContext_SetLoadFrame(PlayState* play, PlayerAnimationHeader* anima
                                    Vec3s* frameTable) {
     AnimationEntry* entry = AnimationContext_AddEntry(&play->animationCtx, ANIMATION_LINKANIMETION);
 
-    if (entry != NULL) {
+    if (GameInteractor_Should(VB_LOAD_PLAYER_ANIMATION_FRAME, entry != NULL, entry, animation, frame, limbCount,
+                              frameTable)) {
         if (ResourceMgr_OTRSigCheck(animation) != 0)
             animation = ResourceMgr_LoadAnimByName(animation);
 
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1244,6 +1244,9 @@ void BenMenu::AddEnhancements() {
         .Options(CheckboxOptions().Tooltip("Restores the appearance of Woodfall mountain to not look poisoned "
                                            "when viewed from Termina Field after clearing Woodfall Temple\n\n"
                                            "Requires a scene reload to take effect."));
+    AddWidget(path, "N64 Weird Frames", WIDGET_CVAR_CHECKBOX)
+        .CVar("gEnhancements.Restorations.N64WeirdFrames")
+        .Options(CheckboxOptions().Tooltip("Restores N64 Weird Frames allowing weirdshots to behave the same as N64."));
     AddWidget(path, "Simulated Input Lag", WIDGET_CVAR_SLIDER_INT)
         .CVar(CVAR_SIMULATED_INPUT_LAG)
         .Options(IntSliderOptions()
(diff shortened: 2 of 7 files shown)


mm/src/code/z_skelanime.c
```diff
@@ -1018,28 +1020,31 @@ void AnimationContext_SetLoadFrame(PlayState* play, PlayerAnimationHeader* anima
...
-        //         (((uintptr_t)linkAnimHeader->segmentVoid - 0x07000000)));
-        //
-        // printf("Streaming %s, seg = %08X\n", animPath, linkAnimHeader->segment);
-
-        s16* animData = /* ResourceMgr_LoadPlayerAnimByName*/ (animation->segmentVoid);
-        // 2S2H [Port] sometimes a HESS can set a negative frame value from a negative playback speed. When converted to
-        // a signed value this will cause a crash due to copying way much data.
-        if (frame < 0) {
-            frame = 0;
+        if (GameInteractor_Should(VB_LOAD_PLAYER_ANIMATION_FRAME, true, animation, frame, limbCount, frameTable)) {
```
garrettjoecox: This whole block is already surrounded in a condition check, you can just use that instead of adding another layer of condition. Just make sure you check the `entry != NULL` in your VB handler.
```suggestion
    if (GameInteractor_Should(VB_LOAD_PLAYER_ANIMATION_FRAME, entry != NULL, animation, frame, limbCount, frameTable)) {
```

garrettjoecox (commented):
This change makes 2Ship behave more like hardware, so this doesn't need a menu item, or even a CVar IMO unless someone would ever want the 2ship specific behavior.

Aside from that I'd like to have someone else QA and verify it's working

Sirius902:
> This change makes 2Ship behave more like hardware, so this doesn't need a menu item, or even a CVar IMO unless someone would ever want the 2ship specific behavior.
> 
> Aside from that I'd like to have someone else QA and verify it's working



> This change makes 2Ship behave more like hardware, so this doesn't need a menu item, or even a CVar IMO unless someone would ever want the 2ship specific behavior.
> 
> Aside from that I'd like to have someone else QA and verify it's working

I think it'd still be nice to have a CVar for it though having it enabled by default makes sense to me. Is it possible to make it a CVar that's enabled by default on a fresh install and then just not put options for it in the UI? In that case if someone really wanted to turn it off they could just change the settings manually.

JoshSanch:
Does this just need QA or is there something else blocking it?

Pepper0ni:
~~Maintainers are on break after the 9.0 release push and subsequent bug fixes, therefore nothing will be merged until they get back.~~

Wait wrong repo, I need a break myself.

garrettjoecox:
> Does this just need QA or is there something else blocking it?

I'd like at least one person (other than author) to verify it fixes weirdshots yes

rubbertoe64:
tested weirdshots and yeah theyre consistent! the 2nd frame one seems a bit different but im also a Wii VC runner so idk if that ones just different on N64 lmao, i think its still good tho. Also the setting should probs be set to "on" by default

garrettjoecox:
- Rebased
- Removed menu option
- Defaulted CVAR to on
- Removed from preset (since it's now defaulted to on)

Merging when build is complete

PR #1078, opened 2025-03-20
Unify FPS and V-Sync controls and behavior between renderer

Spodi:
https://github.com/HarbourMasters/Shipwright/pull/5132 ported to 2ship

- DirectX's "Match Refresh Rate" is now a checkbox like in OpenGL and Metal.
- Clamp max FPS to refresh rate when V-Sync is enabled or can't be disabled, regardless of the FPS sliders setting, to avoid slowdown in Metal / OpenGL and dropping frames in DirectX as much as possible.
- FPS slider allows up to 360 FPS in every renderer.
- More descriptive V-Sync tooltip

Also fixes GetInterpolationFPS() not being used to calculate desired FPS.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2805324766.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2805327400.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2805353058.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -224,18 +224,10 @@ void BenMenu::AddSettings() {
     AddWidget(path, "Audio API", WIDGET_AUDIO_BACKEND);
 
     // Graphics Settings
-    static int32_t maxFps;
-    const char* tooltip = "";
-    if (Ship::Context::GetInstance()->GetWindow()->GetWindowBackend() == Ship::WindowBackend::FAST3D_DXGI_DX11) {
-        maxFps = 360;
-        tooltip = "Uses Matrix Interpolation to create extra frames, resulting in smoother graphics. This is "
-                  "purely visual and does not impact game logic, execution of glitches etc.\n\nA higher target "
-                  "FPS than your monitor's refresh rate will waste resources, and might give a worse result.";
-    } else {
-        maxFps = Ship::Context::GetInstance()->GetWindow()->GetCurrentRefreshRate();
-        tooltip = "Uses Matrix Interpolation to create extra frames, resulting in smoother graphics. This is "
-                  "purely visual and does not impact game logic, execution of glitches etc.";
-    }
+    static int32_t maxFps = 360;
+    const char* tooltip = "Uses Matrix Interpolation to create extra frames, resulting in smoother graphics. This is "
+                          "purely visual and does not impact game logic, execution of glitches etc.\n\nA higher target "
+                          "FPS than your monitor's refresh rate will waste resources, and might give a worse result.";
     path.sidebarName = "Graphics";
     AddSidebarEntry("Settings", "Graphics", 3);
     AddWidget(path, "Toggle Fullscreen", WIDGET_CVAR_CHECKBOX)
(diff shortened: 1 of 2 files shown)


Spodi:
I just noticed, that we need at least this version of LUS for everything to make sense: https://github.com/Kenix3/libultraship/commit/7fb120b3c1e8caac36328122b9b6e4b172f1509d

mm/2s2h/BenGui/BenMenu.cpp
```diff
@@ -224,18 +224,10 @@ void BenMenu::AddSettings() {
...
-        tooltip = "Uses Matrix Interpolation to create extra frames, resulting in smoother graphics. This is "
-                  "purely visual and does not impact game logic, execution of glitches etc.\n\nA higher target "
-                  "FPS than your monitor's refresh rate will waste resources, and might give a worse result.";
-    } else {
-        maxFps = Ship::Context::GetInstance()->GetWindow()->GetCurrentRefreshRate();
-        tooltip = "Uses Matrix Interpolation to create extra frames, resulting in smoother graphics. This is "
-                  "purely visual and does not impact game logic, execution of glitches etc.";
-    }
+    static int32_t maxFps = 360;
+    const char* tooltip = "Uses Matrix Interpolation to create extra frames, resulting in smoother graphics. This is "
```
louist103:
```suggestion
    static const char* tooltip = "Uses Matrix Interpolation to create extra frames, resulting in smoother graphics. This is "
```
Otherwise its good.

louist103:
After playing with this I have a suggestion.
Enabling VSync or match monitor refresh rate should both lock the slider and set it to whatever its been locked to. Currently if VSync is enabled the slider is still active but does nothing, and enabling match refresh rate locks it to whatever it was set to, not what it currently is.

Archez:
> Currently if VSync is enabled the slider is still active but does nothing.

This is not the case, vsync = on has been the default behavior of SDL in LUS since forever basically, and slider has definitely been usable and does something.

The new behavior now is that you can actually turn vsync off for SDL. There was a bug in LUS until just a couple of hours ago where toggling vsync was not working (this is now fixed in LUS but not merged to 2ship yet).

So yeah vsync should not disable the fps slider. All that is happening is the upper bound is clamped to monitor hz, but you can still select 30fps etc with vsync on.

Spodi:
> After playing with this I have a suggestion. Enabling VSync or match monitor refresh rate should both lock the slider and set it to whatever its been locked to. Currently if VSync is enabled the slider is still active but does nothing, and enabling match refresh rate locks it to whatever it was set to, not what it currently is.

If you tick "match refresh rate" the slider gets locked wherever it was, but it actually applies the right limit (check with stats window). And for V-Sync enabled it still does whatever you set, as long it is *below* your refresh rate. I did it like this, so those options never change the value you have set, so that going back "restores" your FPS setting.

mm/2s2h/BenGui/BenMenu.cpp
Spodi: Done

Archez (approved):

PR #1082, opened 2025-03-21
Add O2

louist103:
The default release settings don't set `/O2` on Windows. Instead `/Ob` is set. 

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2801890351.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2801891264.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2801919975.zip)

### DIFF
--- a/mm/CMakeLists.txt
+++ b/mm/CMakeLists.txt
@@ -47,38 +47,6 @@ set(CMAKE_CONFIGURATION_TYPES
 )
 
 ################################################################################
-# Global compiler options
-################################################################################
-if(MSVC)
-    # remove default flags provided with CMake for MSVC
-    set(CMAKE_C_FLAGS "")
-    set(CMAKE_C_FLAGS_DEBUG "")
-    set(CMAKE_C_FLAGS_RELEASE "")
-    set(CMAKE_CXX_FLAGS "")
-    set(CMAKE_CXX_FLAGS_DEBUG "")
-    set(CMAKE_CXX_FLAGS_RELEASE "")
-endif()
-
-################################################################################
-# Global linker options
-################################################################################
-if(MSVC)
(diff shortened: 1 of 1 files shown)


Archez:
Are the Linux and Mac builds already `O2` are should we be setting those as well? Do we know if this would cause any issues?

louist103:
> Are the Linux and Mac builds already `O2` are should we be setting those as well? Do we know if this would cause any issues?

Linux seems to be using `O2` as is. I can't speak for mac.
I played a little bit on Windows and it seems fine but I didn't do a whole run or anything.
image

Archez:
ok, I tried a test build on my fork, it looks like CMake sets this for all platforms:
```
CMAKE_C_FLAGS_RELEASE: -O2 -DNDEBUG
CMAKE_CXX_FLAGS_RELEASE: -O2 -DNDEBUG
```
So yeah Linux and Mac have always been `02`

But for some reason on MSVC we seem to be doing this:
```
if(MSVC)
    # remove default flags provided with CMake for MSVC
    set(CMAKE_C_FLAGS "")
    set(CMAKE_C_FLAGS_DEBUG "")
    set(CMAKE_C_FLAGS_RELEASE "")
    set(CMAKE_CXX_FLAGS "")
    set(CMAKE_CXX_FLAGS_DEBUG "")
    set(CMAKE_CXX_FLAGS_RELEASE "")
endif()
```

Which would explain why windows did not have `02`.

Perhaps we should just remove that block? If there is no other reason to keep it.

Archez (approved):

PR #1084, opened 2025-03-22
Fix sword beam collision overcorrection, convert to actor extension

Eblo:
#993 over-corrected the sword beams to *only* collide with enemies, but some non-enemy actors in the game do normally collide with sword beams (signs, Skull Kid, pirate guards to name a few). This fix pulls back the sweeping damage flag changes and instead performs the collision if the actor is an enemy or already has the `DMG_SWORD_BEAM` flag as is.

This also seemed like a good opportunity to use the new actor extension functionality in lieu of piggybacking on some unused variables, so I did that too.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2801674260.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2801683813.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2801739394.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
+++ b/mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
@@ -19,25 +19,25 @@ ActorExtensionId actorHitBySwordBeamExtId = 0;
 #define CVAR_NAME "gEnhancements.Masks.FierceDeitysAnywhere"
 #define CVAR CVarGetInteger(CVAR_NAME, 0)
 
-bool* GetActorHitBySwordBeam(Actor* actor) {
+bool GetActorHitBySwordBeam(Actor* actor) {
     bool* swordBeamCollisionFlag = (bool*)ActorExtension_Get(actor, actorHitBySwordBeamExtId);
     if (swordBeamCollisionFlag == NULL) {
-        return 0;
+        return false;
     }
 
-    return swordBeamCollisionFlag;
+    return *swordBeamCollisionFlag;
+}
+
+void SetActorHitBySwordBeam(Actor* actor, bool flag) {
+    bool* swordBeamCollisionFlag = (bool*)ActorExtension_Get(actor, actorHitBySwordBeamExtId);
+    if (swordBeamCollisionFlag != NULL) {
+        *swordBeamCollisionFlag = flag;
+    }
 }
 
 void RegisterFierceDeityAnywhere() {
     if (actorHitBySwordBeamExtId == 0) {
         actorHitBySwordBeamExtId = ActorExtension_CreateForAll(sizeof(bool));
-
-        GameInteractor::Instance->RegisterGameHook<GameInteractor::OnActorInit>([](Actor* actor) {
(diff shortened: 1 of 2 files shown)


mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
```diff
@@ -13,13 +14,32 @@ extern "C" {
...
-#define NOT_HIT_BY_SWORD_BEAM 0
+ActorExtensionId actorHitBySwordBeamExtId = 0;
 
 #define CVAR_NAME "gEnhancements.Masks.FierceDeitysAnywhere"
 #define CVAR CVarGetInteger(CVAR_NAME, 0)
 
+bool* GetActorHitBySwordBeam(Actor* actor) {
+    bool* swordBeamCollisionFlag = (bool*)ActorExtension_Get(actor, actorHitBySwordBeamExtId);
+    if (swordBeamCollisionFlag == NULL) {
+        return 0;
```
Archez: This func signature is returning pointers. Returning `0` here is the same as returning `NULL`. The problem is later in code you directly assign to this pointer without null checking.

My recommendation would be that this getter should return the dereferenced data as a standard bool, or false if NULL. Then you should make a `SetActorHitBySwordBeam` which allows you to set the value.

mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
```diff
@@ -13,13 +14,32 @@ extern "C" {
...
+    if (actorHitBySwordBeamExtId == 0) {
+        actorHitBySwordBeamExtId = ActorExtension_CreateForAll(sizeof(bool));
+
+        GameInteractor::Instance->RegisterGameHook<GameInteractor::OnActorInit>([](Actor* actor) {
+            bool* swordBeamCollisionFlag = (bool*)ActorExtension_Get(actor, actorHitBySwordBeamExtId);
+            if (swordBeamCollisionFlag == NULL) {
+                assert(false && "Actor Extension memory not valid");
+            }
+        });
+    }
```
Archez: This init is not actually doing anything, its just performing an assert check and thats it. Do you need this? Or did you mean to set the flag as false? (it already would be false though)

Eblo: Oops, I did set it to false and then accidentally removed that in cleanup. Is it guaranteed to be false if I don't set it?

Archez: Yes, in `ActorExtension_Alloc` the malloc'd data is `memset(data, 0, size)`

Eblo: Went ahead and removed the init since the default state of 0 is what I want.

mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
Eblo: Done.

Archez (approved):

PR #1087, opened 2025-03-23
[Rando] Shuffle Swim

Caladius:
Decided to take this a little seriously this time, this shuffles the ability to Swim and adds new Logic to handle it.

I went through the Regions and added CAN_USE_ABILITY(RI_ABILITY_SWIM) to the checks I could foresee needing it. If someone could review and playtest to ensure swim locations have been properly identified, please and thank you!

TO DO:
- [x] Ask Dana to create an Object/Texture for the Ability (probably something reusable as I plan to add other abilities to the shuffle list)
- [x] Finalize Logic identifiers for Swim.


mm/2s2h/Rando/ActorBehavior/Player.cpp
```diff
@@ -0,0 +1,35 @@
+#include "ActorBehavior.h"
+#include <libultraship/libultraship.h>
```
Eblo:
```suggestion
#include "public/bridge/consolevariablebridge.h"
```

mm/2s2h/Rando/ActorBehavior/Player.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+extern "C" {
+#include "variables.h"
+#include "functions.h"
+
+#include "include/z64player.h"
+}
+
+void Swim_Ability(Player* player) {
+    if (gPlayState->sceneId == SCENE_BOWLING) {
```
Eblo, line 16: Add a comment clarifying that this is Honey and Darling's Shop, and since touching the water ends the minigame anyway, we don't want this overriding that.

mm/2s2h/Rando/ActorBehavior/Player.cpp
```diff
@@ -0,0 +1,35 @@
...
+#include <libultraship/libultraship.h>
+
+extern "C" {
+#include "variables.h"
+#include "functions.h"
+
+#include "include/z64player.h"
+}
+
+void Swim_Ability(Player* player) {
```
Eblo: Since what this does is handle the voiding out if the player doesn't have the swim ability, it should be named something clearer. Maybe `RespawnOnWaterTouch`? Not sure.

mm/2s2h/Rando/ActorBehavior/Player.cpp
```diff
@@ -0,0 +1,35 @@
...
+        player->stateFlags1 |= PLAYER_STATE1_200;
+        Audio_PlaySfx(NA_SE_SY_DEKUNUTS_JUMP_FAILED);
+    }
+}
+
+void Rando::ActorBehavior::InitPlayerBehavior() {
+    COND_ID_HOOK(OnActorUpdate, ACTOR_PLAYER, IS_RANDO, [](Actor* actor) {
+        if (!RANDO_SAVE_OPTIONS[RO_SHUFFLE_SWIM]) {
+            return;
+        }
```
Eblo:
```suggestion
void Rando::ActorBehavior::InitPlayerBehavior() {
    COND_ID_HOOK(OnActorUpdate, ACTOR_PLAYER, IS_RANDO && RANDO_SAVE_OPTIONS[RO_SHUFFLE_SWIM], [](Actor* actor) {
```

mm/2s2h/Rando/Menu.cpp
```diff
@@ -242,8 +242,7 @@ static void DrawItemsTab() {
     f32 columnWidth = ImGui::GetContentRegionAvail().x / 3 - (ImGui::GetStyle().ItemSpacing.x * 2);
     f32 halfHeight = ImGui::GetContentRegionAvail().y / 3 - (ImGui::GetStyle().ItemSpacing.y * 2);
     ImGui::BeginChild("randoItemsColumn1", ImVec2(columnWidth, halfHeight));
-    CVarCheckbox("Bronze Scale", "gPlaceholderBool",
-                 CheckboxOptions({ { .disabled = true, .disabledTooltip = "Coming Soon" } }));
+    CVarCheckbox("Bronze Scale", Rando::StaticData::Options[RO_SHUFFLE_SWIM].cvar);
```
Eblo: Needs a tooltip explaining what this is.

mm/2s2h/Rando/Logic/Regions/Central.cpp
```diff
@@ -120,9 +120,9 @@ static RegisterShipInitFunc initFunc([]() {
     Regions[RR_CLOCK_TOWN_LAUNDRY] = RandoRegion{ .sceneId = SCENE_ALLEY,
         .checks = {
             CHECK(RC_CLOCK_TOWN_STRAY_FAIRY,                    true),
-            CHECK(RC_CLOCK_TOWN_LAUNDRY_FREESTANDING_RUPEE_01,  true),
-            CHECK(RC_CLOCK_TOWN_LAUNDRY_FREESTANDING_RUPEE_02,  true),
-            CHECK(RC_CLOCK_TOWN_LAUNDRY_FREESTANDING_RUPEE_03,  true),
+            CHECK(RC_CLOCK_TOWN_LAUNDRY_FREESTANDING_RUPEE_01,  CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_CLOCK_TOWN_LAUNDRY_FREESTANDING_RUPEE_02,  CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_CLOCK_TOWN_LAUNDRY_FREESTANDING_RUPEE_03,  CAN_USE_ABILITY(RI_ABILITY_SWIM)),
```
Eblo: These are rangable.
```suggestion
            CHECK(RC_CLOCK_TOWN_LAUNDRY_FREESTANDING_RUPEE_01,  CAN_USE_ABILITY(RI_ABILITY_SWIM) || CAN_BE_ZORA),
            CHECK(RC_CLOCK_TOWN_LAUNDRY_FREESTANDING_RUPEE_02,  CAN_USE_ABILITY(RI_ABILITY_SWIM) || CAN_BE_ZORA),
            CHECK(RC_CLOCK_TOWN_LAUNDRY_FREESTANDING_RUPEE_03,  CAN_USE_ABILITY(RI_ABILITY_SWIM) || CAN_BE_ZORA),
```

mm/2s2h/Rando/Logic/Regions/East.cpp
```diff
@@ -146,12 +146,12 @@ static RegisterShipInitFunc initFunc([]() {
                 Must have delivered the Letter to Kafei and met Kafei.(Sakon just does not show up otherwise, as odd as that may sound.)
             */
             EXIT(ENTRANCE(SAKONS_HIDEOUT, 0),               ENTRANCE(IKANA_CANYON, 6), Flags_GetRandoInf(RANDO_INF_OBTAINED_LETTER_TO_KAFEI)),
-            EXIT(ENTRANCE(SECRET_SHRINE, 0),                ENTRANCE(IKANA_CANYON, 12), true),
-            EXIT(ENTRANCE(SOUTHERN_SWAMP_POISONED, 9),               ONE_WAY_EXIT, true),
+            EXIT(ENTRANCE(SECRET_SHRINE, 0),                ENTRANCE(IKANA_CANYON, 12), CAN_USE_ABILITY(RI_ABILITY_SWIM)),
```
Eblo: I tried getting through this with ice arrows. When you cross a certain point, it just instantly despawns ice platforms made on the other side of the tunnel. Well then.

mm/2s2h/Rando/Logic/Regions/East.cpp
```diff
@@ -363,12 +363,12 @@ static RegisterShipInitFunc initFunc([]() {
...
-            CHECK(RC_SECRET_SHRINE_POT_06, CAN_BE_ZORA),
-            CHECK(RC_SECRET_SHRINE_POT_07, CAN_BE_ZORA),
-            CHECK(RC_SECRET_SHRINE_POT_08, CAN_BE_ZORA),
-            CHECK(RC_SECRET_SHRINE_POT_09, CAN_BE_ZORA),
+            CHECK(RC_SECRET_SHRINE_POT_04, CAN_BE_ZORA || CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_SECRET_SHRINE_POT_05, CAN_BE_ZORA || CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_SECRET_SHRINE_POT_06, CAN_BE_ZORA || CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_SECRET_SHRINE_POT_07, CAN_BE_ZORA || CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_SECRET_SHRINE_POT_08, CAN_BE_ZORA || CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_SECRET_SHRINE_POT_09, CAN_BE_ZORA || CAN_USE_ABILITY(RI_ABILITY_SWIM)),
```
Eblo: I think what was here before was insufficient. If you can break them from a distance, then even human Link can dive down to get them.
```suggestion
            CHECK(RC_SECRET_SHRINE_POT_04, CAN_USE_PROJECTILE && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
            CHECK(RC_SECRET_SHRINE_POT_05, CAN_USE_PROJECTILE && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
            CHECK(RC_SECRET_SHRINE_POT_06, CAN_USE_PROJECTILE && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
            CHECK(RC_SECRET_SHRINE_POT_07, CAN_USE_PROJECTILE && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
            CHECK(RC_SECRET_SHRINE_POT_08, CAN_USE_PROJECTILE && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
            CHECK(RC_SECRET_SHRINE_POT_09, CAN_USE_PROJECTILE && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
```

mm/2s2h/Rando/Logic/Regions/GreatBayTemple.cpp
```diff
@@ -66,14 +66,14 @@ static RegisterShipInitFunc initFunc([]() {
             CHECK(RC_GREAT_BAY_TEMPLE_SF_CENTRAL_ROOM_BARREL,           true),
             CHECK(RC_GREAT_BAY_TEMPLE_CENTRAL_ROOM_POT_01,              true),
             CHECK(RC_GREAT_BAY_TEMPLE_CENTRAL_ROOM_POT_02,              true),
-            CHECK(RC_GREAT_BAY_TEMPLE_SF_CENTRAL_ROOM_UNDERWATER_POT,   CAN_BE_ZORA),
+            CHECK(RC_GREAT_BAY_TEMPLE_SF_CENTRAL_ROOM_UNDERWATER_POT,   CAN_BE_ZORA && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
         },
         .connections = {
-            CONNECTION(RR_GREAT_BAY_TEMPLE_COMPASS_ROOM_TUNNEL,     CAN_BE_ZORA && GBT_CAN_REVERSE_WATER_FLOW),
+            CONNECTION(RR_GREAT_BAY_TEMPLE_COMPASS_ROOM_TUNNEL,     CAN_BE_ZORA && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
```
Eblo: Should this be ANDed with `GBT_CAN_REVERSE_WATER_FLOW`, or does swimming preclude that?

mm/2s2h/Rando/Logic/Regions/GreatBayTemple.cpp
```diff
@@ -94,27 +94,27 @@ static RegisterShipInitFunc initFunc([]() {
...
             CHECK(RC_GREAT_BAY_TEMPLE_COMPASS_ROOM_SURFACE_POT_01, true),
             CHECK(RC_GREAT_BAY_TEMPLE_COMPASS_ROOM_SURFACE_POT_02, true),
             CHECK(RC_GREAT_BAY_TEMPLE_COMPASS_ROOM_SURFACE_POT_03, true),
             CHECK(RC_GREAT_BAY_TEMPLE_COMPASS_ROOM_SURFACE_POT_04, true),
-            CHECK(RC_GREAT_BAY_TEMPLE_COMPASS_ROOM_WATER_POT_01,   CAN_BE_ZORA),
-            CHECK(RC_GREAT_BAY_TEMPLE_COMPASS_ROOM_WATER_POT_02,   CAN_BE_ZORA),
-            CHECK(RC_GREAT_BAY_TEMPLE_COMPASS_ROOM_WATER_POT_03,   CAN_BE_ZORA),
+            CHECK(RC_GREAT_BAY_TEMPLE_COMPASS_ROOM_WATER_POT_01,   CAN_BE_ZORA&& CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_GREAT_BAY_TEMPLE_COMPASS_ROOM_WATER_POT_02,   CAN_BE_ZORA&& CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_GREAT_BAY_TEMPLE_COMPASS_ROOM_WATER_POT_03,   CAN_BE_ZORA&& CAN_USE_ABILITY(RI_ABILITY_SWIM)),
```
Eblo:
```suggestion
            CHECK(RC_GREAT_BAY_TEMPLE_COMPASS_ROOM_WATER_POT_01,   CAN_BE_ZORA && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
            CHECK(RC_GREAT_BAY_TEMPLE_COMPASS_ROOM_WATER_POT_02,   CAN_BE_ZORA && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
            CHECK(RC_GREAT_BAY_TEMPLE_COMPASS_ROOM_WATER_POT_03,   CAN_BE_ZORA && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
```

mm/2s2h/Rando/Logic/Regions/GreatBayTemple.cpp
```diff
@@ -321,7 +321,7 @@ static RegisterShipInitFunc initFunc([]() {
     };
     Regions[RR_GREAT_BAY_TEMPLE_WATER_WHEEL_ROOM] = RandoRegion{ .sceneId = SCENE_SEA,
         .checks = {
-            CHECK(RC_GREAT_BAY_TEMPLE_SF_WATER_WHEEL_PLATFORM,           CAN_BE_ZORA || (HAS_ITEM(ITEM_BOW) && (HAS_ITEM(ITEM_MASK_GREAT_FAIRY)))),
+            CHECK(RC_GREAT_BAY_TEMPLE_SF_WATER_WHEEL_PLATFORM,           CAN_BE_ZORA && CAN_USE_ABILITY(RI_ABILITY_SWIM) || (HAS_ITEM(ITEM_BOW) && (HAS_ITEM(ITEM_MASK_GREAT_FAIRY)))),
```
Eblo: Are you positive that this order of operations is correct and that there aren't parentheses needed somewhere?

mm/2s2h/Rando/Logic/Regions/North.cpp
```diff
@@ -318,12 +318,12 @@ static RegisterShipInitFunc initFunc([]() {
...
-            CHECK(RC_TWIN_ISLANDS_FREESTANDING_RUPEE_03, CAN_BE_ZORA && RANDO_EVENTS[RE_CLEARED_SNOWHEAD_TEMPLE]),
-            CHECK(RC_TWIN_ISLANDS_FREESTANDING_RUPEE_04, CAN_BE_ZORA && RANDO_EVENTS[RE_CLEARED_SNOWHEAD_TEMPLE]),
+            CHECK(RC_TWIN_ISLANDS_UNDERWATER_CHEST_01,    CAN_BE_ZORA && CAN_USE_ABILITY(RI_ABILITY_SWIM) && RANDO_EVENTS[RE_CLEARED_SNOWHEAD_TEMPLE]),
+            CHECK(RC_TWIN_ISLANDS_UNDERWATER_CHEST_02,    CAN_BE_ZORA && CAN_USE_ABILITY(RI_ABILITY_SWIM) && RANDO_EVENTS[RE_CLEARED_SNOWHEAD_TEMPLE]),
+            CHECK(RC_TWIN_ISLANDS_FREESTANDING_RUPEE_01, CAN_BE_ZORA && CAN_USE_ABILITY(RI_ABILITY_SWIM) && RANDO_EVENTS[RE_CLEARED_SNOWHEAD_TEMPLE]),
+            CHECK(RC_TWIN_ISLANDS_FREESTANDING_RUPEE_02, CAN_BE_ZORA && CAN_USE_ABILITY(RI_ABILITY_SWIM) && RANDO_EVENTS[RE_CLEARED_SNOWHEAD_TEMPLE]),
+            CHECK(RC_TWIN_ISLANDS_FREESTANDING_RUPEE_03, CAN_BE_ZORA && CAN_USE_ABILITY(RI_ABILITY_SWIM) && RANDO_EVENTS[RE_CLEARED_SNOWHEAD_TEMPLE]),
+            CHECK(RC_TWIN_ISLANDS_FREESTANDING_RUPEE_04, CAN_BE_ZORA && CAN_USE_ABILITY(RI_ABILITY_SWIM) && RANDO_EVENTS[RE_CLEARED_SNOWHEAD_TEMPLE]),
```
Eblo: These are technically attainable without swim, with some ice arrow and boomerang trickery, but I'm not sure if it's fair to logically consider that. Could you at least put a note saying that this is doable? It may be worth considering for tricks in the future.

mm/2s2h/Rando/Logic/Regions/South.cpp
```diff
@@ -320,8 +320,8 @@ static RegisterShipInitFunc initFunc([]() {
         .exits = { //     TO                                         FROM
             EXIT(ENTRANCE(ROAD_TO_SOUTHERN_SWAMP, 1),       ENTRANCE(SOUTHERN_SWAMP_POISONED, 0), true),
             EXIT(ENTRANCE(TOURIST_INFORMATION, 0),          ENTRANCE(SOUTHERN_SWAMP_POISONED, 1), true),
-            EXIT(ENTRANCE(MAGIC_HAGS_POTION_SHOP, 0),       ENTRANCE(SOUTHERN_SWAMP_POISONED, 5), true),
-            EXIT(ENTRANCE(WOODS_OF_MYSTERY, 0),             ENTRANCE(SOUTHERN_SWAMP_POISONED, 7), true),
+            EXIT(ENTRANCE(MAGIC_HAGS_POTION_SHOP, 0),       ENTRANCE(SOUTHERN_SWAMP_POISONED, 5), CAN_BE_DEKU || CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            EXIT(ENTRANCE(WOODS_OF_MYSTERY, 0),             ENTRANCE(SOUTHERN_SWAMP_POISONED, 7), CAN_BE_DEKU || CAN_USE_ABILITY(RI_ABILITY_SWIM)),
```
Eblo: Goron and Zora can just walk through this water without swimming. Though now that I'm looking closely, that whole area might need to be split into a region and have the connections handle the swimming, rather than entrances. With this logic as-is, any mundane check outside of the potion shop, for example, is not logically considering swimming. Ditto other areas in the swamp.

mm/2s2h/Rando/Logic/Regions/West.cpp
```diff
@@ -121,27 +121,27 @@ static RegisterShipInitFunc initFunc([]() {
     };
     Regions[RR_GREAT_BAY_COAST] = RandoRegion{ .sceneId = SCENE_30GYOSON,
         .checks = {
-            CHECK(RC_GREAT_BAY_COAST_FISHERMAN_MINIGAME, RANDO_EVENTS[RE_CLEARED_GREAT_BAY_TEMPLE] && (HAS_ITEM(ITEM_HOOKSHOT) || CAN_USE_MAGIC_ARROW(ICE))),
-            CHECK(RC_GREAT_BAY_COAST_OWL_STATUE, CAN_USE_SWORD),
+            CHECK(RC_GREAT_BAY_COAST_FISHERMAN_MINIGAME, RANDO_EVENTS[RE_CLEARED_GREAT_BAY_TEMPLE] && (HAS_ITEM(ITEM_HOOKSHOT) || CAN_USE_MAGIC_ARROW(ICE)) && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
```
Eblo: You do not need to swim to reach this check. In fact, I don't think you *can* reach it by swimming.

mm/2s2h/Rando/Logic/Regions/West.cpp
```diff
@@ -121,27 +121,27 @@ static RegisterShipInitFunc initFunc([]() {
...
+            CHECK(RC_GREAT_BAY_COAST_OWL_STATUE, CAN_USE_SWORD && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
```
Eblo: You can also get here if owl statues are shuffled and you warp directly here.

mm/2s2h/Rando/Logic/Regions/West.cpp
```diff
@@ -121,27 +121,27 @@ static RegisterShipInitFunc initFunc([]() {
...
-            CHECK(RC_GREAT_BAY_COAST_POT_09, true),
-            CHECK(RC_GREAT_BAY_COAST_POT_10, true),
-            CHECK(RC_GREAT_BAY_COAST_POT_11, true),
-            CHECK(RC_GREAT_BAY_COAST_POT_12, true),
-            CHECK(RC_GREAT_BAY_COAST_LEDGE_POT_01, HAS_ITEM(ITEM_HOOKSHOT)),
-            CHECK(RC_GREAT_BAY_COAST_LEDGE_POT_02, HAS_ITEM(ITEM_HOOKSHOT)),
-            CHECK(RC_GREAT_BAY_COAST_LEDGE_POT_03, HAS_ITEM(ITEM_HOOKSHOT)),
-            CHECK(RC_GREAT_BAY_COAST_TINGLE_MAP_01, (HAS_ITEM(ITEM_BOW) || HAS_ITEM(ITEM_HOOKSHOT)) && CAN_AFFORD(RC_GREAT_BAY_COAST_TINGLE_MAP_01)),
-            CHECK(RC_GREAT_BAY_COAST_TINGLE_MAP_02, (HAS_ITEM(ITEM_BOW) || HAS_ITEM(ITEM_HOOKSHOT)) && CAN_AFFORD(RC_GREAT_BAY_COAST_TINGLE_MAP_02)),
+            CHECK(RC_GREAT_BAY_COAST_PIECE_OF_HEART, CAN_HOOK_SCARECROW && CAN_GROW_BEAN_PLANT && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
```
Eblo: Another place where you can technically use ice arrows to avoid swimming.

mm/2s2h/Rando/Logic/Regions/West.cpp
```diff
@@ -121,27 +121,27 @@ static RegisterShipInitFunc initFunc([]() {
...
+            CHECK(RC_GREAT_BAY_COAST_POT_08, CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_GREAT_BAY_COAST_POT_09, CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_GREAT_BAY_COAST_POT_10, CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_GREAT_BAY_COAST_POT_11, CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_GREAT_BAY_COAST_POT_12, CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_GREAT_BAY_COAST_LEDGE_POT_01, HAS_ITEM(ITEM_HOOKSHOT) && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_GREAT_BAY_COAST_LEDGE_POT_02, HAS_ITEM(ITEM_HOOKSHOT) && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_GREAT_BAY_COAST_LEDGE_POT_03, HAS_ITEM(ITEM_HOOKSHOT) && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_GREAT_BAY_COAST_TINGLE_MAP_01, (HAS_ITEM(ITEM_BOW) || HAS_ITEM(ITEM_HOOKSHOT)) && CAN_AFFORD(RC_GREAT_BAY_COAST_TINGLE_MAP_01) && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_GREAT_BAY_COAST_TINGLE_MAP_02, (HAS_ITEM(ITEM_BOW) || HAS_ITEM(ITEM_HOOKSHOT)) && CAN_AFFORD(RC_GREAT_BAY_COAST_TINGLE_MAP_02) && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
```
Eblo: All of these could be reached by owl warping, but I'm not sure how we might logically gate that.

mm/2s2h/Rando/Logic/Regions/West.cpp
```diff
@@ -151,14 +151,14 @@ static RegisterShipInitFunc initFunc([]() {
         .exits = { //     TO                                         FROM
             EXIT(ENTRANCE(TERMINA_FIELD, 2),                ENTRANCE(GREAT_BAY_COAST, 0), true),
             EXIT(ENTRANCE(ZORA_CAPE, 0),                    ENTRANCE(GREAT_BAY_COAST, 1), true),
-            EXIT(ENTRANCE(PINNACLE_ROCK, 0),                ENTRANCE(GREAT_BAY_COAST, 3), true),
+            EXIT(ENTRANCE(PINNACLE_ROCK, 0),                ENTRANCE(GREAT_BAY_COAST, 3), CAN_USE_ABILITY(RI_ABILITY_SWIM)),
             EXIT(ENTRANCE(FISHERMANS_HUT, 0),               ENTRANCE(GREAT_BAY_COAST, 4), true),
-            EXIT(ENTRANCE(PIRATES_FORTRESS_EXTERIOR, 0),    ENTRANCE(GREAT_BAY_COAST, 5), CAN_BE_ZORA),
-            EXIT(ENTRANCE(MARINE_RESEARCH_LAB, 0),          ENTRANCE(GREAT_BAY_COAST, 7), true),
+            EXIT(ENTRANCE(PIRATES_FORTRESS_EXTERIOR, 0),    ENTRANCE(GREAT_BAY_COAST, 5), CAN_BE_ZORA && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            EXIT(ENTRANCE(MARINE_RESEARCH_LAB, 0),          ENTRANCE(GREAT_BAY_COAST, 7), CAN_USE_ABILITY(RI_ABILITY_SWIM)),
```
Eblo: Technically swimming is not required for this entrance. Maybe we need to sequester the lab exterior into its own region.

mm/2s2h/Rando/Logic/Regions/West.cpp
```diff
@@ -151,14 +151,14 @@ static RegisterShipInitFunc initFunc([]() {
...
             EXIT(ENTRANCE(OCEANSIDE_SPIDER_HOUSE, 0),       ENTRANCE(GREAT_BAY_COAST, 8), true),
         },
         .connections = {
-            CONNECTION(RR_GREAT_BAY_COAST_COW_GROTTO, CAN_HOOK_SCARECROW && CAN_GROW_BEAN_PLANT), // TODO: Grotto mapping
+            CONNECTION(RR_GREAT_BAY_COAST_COW_GROTTO, CAN_HOOK_SCARECROW && CAN_GROW_BEAN_PLANT && CAN_USE_ABILITY(RI_ABILITY_SWIM)), // TODO: Grotto mapping
```
Eblo: blah blah ice arrows blah blah

Eblo (changes requested):
To perhaps no surprise, the bulk of review is logic. The tl;dr of my review is that there are a number of places where you could get around swimming by using ice arrows, maybe with some rangs added to the mix. That seems to push the boundary of "logic" versus tricks, though. I'm fine with such areas not considering that tactic, but they could be commented just so we have reference for it in the future.

Another big thing is to consider defining new regions in affected areas so we don't have to slap the swim ability on so many individual checks. The exterior of the Potion Shop is one place, as is the exterior to the Marine Research Lab. There are other logic areas not touched that should require swim. The Moon is the first place to come to mind, but there may be others.

mm/2s2h/Rando/Logic/Regions/West.cpp
Caladius: Owl warp additions should probably be added through a separate PR as to not Scope Creep, however I don't think the Pots here are the ones behind the Lab.

mm/2s2h/Rando/Logic/Regions/West.cpp
Caladius: I feel Ice Arrow logic should be considered a Trick added later, you also need to account for how much Magic it costs to cross the gap and if the player has something to restore magic or not.

mm/2s2h/Rando/ActorBehavior/Player.cpp
Caladius: OnSwimStateChange? i dont have a preference here. The naming was left over from when I had ambitions of shuffling more than just Swim in this PR.

mm/2s2h/Rando/ActorBehavior/Player.cpp
```diff
@@ -0,0 +1,33 @@
...
+
+extern "C" {
+#include "variables.h"
+#include "functions.h"
+
+#include "include/z64player.h"
+}
+
+void Swim_Ability(Player* player) {
+    // This is Honey & Darlings Shop, touching the water ends the minigame as its vanilla behaivor.
```
Eblo:
```suggestion
    // This is Honey & Darlings Shop, touching the water ends the minigame as its vanilla behavior.
```

mm/2s2h/Rando/ActorBehavior/Player.cpp
Eblo: Let's go with `RespawnOnWaterTouch`.

mm/2s2h/Rando/Logic/Logic.h
```diff
@@ -98,6 +98,7 @@ extern std::unordered_map<RandoRegionId, RandoRegion> Regions;
 #define FOUND_ALL_FROGS                                                                  \
     (CHECK_WEEKEVENTREG(WEEKEVENTREG_33_01) && CHECK_WEEKEVENTREG(WEEKEVENTREG_32_40) && \
      CHECK_WEEKEVENTREG(WEEKEVENTREG_32_80) && CHECK_WEEKEVENTREG(WEEKEVENTREG_33_02))
+#define CAN_USE_ABILITY(ability) (Flags_GetRandoInf(ability - RI_ABILITY_SWIM + RANDO_INF_OBTAINED_SWIM))
```
Eblo: I think we should go this route, to be more in line with other macros such as `CAN_USE_MAGIC_ARROW`. This would allow us to use `CAN_USE_ABILITY(SWIM)` instead of listing the full RI every time and just generally look cleaner in the check tracker:

```suggestion
#define CAN_USE_ABILITY(ability) (Flags_GetRandoInf(RI_ABILITY_##ability - RI_ABILITY_SWIM + RANDO_INF_OBTAINED_SWIM))
```

mm/2s2h/Rando/Logic/Regions/South.cpp
```diff
@@ -318,13 +341,12 @@ static RegisterShipInitFunc initFunc([]() {
             CHECK(RC_SOUTHERN_SWAMP_POISON_GRASS_30, true),
         },
         .exits = { //     TO                                         FROM
-            EXIT(ENTRANCE(ROAD_TO_SOUTHERN_SWAMP, 1),       ENTRANCE(SOUTHERN_SWAMP_POISONED, 0), true),
-            EXIT(ENTRANCE(TOURIST_INFORMATION, 0),          ENTRANCE(SOUTHERN_SWAMP_POISONED, 1), true),
-            EXIT(ENTRANCE(MAGIC_HAGS_POTION_SHOP, 0),       ENTRANCE(SOUTHERN_SWAMP_POISONED, 5), true),
-            EXIT(ENTRANCE(WOODS_OF_MYSTERY, 0),             ENTRANCE(SOUTHERN_SWAMP_POISONED, 7), true),
+            EXIT(ENTRANCE(MAGIC_HAGS_POTION_SHOP, 0),       ENTRANCE(SOUTHERN_SWAMP_POISONED, 5), CAN_BE_DEKU || CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            EXIT(ENTRANCE(WOODS_OF_MYSTERY, 0),             ENTRANCE(SOUTHERN_SWAMP_POISONED, 7), CAN_BE_DEKU || CAN_USE_ABILITY(RI_ABILITY_SWIM)),
```
Eblo: These entrances themselves are not gated; once at the potion shop exterior there's nothing logically required to enter the potion shop or woods.
```suggestion
            EXIT(ENTRANCE(MAGIC_HAGS_POTION_SHOP, 0),       ENTRANCE(SOUTHERN_SWAMP_POISONED, 5), true)),
            EXIT(ENTRANCE(WOODS_OF_MYSTERY, 0),             ENTRANCE(SOUTHERN_SWAMP_POISONED, 7), true),
```

mm/2s2h/Rando/Logic/Regions/West.cpp
```diff
@@ -151,14 +151,14 @@ static RegisterShipInitFunc initFunc([]() {
         .exits = { //     TO                                         FROM
             EXIT(ENTRANCE(TERMINA_FIELD, 2),                ENTRANCE(GREAT_BAY_COAST, 0), true),
             EXIT(ENTRANCE(ZORA_CAPE, 0),                    ENTRANCE(GREAT_BAY_COAST, 1), true),
-            EXIT(ENTRANCE(PINNACLE_ROCK, 0),                ENTRANCE(GREAT_BAY_COAST, 3), true),
+            EXIT(ENTRANCE(PINNACLE_ROCK, 0),                ENTRANCE(GREAT_BAY_COAST, 3), CAN_USE_ABILITY(RI_ABILITY_SWIM)),
             EXIT(ENTRANCE(FISHERMANS_HUT, 0),               ENTRANCE(GREAT_BAY_COAST, 4), true),
-            EXIT(ENTRANCE(PIRATES_FORTRESS_EXTERIOR, 0),    ENTRANCE(GREAT_BAY_COAST, 5), CAN_BE_ZORA),
-            EXIT(ENTRANCE(MARINE_RESEARCH_LAB, 0),          ENTRANCE(GREAT_BAY_COAST, 7), true),
+            EXIT(ENTRANCE(PIRATES_FORTRESS_EXTERIOR, 0),    ENTRANCE(GREAT_BAY_COAST, 5), CAN_BE_ZORA && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            EXIT(ENTRANCE(MARINE_RESEARCH_LAB, 0),          ENTRANCE(GREAT_BAY_COAST, 7), (CAN_USE_ABILITY(RI_ABILITY_SWIM) || (CAN_PLAY_SONG(SOARING) && CAN_OWL_WARP(OWL_WARP_GREAT_BAY_COAST)))),
```
Eblo: I think this, along with every check immediately surrounding the exterior, needs to be in a new region with connections based on the swim ability.

mm/2s2h/Rando/Logic/Regions/West.cpp
```diff
@@ -121,27 +121,27 @@ static RegisterShipInitFunc initFunc([]() {
     };
     Regions[RR_GREAT_BAY_COAST] = RandoRegion{ .sceneId = SCENE_30GYOSON,
         .checks = {
-            CHECK(RC_GREAT_BAY_COAST_FISHERMAN_MINIGAME, RANDO_EVENTS[RE_CLEARED_GREAT_BAY_TEMPLE] && (HAS_ITEM(ITEM_HOOKSHOT) || CAN_USE_MAGIC_ARROW(ICE))),
-            CHECK(RC_GREAT_BAY_COAST_OWL_STATUE, CAN_USE_SWORD),
+            CHECK(RC_GREAT_BAY_COAST_FISHERMAN_MINIGAME, RANDO_EVENTS[RE_CLEARED_GREAT_BAY_TEMPLE] && (HAS_ITEM(ITEM_HOOKSHOT) || CAN_USE_MAGIC_ARROW(ICE)) && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_GREAT_BAY_COAST_OWL_STATUE, (CAN_USE_SWORD && CAN_USE_ABILITY(RI_ABILITY_SWIM)) || (CAN_PLAY_SONG(SOARING) && CAN_OWL_WARP(OWL_WARP_GREAT_BAY_COAST))),
```
Eblo: See my other comment regarding a Marine Research Lab region. The Owl Statue check itself should only need a sword; the logic of getting to the owl statue can be handled by regions and connections.

mm/2s2h/Rando/Logic/Regions/West.cpp
```diff
@@ -121,27 +121,27 @@ static RegisterShipInitFunc initFunc([]() {
...
+            CHECK(RC_GREAT_BAY_COAST_POT_11, CAN_USE_ABILITY(RI_ABILITY_SWIM) || (CAN_PLAY_SONG(SOARING) && CAN_OWL_WARP(OWL_WARP_GREAT_BAY_COAST))),
+            CHECK(RC_GREAT_BAY_COAST_POT_12, CAN_USE_ABILITY(RI_ABILITY_SWIM) || (CAN_PLAY_SONG(SOARING) && CAN_OWL_WARP(OWL_WARP_GREAT_BAY_COAST))),
+            CHECK(RC_GREAT_BAY_COAST_LEDGE_POT_01, HAS_ITEM(ITEM_HOOKSHOT) && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_GREAT_BAY_COAST_LEDGE_POT_02, HAS_ITEM(ITEM_HOOKSHOT) && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_GREAT_BAY_COAST_LEDGE_POT_03, HAS_ITEM(ITEM_HOOKSHOT) && CAN_USE_ABILITY(RI_ABILITY_SWIM)),
+            CHECK(RC_GREAT_BAY_COAST_TINGLE_MAP_01, (HAS_ITEM(ITEM_BOW) || HAS_ITEM(ITEM_HOOKSHOT)) && CAN_AFFORD(RC_GREAT_BAY_COAST_TINGLE_MAP_01) && (CAN_USE_ABILITY(RI_ABILITY_SWIM) || (CAN_PLAY_SONG(SOARING) && CAN_OWL_WARP(OWL_WARP_GREAT_BAY_COAST)))),
+            CHECK(RC_GREAT_BAY_COAST_TINGLE_MAP_02, (HAS_ITEM(ITEM_BOW) || HAS_ITEM(ITEM_HOOKSHOT)) && CAN_AFFORD(RC_GREAT_BAY_COAST_TINGLE_MAP_02) && (CAN_USE_ABILITY(RI_ABILITY_SWIM) || (CAN_PLAY_SONG(SOARING) && CAN_OWL_WARP(OWL_WARP_GREAT_BAY_COAST)))),
```
Eblo: I goofed when discussing the warping. All warping is already handled in one location, so it would be redundant to consider it for individual checks.

Eblo (changes requested):
I am in full agreement with the ice arrows, so let's forget about that suggestion. I do however have some outstanding feedback and questions, as well as more now.

mm/2s2h/Rando/Logic/Regions/West.cpp
Caladius: For what purpose? You need to be able to Swim to even enter the Exterior to begin with right? Requiring it one time for the ENTRANCE seems much less invasive then creating a whole new region.

I think I'm missing something with your reasoning haha.

Eblo: > You need to be able to Swim to even enter the Exterior to begin with right?

Owl warp aside, that may be true *now*, but the distinction will be significant for entrance rando, as there would be a new route to reach this area without SoS or swim. I think the correct approach for the three ways of reaching the exterior is:

1. Swim there from the beach, which requires swim. This would be logically represented by having a region for the exterior, with a connection to GBC that requires swim.
2. Owl statue, which requires ocarina+SoS. I think what you have on the entrance logic here for owl warp can just be removed, as owl warps are logically handled altogether elsewhere.
3. Entrance from interior. Right now there's no difference since you can't get to the interior without already reaching the exterior, but with entrance rando this will matter. So far we've been doing a solid job of covering entrance considerations along the way instead of putting it off, so I'd like to keep that going.

Caladius: Ok, if you are planning ahead for Entrance Shuffle I can see the intent. Now, which RR_ specifically do you want separated? I see we already have MOAT split into upper and lower regions already which account for SWIM currently.

Eblo:
Code-wise, LGTM. Approval pending the ability model.

mm/2s2h/Rando/Logic/Regions/South.cpp
```diff
@@ -268,6 +280,48 @@ static RegisterShipInitFunc initFunc([]() {
...
+            CHECK(RC_SOUTHERN_SWAMP_POISON_GRASS_10, true),
+            CHECK(RC_SOUTHERN_SWAMP_POISON_GRASS_11, true),
+            CHECK(RC_SOUTHERN_SWAMP_POISON_GRASS_12, true),
+        },
+        .exits = { //     TO                                         FROM
+            EXIT(ENTRANCE(ROAD_TO_SOUTHERN_SWAMP, 1),       ENTRANCE(SOUTHERN_SWAMP_POISONED, 0), true),
+            EXIT(ENTRANCE(TOURIST_INFORMATION, 0),          ENTRANCE(SOUTHERN_SWAMP_POISONED, 1), true),
+        },
+        .connections = {
+            CONNECTION(RR_SOUTHERN_SWAMP_SOUTH, CAN_GET_PAST_OCTOROK),
```
Eblo:
```suggestion
            CONNECTION(RR_SOUTHERN_SWAMP_SOUTH, CanGetPastBigOcto()),
```

mm/2s2h/Rando/Logic/Regions/South.cpp
```diff
@@ -318,13 +360,12 @@ static RegisterShipInitFunc initFunc([]() {
...
         },
         .exits = { //     TO                                         FROM
-            EXIT(ENTRANCE(ROAD_TO_SOUTHERN_SWAMP, 1),       ENTRANCE(SOUTHERN_SWAMP_POISONED, 0), true),
-            EXIT(ENTRANCE(TOURIST_INFORMATION, 0),          ENTRANCE(SOUTHERN_SWAMP_POISONED, 1), true),
             EXIT(ENTRANCE(MAGIC_HAGS_POTION_SHOP, 0),       ENTRANCE(SOUTHERN_SWAMP_POISONED, 5), true),
             EXIT(ENTRANCE(WOODS_OF_MYSTERY, 0),             ENTRANCE(SOUTHERN_SWAMP_POISONED, 7), true),
         },
         .connections = {
-            CONNECTION(RR_SOUTHERN_SWAMP_SOUTH, (RANDO_EVENTS[RE_SOUTHERN_SWAMP_KILL_OCTOROK] || RANDO_EVENTS[RE_CLEARED_WOODFALL_TEMPLE])),
+            CONNECTION(RR_SOUTHERN_SWAMP_SOUTH, CAN_GET_PAST_OCTOROK),
```
Eblo:
```suggestion
            CONNECTION(RR_SOUTHERN_SWAMP_SOUTH, CanGetPastBigOcto()),
```

mm/2s2h/Rando/Logic/Regions/South.cpp
```diff
@@ -348,7 +388,8 @@ static RegisterShipInitFunc initFunc([]() {
             EXIT(ENTRANCE(SWAMP_SPIDER_HOUSE, 0),           ENTRANCE(SOUTHERN_SWAMP_POISONED, 8), CAN_LIGHT_TORCH_NEAR_ANOTHER && (CAN_BE_DEKU || RANDO_EVENTS[RE_CLEARED_WOODFALL_TEMPLE])),
         },
         .connections = {
-            CONNECTION(RR_SOUTHERN_SWAMP_NORTH, (RANDO_EVENTS[RE_SOUTHERN_SWAMP_KILL_OCTOROK] || RANDO_EVENTS[RE_CLEARED_WOODFALL_TEMPLE])),
+            CONNECTION(RR_SOUTHERN_SWAMP_NORTH, CAN_GET_PAST_OCTOROK),
+            CONNECTION(RR_SOUTHERN_SWAMP_NEAR_WOODS, CAN_GET_PAST_OCTOROK),
```
Eblo:
```suggestion
            CONNECTION(RR_SOUTHERN_SWAMP_NORTH, CanGetPastBigOcto()),
            CONNECTION(RR_SOUTHERN_SWAMP_NEAR_WOODS, CanGetPastBigOcto()),
```

mm/2s2h/Rando/Logic/Regions/South.cpp
```diff
@@ -6,6 +6,25 @@
...
+ *    the damage boost method not being logically considered. This piece may need expanded as tricks and glitches are
+ *    logically introduced.
+ * 3. Clear Woodfall Temple. The water is clear in this state, so swimming comes into question. Either Link uses the
+ *    swim ability, he hops over the water as Deku, or he uses one of the forms tall enough to walk through the water.
+ *    This may also be expanded if FD is ever logically considered.
+ */
+#define CAN_GET_PAST_OCTOROK \
+    (RANDO_EVENTS[RE_SOUTHERN_SWAMP_RIDE_BOAT] || \
+    (RANDO_EVENTS[RE_SOUTHERN_SWAMP_KILL_OCTOROK] && CAN_BE_DEKU) || \
+    (RANDO_EVENTS[RE_CLEARED_WOODFALL_TEMPLE] && (CAN_USE_ABILITY(SWIM) || CAN_BE_DEKU || CAN_BE_ZORA || CAN_BE_GORON)))
```
Eblo: Please put this before `//clang-format off`; GitHub won't let me suggest outside of the modified code:

```suggestion
inline bool CanGetPastBigOcto() {
    // Boat ride. This takes Link to a landing in the next region and thus has no additional requirements.
    if (RANDO_EVENTS[RE_SOUTHERN_SWAMP_RIDE_BOAT]) {
        return true;
    }
    /*
     * Kill the Big Octo. This means the water is still in a poison state, so swimming is not logically considered due
     * to the damage boost method not being logically considered. This piece may need expanded as tricks and glitches
     * are logically introduced.
     */
    if (RANDO_EVENTS[RE_SOUTHERN_SWAMP_KILL_OCTOROK] && CAN_BE_DEKU) {
        return true;
    }
    /*
     * The water is clear in this state, so swimming comes into question. Either Link uses the swim ability, he hops
     * over the water as Deku, or he uses one of the forms tall enough to walk through the water. This may also be
     * expanded if FD is ever logically considered.
     */
    if (RANDO_EVENTS[RE_CLEARED_WOODFALL_TEMPLE] &&
        (CAN_USE_ABILITY(SWIM) || CAN_BE_DEKU || CAN_BE_ZORA || CAN_BE_GORON)) {
            return true;
        }
    return false;
}
```

Eblo (commented):
Flippers model looks good! One last thing, hopefully. I know the Big Octo macro was my initial suggestion, but after reconsidering I think it should be an inline bool.

mm/2s2h/Rando/Logic/Regions/West.cpp
```diff
@@ -122,26 +122,9 @@ static RegisterShipInitFunc initFunc([]() {
     Regions[RR_GREAT_BAY_COAST] = RandoRegion{ .sceneId = SCENE_30GYOSON,
         .checks = {
             CHECK(RC_GREAT_BAY_COAST_FISHERMAN_MINIGAME, RANDO_EVENTS[RE_CLEARED_GREAT_BAY_TEMPLE] && (HAS_ITEM(ITEM_HOOKSHOT) || CAN_USE_MAGIC_ARROW(ICE))),
-            CHECK(RC_GREAT_BAY_COAST_OWL_STATUE, CAN_USE_SWORD),
             CHECK(RC_GREAT_BAY_COAST_MIKAU, CAN_PLAY_SONG(HEALING)),
```
Eblo: I forgot about this one because the cutscene skip gets around it. Can you make this change, and with it, modify `SkipPushingMikau` to not run if `IS_RANDO && !CAN_USE_ABILITY(SWIM)`?
```suggestion
            CHECK(RC_GREAT_BAY_COAST_MIKAU, CAN_USE_ABILITY(SWIM) && CAN_PLAY_SONG(HEALING)),
```

mm/2s2h/Rando/Logic/Regions/South.cpp
```diff
@@ -5,17 +5,42 @@
...
+     * over the water as Deku, or he uses one of the forms tall enough to walk through the water. This may also be
+     * expanded if FD is ever logically considered.
+     */
+    if (RANDO_EVENTS[RE_CLEARED_WOODFALL_TEMPLE] &&
+        (CAN_USE_ABILITY(SWIM) || CAN_BE_DEKU || CAN_BE_ZORA || CAN_BE_GORON)) {
+        return true;
+    }
+    return false;
+}
+
```
Eblo, line 32: Sorry, I meant to for `CanGetPastBigOcto` to go before the clang format line, but not to remove it. Can you walk back the `e09482f` clang format commit and add this comment back? What I want is for `CanGetPastBigOcto` to be formatted, but not the rest of the file.
```suggestion

// clang-format off
```

Caladius: I gotcha now, fixed.

Eblo (approved):

PR #1090, opened 2025-03-25
[RANDO] Corrected Snowhead temple COMPASS_ROOM_LEDGE_CHEST logic

sitton76:
This check was previously listed as requiring Hookshot or Fire Arrows, but just requires Fire Arrows.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2812897469.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2812902085.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2812929920.zip)


Eblo (approved):

Caladius:
Is this the room where the Chest appears after you press the switch encased in Ice?

sitton76:
> Is this the room where the Chest appears after you press the switch encased in Ice?

Yes that is right, after the switch is pressed there is a block also encased in ice that Link can push, so only Fire Arrow requirements should be needed for this.

Archez (approved):

PR #1091, opened 2025-03-25
[Enhancement] Underwater Ocarina

Sirius902:
Allows Zora to use the Ocarina of Time when grounded underwater.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3599825809.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3599930285.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3600031277.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Player/UnderwaterOcarina.cpp
+++ b/mm/2s2h/Enhancements/Player/UnderwaterOcarina.cpp
@@ -1,0 +1,21 @@
+#include "public/bridge/consolevariablebridge.h"
+#include "2s2h/GameInteractor/GameInteractor.h"
+#include "2s2h/ShipInit.hpp"
+
+extern "C" {
+#include "z64item.h"
+}
+
+#define CVAR_NAME "gEnhancements.Player.UnderwaterOcarina"
+#define CVAR CVarGetInteger(CVAR_NAME, 0)
+
+void RegisterUnderwaterOcarina() {
+    COND_VB_SHOULD(VB_DISABLE_ITEM_UNDERWATER_FLOOR, CVAR, {
+        const auto item = va_arg(args, s32);
+        if (item == ITEM_OCARINA_OF_TIME) {
+            *should = false;
+        }
+    });
+}
+
+static RegisterShipInitFunc initFunc(RegisterUnderwaterOcarina, { CVAR_NAME });
--- a/mm/2s2h/GameInteractor/GameInteractor.h
+++ b/mm/2s2h/GameInteractor/GameInteractor.h
@@ -245,6 +245,7 @@ typedef enum {
     VB_BE_NEAR_DOOR,
     VB_LOAD_PLAYER_ANIMATION_FRAME,
     VB_PLAY_SCENE_SEQUENCE,
+    VB_DISABLE_ITEM_UNDERWATER_FLOOR,
 } GIVanillaBehavior;
 
 typedef enum {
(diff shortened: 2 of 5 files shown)


mm/src/code/z_parameter.c
```diff
@@ -2585,8 +2585,10 @@ void Interface_UpdateButtonsPart2(PlayState* play) {
         for (i = EQUIP_SLOT_C_LEFT; i <= EQUIP_SLOT_C_RIGHT; i++) {
             if (GET_CUR_FORM_BTN_ITEM(i) != ITEM_MASK_ZORA) {
                 if (Player_GetEnvironmentalHazard(play) == PLAYER_ENV_HAZARD_UNDERWATER_FLOOR) {
-                    if (!((GET_CUR_FORM_BTN_ITEM(i) >= ITEM_BOTTLE) &&
-                          (GET_CUR_FORM_BTN_ITEM(i) <= ITEM_OBABA_DRINK))) {
+                    if (!GameInteractor_Should(VB_ENABLE_ITEM_UNDERWATER_FLOOR,
```
Archez: This `_Should` should read like whats being performed ("if should X"). Currently it is inverted.
Specifically, the `!` negation should be moved back to the condition value being passed in to retain the vanilla condition, and then the enum can be written like `VB_DISABLE_ITEM_UNDERWATER_FLOOR`.

mm/2s2h/Enhancements/Player/UnderwaterOcarina.cpp
```diff
@@ -0,0 +1,21 @@
...
+extern "C" {
+#include "z64item.h"
+}
+
+#define CVAR_NAME "gEnhancements.Player.UnderwaterOcarina"
+#define CVAR CVarGetInteger(CVAR_NAME, 0)
+
+void RegisterUnderwaterOcarina() {
+    COND_VB_SHOULD(VB_ENABLE_ITEM_UNDERWATER_FLOOR, CVAR, {
+        const auto item = va_arg(args, std::int32_t);
```
Archez: In the code base we prefer using the typedef version of width types as defined by libultratypes, not the ones coming in the standard namespace.
```suggestion
        const auto item = va_arg(args, int32_t);
```

mm/2s2h/Enhancements/Player/UnderwaterOcarina.cpp
```diff
@@ -0,0 +1,21 @@
+#include <libultraship/bridge.h>
```
Archez: Recently we stopped trying to include the `libultraship/bridge.h` header as it basically includes almost everything from LUS, increasing compile times for files that don't need it.
This can be replaced with 
```suggestion
#include "public/bridge/consolevariablebridge.h"
```

mm/2s2h/GameInteractor/GameInteractor.h
```diff
@@ -245,6 +245,8 @@ typedef enum {
     VB_BE_NEAR_DOOR,
     VB_LOAD_PLAYER_ANIMATION_FRAME,
     VB_PLAY_SCENE_SEQUENCE,
+    // Vanilla Condition: !((GET_CUR_FORM_BTN_ITEM(i) >= ITEM_BOTTLE) && (GET_CUR_FORM_BTN_ITEM(i) <= ITEM_OBABA_DRINK))
```
Eblo: Since the two uses of this new VB are slightly different (dpad vs normal button), let's go ahead and remove this comment. Developers will see the vanilla conditions when they look at where this is used.

Eblo (changes requested):
I left only the tiniest nitpick regarding a comment, but this is nice! A welcome addition for players with MM3D muscle memory.

Eblo (approved):

balloondude2 (approved):

PR #1094, opened 2025-03-26
[Rando] Halt moving objects when obtaining items

Eblo:
Fixes an issue with rando where moving objects do not halt while the player is obtaining an item. This results in the Secret Shrine and Deku Butler race freestanding items being particularly annoying to get, as the player may have to make multiple attempts to get everything. These areas were not designed for the player to enter into a Get Item animation, so they do not halt. This change simply halts the updates for the relevant actors if the player is in the Get Item state for randomizers.

I put this together in one actor behavior file since it's mostly the same logic across different actors. Plus I suspect more things may need added in the future.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2847488174.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2847488750.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2847509846.zip)


Eblo:
I noticed the Deku Butler race had the same issue as beans, making it frustrating to collect freestanding items. I expanded this PR to handle the Deku Butler, his doors, and other moving platforms.

PR #1095, opened 2025-03-26
Fix MSVC flags and remove RTC option

Archez:
It looks like MSVC sets `/RTC1` by default as a DEBUG flag, but in some places we explicitly set `/O2` in DEBUG, which these two options are not compatible. Since before #1082 we were blanking the flags, this issue only recently cropped up.

This PR adds a string replace to remove any RTC options from the DEBUG flags, and also removes the previous `-g` debug flag as that was incompatible with MSVC, but is a default for GCC.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2826381453.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2826386916.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2826566867.zip)


Eblo:
I can confirm I can build this locally via MSVC when I could not prior.

PR #1097, opened 2025-03-28
Fix torch lighting bug with one point skips

### LINKED ISSUE #1016: Torches unable to be re-lit after burning out with one point skips
Unsure if this applies to all torches, but with One Point skips enabled, if you light the torch in Woodfall maze room, then let it burn out, it will no longer be light-able.

Archez:
With onepoint skips enabled, torches where getting stuck in a update loop with playing the cutscene over and over (which got skipped over and over), we were already setting the snuff timer value, but not the pendingAction value to get the torch out of the cutscene state.

This applies the pendingAction fix, and also replicated the switch flag set in the skip as now that wont execute in vanilla.

Fixes #1016

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2842627540.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2842630670.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2842658262.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Cutscenes/SkipOnePointCutscenes.cpp
+++ b/mm/2s2h/Enhancements/Cutscenes/SkipOnePointCutscenes.cpp
@@ -38,6 +38,14 @@ void RegisterSkipOnePointCutscenes() {
                 }
 
                 ObjSyokudai* torch = (ObjSyokudai*)actor;
+                s32 switchFlag = OBJ_SYOKUDAI_GET_SWITCH_FLAG(actor);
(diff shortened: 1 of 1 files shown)


Archez:
Apparently this also fixes the Zora hall torch lighting softlock, so I've removed that edge case now.
Sweet!

PR #1111, opened 2025-04-07
[Enhancement] Faster scene transitions

rachaellama:
Speeds up the fade in/out when moving between areas on the map.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2924187475.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2924208802.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2924224283.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -686,9 +686,6 @@ void BenMenu::AddEnhancements() {
     AddWidget(path, "Manual Jump", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.Player.ManualJump")
         .Options(CheckboxOptions().Tooltip("Z + A to Jump and B while midair to Jump Attack."));
-    AddWidget(path, "Faster Scene Transitions", WIDGET_CVAR_CHECKBOX)
-        .CVar("gEnhancements.Player.FasterSceneTransitions")
-        .Options(CheckboxOptions().Tooltip("Fade in and out more quickly when moving between areas."));
     AddWidget(path, "Dpad Equips", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.Dpad.DpadEquips")
         .Options(CheckboxOptions().Tooltip("Allows you to equip items to your D-pad."));
@@ -1163,6 +1160,9 @@ void BenMenu::AddEnhancements() {
     AddWidget(path, "Fast Dampe Flame Digging", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.Timesavers.DampeDiggingSkip")
         .Options(CheckboxOptions().Tooltip("Only requires digging up one flame to spawn the big poe."));
+    AddWidget(path, "Faster Scene Transitions", WIDGET_CVAR_CHECKBOX)
+        .CVar("gEnhancements.Timesavers.FasterSceneTransitions")
+        .Options(CheckboxOptions().Tooltip("Fade in and out more quickly when moving between areas."));
 
     // Fixes
     path = { "Enhancements", "Fixes", 1 };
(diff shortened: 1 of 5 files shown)


mm/2s2h/Enhancements/Player/FasterSceneTransitions.cpp
```diff
@@ -0,0 +1,54 @@
+#include <libultraship/bridge.h>
```
Archez: We recently moved away from importing the LUS bridge header as its a vary large header and increases compile times. This can be swapped for -- to include the minimum for what you are using
```suggestion
#include "public/bridge/consolevariablebridge.h"
```

mm/2s2h/BenGui/BenMenu.cpp
```diff
@@ -686,6 +686,9 @@ void BenMenu::AddEnhancements() {
     AddWidget(path, "Manual Jump", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.Player.ManualJump")
         .Options(CheckboxOptions().Tooltip("Z + A to Jump and B while midair to Jump Attack."));
+    AddWidget(path, "Faster Scene Transitions", WIDGET_CVAR_CHECKBOX)
+        .CVar("gEnhancements.Player.FasterSceneTransitions")
```
Archez: I think this toggle position/cvar is probably better off in the TimeSavers menu, as its not directly related to the _player_. (This would also mean moving the .cpp file as well).

mm/2s2h/Enhancements/Player/FasterSceneTransitions.cpp
```diff
@@ -0,0 +1,54 @@
...
+            transitionType = gSaveContext.save.isNight ? TRANS_TYPE_FADE_BLACK_FAST : TRANS_TYPE_FADE_WHITE_FAST;
+            break;
+    }
+
+    gPlayState->transitionType = transitionType;
+}
+
+void RegisterFastExitFade() {
+    COND_VB_SHOULD(VB_SETUP_TRANSITION, CVAR, {
+        if (gPlayState->sceneId != SCENE_SPOT00) {
```
Archez: I noticed the fast transition when entering a grotto causes the player to receive half a heart of fall damage. Not sure why exactly, but presumably the player state is changed earlier while falling causing the damage trigger.

Can probably check for the grottos scene ID of `SCENE_KAKUSIANA` to avoid setting the fast transition (since the player still has to wait for control haven't falling down).

Archez (commented):
Hey, awesome first PR! Thanks for adding these changes following our existing patterns 🙂 

I've left some feedback below after trying this out. I haven't fully exhausted all the different entrances though.

mm/2s2h/BenGui/BenMenu.cpp
rachaellama: Done. I put this under "other" timesavers because I'm assuming the fade in/out isn't really considered a cutscene.

mm/2s2h/Enhancements/Player/FasterSceneTransitions.cpp
rachaellama: Done. Checking for `SCENE_KAKUSIANA` seems to fix it.

Archez: I found one more that is impacted by fall damage for beneath the graveyard `Z2_HAKASHITA`.
Other ones do not seem to cause damage (deku playground, dampe's grave, moon trails)

rachaellama: Okay, I've got that one now as well.

Archez: Alright last suggestion from me, now that there are a couple scene exclusions here, it would be great to have a comment in the code capturing why they are exclude for future reference 🙂

rachaellama: Done.

rachaellama: Done.

Archez (approved):
Thanks for getting the feedback resolved 🙂

PR #1115, opened 2025-04-09
[Rando] Add Frog Shuffle

balloondude2:
This adds an option to shuffle the 4 frogs that need to be collected throughout Termina for the Frog Choir. When enabled, the frogs are replaced with an item that can be picked up, while the frogs will be found at random checks. The frogs will stay with the player over cycle resets, like Stray Fairies. 

Huge thanks to @Caladius for helping, especially with Item Draw stuff. 

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3771633027.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3771644991.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3771729477.zip)


mm/2s2h/Rando/DrawItem.cpp
```diff
@@ -14,8 +14,12 @@ extern "C" {
 #include "objects/object_gi_liquid/object_gi_liquid.h"
 #include "objects/object_sek/object_sek.h"
 #include "objects/object_st/object_st.h"
+/* Minifrog */ #include "objects/object_fr/object_fr.h"
```
Eblo: Do we need this comment?

mm/2s2h/Rando/DrawItem.cpp
```diff
@@ -14,8 +14,12 @@ extern "C" {
...
+#include "overlays/actors/ovl_En_Minifrog/z_en_minifrog.h"
 
-Gfx* ResourceMgr_LoadGfxByName(const char* path);
+    Gfx*
+    ResourceMgr_LoadGfxByName(const char* path);
```
Eblo: Was this reformatting a clang-formatism, or intentional?

mm/2s2h/Rando/DrawItem.cpp
```diff
@@ -296,6 +300,69 @@ void DrawSkulltulaToken(RandoItemId randoItemId, Actor* actor) {
...
+
+        Matrix_ReplaceRotation(&play->billboardMtxF);
+        gSPMatrix(POLY_OPA_DISP++, Matrix_NewMtx(play->state.gfxCtx), G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
+        gSPDisplayList(POLY_OPA_DISP++, *dList);
+
+        CLOSE_DISPS(play->state.gfxCtx);
+    }
+}
+
+void DrawMinifrog(RandoItemId randoItemId, Actor* actor) {
```
Eblo: In general, I wonder if this is the best place for these draw funcs. We seem to have things split between  `DrawFuncs.cpp` and `DrawItem.cpp`. The former so far is just for boss souls, while the latter has everything else. What are your current thoughts?

Eblo (commented):
Left some questions. Builds are expired, so I can't test as-is, though I would rather test against an up-to-date build anyway.

mm/2s2h/Rando/DrawItem.cpp
balloondude2: Probably not necessary at the moment. I think it's there to match how Cal was documenting the includes for Enemy Souls.

mm/2s2h/Rando/DrawItem.cpp
balloondude2: Not intentional. I think it was a clang format thing, but it's been a while, so I'm not sure.

mm/2s2h/Rando/DrawItem.cpp
balloondude2: It probably would make more sense in DrawFuncs. I can't remember for sure, but I think I did try to have it there, but was running into issues. I'll check again.

balloondude2:
Updated, so this should be ready for testing and review again.

mm/2s2h/Rando/StaticData/Checks.cpp
```diff
@@ -1371,6 +1371,11 @@ std::map<RandoCheckId, RandoStaticCheck> Checks = {
...
 
+    // Frogs //
+    RC(RC_CLOCK_TOWN_LAUNDRY_FROG,                                       RCTYPE_FROG,             SCENE_ALLEY,                    FLAG_NONE,                   0x00,                                                                RI_FROG_WHITE),
+    RC(RC_GREAT_BAY_TEMPLE_GEKKO_FROG,                                   RCTYPE_FROG,             SCENE_SEA,                      FLAG_NONE,                   0x00,                                                                RI_FROG_PINK),
+    RC(RC_SOUTHERN_SWAMP_FROG,                                           RCTYPE_FROG,             SCENE_20SICHITAI,               FLAG_NONE,                   0x00,                                                                RI_FROG_BLUE),
+    RC(RC_WOODFALL_TEMPLE_GEKKO_FROG,                                    RCTYPE_FROG,             SCENE_MITURIN,                  FLAG_NONE,                   0x00,                                                                RI_FROG_CYAN),
```
Eblo, line 1378: Should these be `FLAG_NONE` when each frog is synonymous with a flag?

Eblo (commented):
Now that I've playtested, I'm curious what the rationale is for having the frog checks be collectible pickups. Normally these checks are gated by Don Gero's Mask. I was thinking that this would work by talking to the frogs, then they vanish and the flag change triggers the item.

mm/2s2h/Rando/StaticData/Checks.cpp
balloondude2: We want the flags tied to the RIs instead of the RCs in this case. For example: If we did put the flags on the RCs, finding RI_FROG_WHITE would set the flag for finding that frog, which would then trigger RC_CLOCK_TOWN_LAUNDRY_FROG, thus granting another check.

balloondude2:
Thanks @Eblo. The new frog checks now also require Don Gero's mask to obtain.

mm/2s2h/Rando/ActorBehavior/EnMinifrog.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include "overlays/actors/ovl_En_Minifrog/z_en_minifrog.h"
+void EnMinifrog_TurnToPlayer(EnMinifrog* enMinifrog);
+void EnMinifrog_Jump(EnMinifrog* enMinifrog);
+void EnMinifrog_JumpTimer(EnMinifrog* enMinifrog);
+void EnMinifrog_SpawnDust(EnMinifrog* enMinifrog, PlayState* play);
+}
+
+#define SHUFFLED_FROGS (IS_RANDO && RANDO_SAVE_OPTIONS[RO_SHUFFLE_FROGS])
+
+static u16 sIsFrogReturnedFlags[] = {
```
Eblo, line 16: I would suggest externing this, but the source already defines `sIsFrogReturnedFlags` in multiple different places, and those definitions are not all identical. What the heck.

mm/2s2h/Rando/ActorBehavior/EnMinifrog.cpp
```diff
@@ -0,0 +1,103 @@
...
+            return;
+        }
+        *should = RANDO_SAVE_CHECKS[frogCheck].cycleObtained;
+    });
+
+    COND_ID_HOOK(OnActorInit, ACTOR_EN_MINIFROG, SHUFFLED_FROGS, [](Actor* actor) {
+        EnMinifrog* enMinifrog = (EnMinifrog*)actor;
+        // If this is not a Mountain Village frog, use the custom actionFunc
+        if (enMinifrog->frogIndex != FROG_YELLOW && !EN_FROG_IS_RETURNED(actor)) {
+            enMinifrog->actor.textId = CHECK_WEEKEVENTREG(sIsFrogReturnedFlags[enMinifrog->frogIndex]) ? 0xD82 : 0xD81;
```
Eblo: Since this is handled on init, this creates a possible scenario where:

1. Frog is initialized with dialog for not having been found
2. Player finds that check within the same scene
3. Player then talks to that frog and gets the wrong dialog

Try leaving this as the single text id, then put `CHECK_WEEKEVENTREG(sIsFrogReturnedFlags[enMinifrog->frogIndex])` in the OnOpenText hook to determine which string to use on demand. See the OnOpenText hook for `0x72F` as an example of getting the actor of interest.

Eblo (changes requested):

mm/2s2h/Rando/ActorBehavior/EnMinifrog.cpp
balloondude2: Got around to updating this

Eblo (approved):

PR #1117, opened 2025-04-10
[Enhancement] Auto Bombers' code

rachaellama:
Automatically fill in the Bombers' code once you've got the notebook.

It doesn't look like this can be done with the GameInteractor if you want to actually update the message box.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2959179125.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2959179655.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2959235460.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Dialogue/AutoBombersCode.cpp
+++ b/mm/2s2h/Enhancements/Dialogue/AutoBombersCode.cpp
@@ -1,0 +1,25 @@
+#include "public/bridge/consolevariablebridge.h"
+#include "2s2h/GameInteractor/GameInteractor.h"
+#include "2s2h/ShipInit.hpp"
+
+extern "C" {
+#include "variables.h"
+#include "functions.h"
+}
+
+#define CVAR_NAME "gEnhancements.Dialogue.AutoBombersCode"
+#define CVAR CVarGetInteger(CVAR_NAME, 0)
+
+void RegisterAutoBombersCode() {
+    COND_ID_HOOK(OnActorUpdate, ACTOR_EN_BOMBERS2, CVAR, [](Actor* actor) {
+        if (gPlayState->msgCtx.currentTextId == 0x0729 && CHECK_QUEST_ITEM(QUEST_BOMBERS_NOTEBOOK)) {
+            for (int i = 0; i < 5; i++) {
+                u8 digit = gSaveContext.save.saveInfo.bomberCode[i];
+                gPlayState->msgCtx.unk12054[i] = digit;
+                Font_LoadCharNES(gPlayState, '0' + digit, gPlayState->msgCtx.unk120C4 + (i << 7));
+            }
+        }
+    });
+}
+
+static RegisterShipInitFunc initFunc(RegisterAutoBombersCode, { CVAR_NAME });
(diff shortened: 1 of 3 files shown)


Caladius:
Try taking a look at what i did for the Banker dialogue.

https://github.com/HarbourMasters/2ship2harkinian/pull/591/files#diff-e29fd14d5f40bdcb43762427e8a75236d00f83987b6ee34ddcf7c3882724d168R10

I think the Font_LoadCharNES is what updates the dialogue box.

This is awesome btw.

rachaellama:
I rewrote this to use `ACTOR_EN_BOMBERS2` and `Font_LoadCharNES()`, which is probably better. Thanks for the feedback!

mm/2s2h/Enhancements/Dialogue/AutoBombersCode.cpp
```diff
@@ -0,0 +1,25 @@
+#include "public/bridge/consolevariablebridge.h"
+#include "2s2h/GameInteractor/GameInteractor.h"
+#include "2s2h/ShipInit.hpp"
+#include "global.h"
```
Archez: We don't want to have to include the global header unless absolutely necessary. For this files including `variables.h` should be sufficient. Also we try to keep the includes for original game src within `extern "C" {}` blocks. You can add it with the forward declare below.

mm/2s2h/Enhancements/Dialogue/AutoBombersCode.cpp
```diff
@@ -0,0 +1,25 @@
...
+#define CVAR_NAME "gEnhancements.Dialogue.AutoBombersCode"
+#define CVAR CVarGetInteger(CVAR_NAME, 0)
+
+extern "C" {
+void Font_LoadCharNES(PlayState* play, u8 codePointIndex, s32 offset);
+}
+
+void RegisterAutoBombersCode() {
+    COND_ID_HOOK(OnActorUpdate, ACTOR_EN_BOMBERS2, CVAR, [](Actor* actor) {
+        if (gPlayState->msgCtx.currentTextId == 0x729 && CHECK_QUEST_ITEM(QUEST_BOMBERS_NOTEBOOK)) {
```
Archez: Text IDs are 4 bytes wide, and the game source puts leading zeros to have them all written as 4 bytes. Lets do the same for consistency.
```suggestion
        if (gPlayState->msgCtx.currentTextId == 0x0729 && CHECK_QUEST_ITEM(QUEST_BOMBERS_NOTEBOOK)) {
```

Archez (commented):
Great job updating this to an isolated solution! Just have minor feedback below.

rachaellama:
I added these suggestions. `Font_LoadCharNES()` is already in functions.h, so I just included that. Let me know if you'd rather not do it that way, but that seems like a common pattern.

Archez:
> Font_LoadCharNES() is already in functions.h, so I just included that

That's fine yea.

Archez (approved):

Caladius:
Quality of life things like this that you don't even think about are what make this port incredible.

Nice work!

PR #1119, opened 2025-04-12
[Rando] Changed switch logic for GBT.

sitton76:
When I initially wrote the switch logic for GBT I made a mistake where I attempted to account for the active switch state of the waterflow, I realize now that was a mistake and created a exception to how the check tracker works.(Where access to items could be enabled or disabled depending on the current waterflow.)

With these changes now the logic only cares about what the player can reach. I also happen to spot a small logic issue with a Stray Fairy that I corrected.

At Archez's suggestion I kept the macros in but just changed the definitions for the sake of check tracker clarity, however I did remove the EITHER_FLOW macro since we no longer check for the current flow actively.

Testing these changes in game seems to have the desired behavior.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2933985824.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2933993556.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2933996047.zip)


mm/2s2h/Rando/Logic/Regions/GreatBayTemple.cpp
```diff
@@ -324,21 +324,16 @@ static RegisterShipInitFunc initFunc([]() {
...
-            CHECK(RC_GREAT_BAY_TEMPLE_WATER_WHEEL_FREESTANDING_RUPEE_03, GBT_EITHER_FLOW),
-            CHECK(RC_GREAT_BAY_TEMPLE_WATER_WHEEL_FREESTANDING_RUPEE_04, GBT_EITHER_FLOW),
-            CHECK(RC_GREAT_BAY_TEMPLE_WATER_WHEEL_FREESTANDING_RUPEE_05, GBT_EITHER_FLOW),
-
+            CHECK(RC_GREAT_BAY_TEMPLE_SF_WATER_WHEEL_SKULLTULA,          CAN_BE_ZORA || (HAS_ITEM(ITEM_BOW) && HAS_ITEM(ITEM_MASK_GREAT_FAIRY))),
+            CHECK(RC_GREAT_BAY_TEMPLE_WATER_WHEEL_FREESTANDING_RUPEE_01, CAN_BE_ZORA),
+            CHECK(RC_GREAT_BAY_TEMPLE_WATER_WHEEL_FREESTANDING_RUPEE_02, CAN_BE_ZORA),
+            CHECK(RC_GREAT_BAY_TEMPLE_WATER_WHEEL_FREESTANDING_RUPEE_03, CAN_BE_ZORA),
+            CHECK(RC_GREAT_BAY_TEMPLE_WATER_WHEEL_FREESTANDING_RUPEE_04, CAN_BE_ZORA),
+            CHECK(RC_GREAT_BAY_TEMPLE_WATER_WHEEL_FREESTANDING_RUPEE_05, CAN_BE_ZORA),
```
balloondude2: I think this group of checks could just be `true`. You can reach the platform with the skulltula and freestanding rupees without Zora. Additionally, if you kill the skulltula with your sword, you do not need the Great Fairy Mask to collect the drop.

mm/2s2h/Rando/Logic/Logic.h
```diff
@@ -69,10 +69,10 @@ extern std::unordered_map<RandoRegionId, RandoRegion> Regions;
...
 #define CAN_PLAY_SONG(song) (HAS_ITEM(ITEM_OCARINA_OF_TIME) && CHECK_QUEST_ITEM(QUEST_SONG_##song))
 #define CAN_RIDE_EPONA (CAN_PLAY_SONG(EPONA))
-#define GBT_REGULAR_WATER_FLOW (!RANDO_EVENTS[RE_GREAT_BAY_YELLOW_TOGGLE] && !RANDO_EVENTS[RE_GREAT_BAY_RED_TOGGLE])
-#define GBT_REVERSE_WATER_FLOW (RANDO_EVENTS[RE_GREAT_BAY_YELLOW_TOGGLE] && RANDO_EVENTS[RE_GREAT_BAY_RED_TOGGLE])
-#define GBT_EITHER_FLOW (!RANDO_EVENTS[RE_GREAT_BAY_YELLOW_TOGGLE] || RANDO_EVENTS[RE_GREAT_BAY_RED_TOGGLE])
-#define GBT_RED_SWITCH_FLOW (RANDO_EVENTS[RE_GREAT_BAY_RED_SWITCH_1] && RANDO_EVENTS[RE_GREAT_BAY_RED_SWITCH_2])
+#define GBT_REGULAR_WATER_FLOW (CAN_BE_ZORA) // Keeping for the sake of check tracker clarity
+#define GBT_REVERSE_WATER_FLOW                                                             \
+    (RANDO_EVENTS[RE_GREAT_BAY_RED_SWITCH_1] && RANDO_EVENTS[RE_GREAT_BAY_RED_SWITCH_2] && \
+     HAS_ITEM(ITEM_HOOKSHOT)) // Keeping for the sake of check tracker clarity
```
balloondude2: These macros are currently only used in Connections, so they don't actually show up on the check tracker. I'm not sure if there is much merit in keeping `GBT_REGULAR_WATER_FLOW` as that's the default state of the dungeon, so long as the appropriate Connections have `CAN_BE_ZORA` (2 of the 3 connections with `GBT_REGULAR_WATER_FLOW` already have that condition explicitly). 

Also, I'd personally prefer if `GBT_REVERSE_WATER_FLOW` was renamed to something like `GBT_CAN_REVERSE_WATER_FLOW`.

balloondude2 (changes requested):
Thanks for getting this cleaned up. It's nice seeing the GBT checks actually start showing as accessible on the tracker.

mm/2s2h/Rando/Logic/Regions/GreatBayTemple.cpp
sitton76: AH right, was going off memory thinking the switch underwater started the waterwheel when it in fact just enables the sprout, ill correct this inna bit

mm/2s2h/Rando/Logic/Logic.h
sitton76: Def can rename them to include the CAN.

I can prob just remove the regular flow and keep the reverse one.

mm/2s2h/Rando/Logic/Regions/GreatBayTemple.cpp
sitton76: Made the necessary corrections

mm/2s2h/Rando/Logic/Logic.h
sitton76: Okay removed the REGULAR_WATER_FLOW macro and renamed REVERSE_WATER_FLOW.

balloondude2 (approved):

PR #1127, opened 2025-04-18
Speed up Ikana block

Eblo:
The fast push/pull enhancement didn't account for `ACTOR_BG_IKANA_BLOCK`. This adds it to the mix, and also throws in a one point skip since the actor normally has a tiny cutscene when the puzzle is solved.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2972521235.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2972521628.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2972533802.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Player/FasterPushAndPull.cpp
+++ b/mm/2s2h/Enhancements/Player/FasterPushAndPull.cpp
@@ -25,24 +25,18 @@ void RegisterFasterPushAndPull() {
     });
 
     COND_VB_SHOULD(VB_PUSH_BLOCK_SET_TIMER, CVAR, {
-        ObjOshihiki* objOshihiki = va_arg(args, ObjOshihiki*);
-        objOshihiki->timer = 2;
+        Actor* actor = va_arg(args, Actor*);
+        if (actor->id == ACTOR_OBJ_OSHIHIKI) {
+            ((ObjOshihiki*)actor)->timer = 2;
+        } else if (actor->id == ACTOR_BG_IKANA_BLOCK) {
+            ((BgIkanaBlock*)actor)->unk_17B = 11;
+        }
         *should = false;
     });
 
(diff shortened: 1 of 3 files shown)


mm/2s2h/Enhancements/Player/FasterPushAndPull.cpp
```diff
@@ -31,7 +32,17 @@ void RegisterFasterPushAndPull() {
 
     COND_VB_SHOULD(VB_SKATE_BLOCK_BEGIN_MOVE, CVAR, { *should = true; });
 
-    COND_VB_SHOULD(VB_BLOCK_BEGIN_MOVE, CVAR, { *should = true; });
+    COND_VB_SHOULD(VB_BLOCK_BEGIN_MOVE, CVAR, {
+        Actor* block = va_arg(args, Actor*);
+        // Cannot always allow the Ikana block to be moved, or it will move more than one space
+        if (block->id == ACTOR_BG_IKANA_BLOCK) {
+            if (((BgIkanaBlock*)block)->unk_17B > 0) {
```
Archez: Should this kind of check be applied to the other block types (which vanilla parts check greater than a value)? Curious why this might only be an issue for the ikana one. I feel like I recall people showing the woodfall block floating over the water, but not sure if thats related.

If not, (just a nit pick) but these two if conditions can be combined together.

Eblo: `unk_17B` is the Ikana block actor's push timer value, and because of that, I think I used the wrong hook for this purpose. Another push block actor, `ObjOshihiki`, uses the `VB_PUSH_BLOCK_SET_TIMER` hook for handling its timer for pushing. I changed the Ikana block actor to match that.

PR #1152, opened 2025-05-13
[Rando] Separate CAN_PLAY_SONG from MeetsMoonRequirements

### LINKED ISSUE #1150: Hard Requiring Ocarina For Moon Access?
I'm not sure if there's a reason as to why, but currently we hard require the ocarina to have access to the moon.

I spoke about adding bottle OI to allow for this, but Blue mentioned they think that it's not supposed to hard require it, but currently does.

Could this be changed/adjusted?

Recreate: 
1). Have Oath and NOT all required remains for Majoras (Eg. 2 Remains but 4 required for Majora)
2). Enter Clown Tower Roof and play Oath
3). Watch cutscene whether you have the skip enabled or disabled 🤪 
4). Return to the roof instead of the moon.

Eblo:
The Boss Remains requirement setting has `CAN_PLAY_SONG` coupled with the condition, which inadvertently prevents going to the moon if the player used Ocarina Items in lieu of the actual Ocarina. This PR moves the `CAN_PLAY_SONG` condition from the shared boolean to the logic file so that only logic cares about the Ocarina, not the warp itself.

Fixes #1150

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3111191819.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3111221948.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3111237612.zip)

### DIFF
--- a/mm/2s2h/Rando/Logic/Logic.h
+++ b/mm/2s2h/Rando/Logic/Logic.h
@@ -180,7 +180,7 @@ inline uint32_t RemainsCount() {
 }
 
 inline bool MeetsMoonRequirements() {
-    return CAN_PLAY_SONG(OATH) && RemainsCount() >= RANDO_SAVE_OPTIONS[RO_ACCESS_MOON_REMAINS_COUNT];
+    return RemainsCount() >= RANDO_SAVE_OPTIONS[RO_ACCESS_MOON_REMAINS_COUNT];
 }
 
 inline bool CanKillEnemy(ActorId EnemyId) {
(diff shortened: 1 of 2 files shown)


balloondude2 (approved):
LGTM. Were you able to test that OI worked with this change?

Eblo:
I found an easier way to test by pulling out the Ocarina as normal, then removing it with the save editor before playing Oath.

Issue #410, opened 2024-05-17
Known Issues: Audio


garrettjoecox:
## Sound effects playing with wrong pitch/modulation
- Deku Link jumping/falling in water
- Goron Link rolling sound
- Damaging Cursed Man in Swamp Spider House
- Chuchu respawning/hopping sounds
- Hiploop (in woodfall swamp area) roaring sounds

## Sound effects not playing
- Skull kid not playing `NA_SE_EN_STALKIDS_SURPRISED` in intro sequence
- When link is grabbed by other actors, he will lose his voice for a period of time.
- Minions are silent while in learning Elegy of Emptiness sequence

## Music
- After obtaining Zora mask, music for Great Bay does not resume

MarthurSR:
The issue with the music not resuming after obtaining Zora mask occurs in other instances as well. The Termina music won't resume after obtaining the stone mask, and I believe I've experienced this in a few other areas and forgot to report it.

shipmaster64:
Is this a complicated thing to tackle? Just wondering if it's planned to be addressed in the next release, if possible

ProNerdy:
I've had similar issues with the dungeon music not resuming in Wood fall and Stone Tower. In Stone Tower after getting Light Arrows and leaving the room, thd music did not resume. I had to use Song of soaring and warp to thd beginning of the dungeon to resume it

JordanLongstaff:
How many of these issues are still present? Shouldn't this be tracked as a checklist, or a list of sub-issues like #407?

VirtuaTyKing:
Minions are silent while in learning Elegy of Emptiness sequence.
Still happens.

JordanLongstaff:
> Skull kid not playing `NA_SE_EN_STALKIDS_SURPRISED` in intro sequence

This one seems to be fixed. It played on cue for me.

PR #1156, opened 2025-05-18
Fix issues with sound after Link gets grabbed

Eblo:
Same problem, same solution as SoH: https://github.com/HarbourMasters/Shipwright/pull/5506. Thanks to @briaguya-ai, @Malkierian, and @garrettjoecox for combing through SoH commits and reasoning out the issue.

 `osAiSetFrequency` is used to calculate the length of a sound effect:

https://github.com/HarbourMasters/2ship2harkinian/blob/bde6fcfc15c0a7f737d48f2b6cee9bcc54ed49aa/mm/src/audio/lib/heap.c#L948
https://github.com/HarbourMasters/2ship2harkinian/blob/bde6fcfc15c0a7f737d48f2b6cee9bcc54ed49aa/mm/src/audio/lib/heap.c#L987
https://github.com/HarbourMasters/2ship2harkinian/blob/bde6fcfc15c0a7f737d48f2b6cee9bcc54ed49aa/mm/src/audio/lib/seqplayer.c#L994
https://github.com/HarbourMasters/2ship2harkinian/blob/bde6fcfc15c0a7f737d48f2b6cee9bcc54ed49aa/mm/src/audio/lib/seqplayer.c#L1003
https://github.com/HarbourMasters/2ship2harkinian/blob/bde6fcfc15c0a7f737d48f2b6cee9bcc54ed49aa/mm/src/audio/lib/seqplayer.c#L1010-L1016

Which ultimately seems to result in 2ship thinking the sound effect is longer than it is, thus thinking that sounds are still playing after they are done, and thus not flushing those SFX when they should be. With higher priority sounds such as Link being grabbed, this results in lower priority sounds not being able to play.

The precalculated math by briaguya-ai to correct this is in the comment of this change. This assumes that 2ship's sampling rate is 32kHz. The non-stubbed version comes from https://github.com/HarbourMasters/2ship2harkinian/blob/develop/mm/src/libultra/io/aisetfreq.c

Fixes part of #410. The grabbing issue was known, though this uncovered that the ledge grabbing issue from SoH persists in 2ship and is fixed with this as well.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3146918870.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3146920231.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3146937937.zip)

### DIFF
--- a/mm/src/code/stubs.c
+++ b/mm/src/code/stubs.c
@@ -264,7 +264,42 @@ void osMapTLBRdb(void) {
 u32 __osProbeTLB(void* param_1) {
 }
 s32 osAiSetFrequency(u32 frequency) {
-    return 1;
+    // this is based off the math from the original method
+    /*
+    s32 osAiSetFrequency(u32 frequency) {
+        u8 bitrate;
+        f32 dacRateF = ((f32)osViClock / frequency) + 0.5f;
+        u32 dacRate = dacRateF;
+        if (dacRate < 132) {
+            return -1;
+        }
+        bitrate = (dacRate / 66);
+        if (bitrate > 16) {
+            bitrate = 16;
+        }
+        HW_REG(AI_DACRATE_REG, u32) = dacRate - 1;
+        HW_REG(AI_BITRATE_REG, u32) = bitrate - 1;
+        return osViClock / (s32)dacRate;
+    }
+    */
+
+    // bitrate is unused
+
+    // osViClock comes from
+    // #define VI_NTSC_CLOCK 48681812 /* Hz = 48.681812 MHz */
+    // s32 osViClock = VI_NTSC_CLOCK;
+
+    // frequency was originally 32000
+
+    // given all of that, dacRate is
+    // (u32)(((f32)48681812 / 32000) + 0.5f)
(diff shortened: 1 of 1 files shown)


briaguya0:
i'd like to add that SoH was using a 44.1k sample rate when i made the SoH PR, but using 44.1 for `freq` in the calculation led to SFX getting cut off early again (https://github.com/HarbourMasters/Shipwright/issues/2354). using 32k (the sample rate used on console) for `freq` works as expected even when using a 44.1k sample rate for playback.

Malkierian:
Yeah, the current audio output frequency seemed to have no effect on what this value should be, it seems to work for both 44.1khz and 32khz, so I'd guess that there's some other hardcoding somewhere that makes 44.1 being returned from this (which it should have been with the math on what would be passed in) too much.

All the more reason to restore 32khz output on SoH ASAP.

PR #1157, opened 2025-05-18
Fix inaccessible playstate check

garrettjoecox:


inspectredc (approved):

Eblo:
Can confirm from some quick and dirty playtesting that this fixes the issue.

PR #1158, opened 2025-05-18
[Enhancement] Hidden grottos visibility marker

Eblo:
Adds the MM equivalent to SoH's Shard of Agony feature. The visual effect and "Wear Mask of Truth" option are inspired by MM3D.

- **Off**: No visual marker, vanilla behavior
- **Wear Mask of Truth**: Hidden grottos are marked when the Mask of Truth is equipped. This is how it works in MM3D.
- **Have Mask of Truth**: Hidden grottos are marked once the Mask of Truth is obtained, but it does not need to be worn. For convenience's sake, making it similar to the Stone of Agony in SoH.
- **Always**: Hidden grottos are always marked, regardless of the Mask of Truth.

image

The enhancement accounts for Song of Storms hidden grottos just because they are defined in the code, but I don't think any actually exist in MM.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3147813322.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3147813965.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3147825008.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1295,6 +1295,18 @@ void BenMenu::AddEnhancements() {
                          "- No trade: Gibdos will vanish without taking items.")
                 .DefaultIndex(GibdoTradeSequenceOptions::GIBDO_TRADE_SEQUENCE_VANILLA)
                 .ComboMap(gibdoTradeSequenceOptions));
+    AddWidget(path, "Hidden Grottos Visibility", WIDGET_CVAR_COMBOBOX)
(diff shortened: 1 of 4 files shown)


garrettjoecox (approved):
Code looks good, probably just one want other person to QA it real quick then merge

balloondude2 (approved):
LGTM. Did some QA, and everything worked as expected.

PR #1159, opened 2025-05-18
Add vector support to comboboxes

louist103:
Adds support for `std::vector` in the menu combo boxes. I had to use map and vector pointer because there are a few places where empty `ComboboxOptions` are passed in and that is not allowed with a reference.
This is part of a series of changes I want to make involving the menu and CVars.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3225365180.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3225365418.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3225391040.zip)


garrettjoecox (approved):
No qualms here but I'll defer to malk's approval as I think they had some thoughts about this and it's integration with the modern menu

Malkierian (approved):
Well, louis and I had a discussion about union vs variant, and one union among all the other variants feels out of place, but it is functional as is. Very much approve of moving the vectors, maps, and other such data out of the header and into the source file, though.

louist103:
> Well, louis and I had a discussion about union vs variant, and one union among all the other variants feels out of place, but it is functional as is. Very much approve of moving the vectors, maps, and other such data out of the header and into the source file, though.

I intend to move them to a variant instead of a union. I just haven't had a chance to yet.

Malkierian:
Yeah, I just couldn't remember where that discussion had ended up XD.

Eblo:
#1158 snuck in before this, so there is a merge conflict.

PR #1168, opened 2025-05-26
This adds several mods that when combined should allow a 3-day 100% challenge

athendrix:
Specifically, this adds the following mods to a new "3 Day 100%" section under Difficulty Options:
- Allow Deku Link to damage Sakon the Thief to be able to get the Blast Mask and when combined with others the All-Night Mask during the initial Deku Cycle
- Make the Curiosity Shop owner not racist, and talk to you despite your race (which combined with the next one lets you buy the All-Night Mask)
- Make it so you can purchase the All-Night Mask with bank funds in addition to your wallet funds (It uses your wallet funds first, and then bank funds for the rest)
- Allow the All-Night Mask to keep you from timeskipping to Anju's Grandmother's stories so you can get both Pieces of Heart from her during the 3-Day Challenge without ruining the rest of the time.
- Keep the Mayoral argument going until moonfall (So you can use the Couple's Mask to get the Piece of Heart from them immediately after getting it)

It also moves the "Keep Express Mail" mod under the "3 Day 100%" since you also need that to complete the challenge.

Finally, it moves the "Stop Oceanside Spider House squatter" mod to a "3 Day 100% - Optional" section since it directly assists in the 3 Day Challenge goals without being technically necessary to completing it.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3373191745.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3373239347.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3373430373.zip)


mm/2s2h/BenGui/BenMenu.cpp
```diff
@@ -1414,6 +1404,44 @@ void BenMenu::AddEnhancements() {
...
+        .CVar("gEnhancements.DifficultyOptions.DekuDamageSuttari")
+        .Options(CheckboxOptions().Tooltip(
+            "Allows the player to damage Sakon the Thief as Deku Link allowing the player to obtain the "
+            "Blast Mask and the All-Night Mask in the first Deku Link Only cycle."));
+    AddWidget(path, "Make Curiosity Shop not racist", WIDGET_CVAR_CHECKBOX)
+        .CVar("gEnhancements.DifficultyOptions.NotRacistCuriosityShop")
+        .Options(
+            CheckboxOptions().Tooltip("Allows the player to shop at the curiosity shop regardless of race. "
+                                      "This lets you purchase the All-Night Mask in the first Deku Link Only cycle."));
+    AddWidget(path, "Allow bank money for AN Mask", WIDGET_CVAR_CHECKBOX)
```
balloondude2, line 1436: I would prefer if All-Night Mask wasn't abbreviated to AN Mask. I find that a bit confusing.

mm/src/overlays/actors/ovl_En_Baisen/z_en_baisen.c
```diff
@@ -86,14 +87,16 @@ void EnBaisen_Init(Actor* thisx, PlayState* play) {
     this->paramCopy = this->actor.params;
     if (this->actor.params == 0) {
         this->unk290 = true;
-        if (!CHECK_WEEKEVENTREG(WEEKEVENTREG_63_80) && ((gSaveContext.save.day != 3) || !gSaveContext.save.isNight)) {
+        if (!CHECK_WEEKEVENTREG(WEEKEVENTREG_63_80) && (!GameInteractor_Should(VB_MAYOR_STOP_ARGUING, true) ||
+                                                        (gSaveContext.save.day != 3) || !gSaveContext.save.isNight)) {
```
balloondude2, line 91: Instead of adding `GameInteractor_Should(VB_, true)` to the condition, replace `true` with the vanilla condition you're wanting to override. Since it seems like you're wanting to override the Day 3 night part, put that part of the condition in the Should, like:

`if (!CHECK_WEEKEVENTREG(WEEKEVENTREG_63_80) && GameInteractor_Should(VB_MAYOR_STOP_ARGUING, ((gSaveContext.save.day != 3) || !gSaveContext.save.isNight))) {`

athendrix:
I moved the "Stop Oceanside Spider House squatter" back because it didn't do what I thought it did.
So I went ahead and added a mod that *did* do what I thought it did.
Make it so you can get the Giant's wallet any day from the Oceanside Spider house guy (This is also how it works in MM3D)

I also added another optional mod that prevents the timeskip after defeating Sakon's hideout.
All to assist in the 3-day 100% challenge.
It should be ready for merging again. Please let me know what you think!

balloondude2 (commented):
Thanks for using hooks throughout. I did have a general suggestion on how you're using them though.

Eblo (commented):
From a high level perspective of scope, I think bundling all of these different enhancements together for this one purpose is the wrong approach. Some of these enhancements are useful time savers or convenient of their own accord, independently of others. For example, the enhancement to let the player directly use bank funds for purchases sounds like a good streamline to the general experience of the game, whether or not the player is playing to 100% the game as fast as possible.

Conversely, some of these seem to only benefit this niche goal, and I'm not sure if that belongs in upstream 2ship. Modding territory feels more appropriate. Allowing Deku to damage Sakon doesn't have much application beyond the singular purpose of getting the Blast Mask and other rewards in the first cycle, for instance.

athendrix:
@Eblo So at the moment there are 7 modifications here total.
1. Deku Damaging Sakon
2. Curiosity Shop Guy talking to Deku (And others, but I could change it to just Deku)
3. Bank funds for the All-Night Mask only
4. No Timeskip for Anju's Grandma's stories **if and only if you're wearing the All-Night Mask**
5. Keep the Mayor arguing so you can get the heart piece immediately after getting the couple's mask
6. Get Wallet from Oceanside Spider House on Any Day (like MM3D)
7. No Timeskip for Sakon's hideout (you skip to midnight after exiting the hideout)

Probably the useful ones for general use would be:
3 (Modified so it works on all purchases like you suggested)
6, since it's a MM3D restoration
4 and 7 since they stop some of the more annoying timeskips in the game
and *maybe* 5 since it's fairly convenient to be able to get that piece of heart immediately after getting the couple's mask, and before a reset.

This leaves 1 and 2 as the less useful ones for general purposes, so what if I bundled them together under one "Useful Deku Cycle" option that does 1 and 2 together and also forces 3 on (since there's no point talking to him if you can't buy anything). The idea being that together, they make the Deku Cycle more exciting than it otherwise is for any playthrough. Or if that's not enough together, I could also add making the guards ignore you as a Deku scrub, since it's a fairly minor glitch to sneak past them anyway.

Eblo:
My recommendation at this stage is to separate all of these enhancements into individual pull requests so that they be evaluated on a case-by-case basis. For example:

>  Make it so you can get the Giant's wallet any day from the Oceanside Spider house guy (This is also how it works in MM3D)

A PR that does that just got approved. With the nature of your PR as a monolith of many different enhancements, it now carries a redundant feature in the mix. If all of your enhancements were separated, then the others would not be impacted.

I'm still not sold on the idea of having a single package of code labeled for hitting this particular self-imposed challenge in the game, but having many individually useful enhancements with a preset to bring them together is the way to go.

> Or if that's not enough together, I could also add making the guards ignore you as a Deku scrub, since it's a fairly minor glitch to sneak past them anyway.

I think something like a "Guards Ignore Link" cheat is fair game. There's already some flavor of that with fixing a Fierce Deity annoyance. That would be one way to do this in a way that players might use on its own for reasons besides this challenge run.

Regarding your code itself, you have a review comment regarding your use of `GameInteractor_Should`. I'll add on to that to say it applies to all of your uses of `GameInteractor_Should`. Source modifications should be limited to wrapping existing conditions with that call and then putting all custom logic within the enhancement file, if possible. It certainly is possible in these cases.

garrettjoecox:
I also agree with this direction

> My recommendation at this stage is to separate all of these enhancements into individual pull requests so that they be evaluated on a case-by-case basis.

Going to close this out for now, feel free to submit them separately

PR #1169, opened 2025-05-27
Fix BGM replay bug caused by enhancements that bypass normal scene transitions

Eblo:
Note: This is separate albeit symptomatically similar to the Great Bay Coast cutscene music bug.

Some of 2ship's enhancements give rise to a bug where the main BGM sequence is seen as `NA_BGM_DISABLED` even if music is playing. This breaks the behavior of `Audio_PlayBgm_StorePrevBgm` and causes `Audio_RestorePrevBgm` to play silence. So far, it seems like debug warping, faster scene transitions, and loading into a pause save result in this bug for the immediate subsequent scene, until the player transfers to another scene in a vanilla way.

In a normal transition, the main BGM sequence player is supposed to be disabled when the fadeTimer is exactly 0. The enhancements in question bypass this in one way or another, causing the bug. This fix is enabled when any of those three enhancements is enabled.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3348903759.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3348975350.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3349251125.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Fixes/BgmReplay.cpp
+++ b/mm/2s2h/Enhancements/Fixes/BgmReplay.cpp
@@ -7,9 +7,13 @@ extern "C" {
 #include "functions.h"
 }
 
-#define CVAR_FASTER_SCENE_TRANSITIONS CVarGetInteger("gEnhancements.Timesavers.FasterSceneTransitions", 0)
-#define CVAR_PAUSE_SAVE CVarGetInteger("gEnhancements.Saving.PauseSave", 0)
-#define CVAR_DEBUG_MODE CVarGetInteger("gDeveloperTools.DebugEnabled", 0)
+#define CVAR_NAME_FASTER_SCENE_TRANSITIONS "gEnhancements.Timesavers.FasterSceneTransitions"
+#define CVAR_NAME_PAUSE_SAVE "gEnhancements.Saving.PauseSave"
+#define CVAR_NAME_DEBUG_MODE "gDeveloperTools.DebugEnabled"
+
+#define CVAR_FASTER_SCENE_TRANSITIONS CVarGetInteger(CVAR_NAME_FASTER_SCENE_TRANSITIONS, 0)
+#define CVAR_PAUSE_SAVE CVarGetInteger(CVAR_NAME_PAUSE_SAVE, 0)
+#define CVAR_DEBUG_MODE CVarGetInteger(CVAR_NAME_DEBUG_MODE, 0)
 
 /*
  * Certain 2ship features circumvent sequence player state manipulation: debug warping, loading into a dungeon via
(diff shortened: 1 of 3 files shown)


mm/2s2h/Enhancements/Fixes/BgmReplay.cpp
```diff
@@ -0,0 +1,29 @@
...
+ * pause save, and faster scene transitions for example. Something about these different features causes
+ * SEQ_PLAYER_BGM_MAIN to have the wrong isSeqPlayerInit and enabled flags. A result of this is that the active main BGM
+ * sequence is seen as NA_BGM_DISABLED even if BGM is playing. Attempts to store and restore BGM from this point will
+ * play silence, as seen with mini-boss battles.
+ *
+ * Rather than get in the weeds of tracking audio state information between different enhancements, a simple workaround
+ * is to call AudioScript_SequencePlayerDisable in an OnSceneInit hook. This forces SEQ_PLAYER_BGM_MAIN to reset its
+ * state information.
+ *
+ * Enabled by default, as different enhancements result in the same bug. This shouldn't hurt any vanilla behavior.
```
garrettjoecox: _shouldn't_

should we gate this behind a check for any of the CVars to be cautious, or are you confident in this assumption that it won't cause any more?

Eblo: > or are you confident in this assumption that it won't cause any more?

~~Not 100%. After submitting this PR, it occurred to me that I could gate this behind an OR of the three enhancements that I know cause this bug. If that sounds good, I can do that.~~

Big oops, this breaks transitions where both scenes have the same BGM. I'm marking this as a draft now to revisit later.

Eblo:
I did some more digging to isolate the exact point where this becomes a problem. Transitions, which normally fade at 60 frames, kill the main BGM sequence player when the fadeTimer is exactly 0:

https://github.com/HarbourMasters/2ship2harkinian/blob/298fea47fa20f7603757b785392311f4d9ec8d0b/mm/src/audio/lib/effects.c#L80-L84

Faster scene transitions change scenes before the fadeTimer hits 0. Debug warping or loading into a pause save skip the fading out part of the transition entirely. All cases result in not calling `AudioScript_SequencePlayerDisable` between the old BGM and new BGM playing.

My previous approach was close, but the `SceneInit` hook is too late in the logic. I added a new hook that operates at the point where a scene plays its BGM. This fix is still only enabled when any of the three relevant enhancements are enabled.

mm/src/audio/code_8019AF00.c
```diff
@@ -5478,6 +5478,11 @@ void Audio_StartMorningSceneSequence(u16 seqId) {
 }
 
 void Audio_PlaySceneSequence(u16 seqId, u8 dayMinusOne) {
+    if (!GameInteractor_Should(VB_PLAY_SCENE_SEQUENCE, true, &sRequestedSceneSeqId, &sPrevMainBgmSeqId, &seqId,
```
balloondude2: Non blocking for me, but I would prefer `if (GameInteractor_Should(VB_PLAY_SCENE_SEQUENCE, false,...`.

I think that would make it clearer that the early return isn't vanilla.

Edit: Well, I guess a VB return at the top of a function inherently says "2ship". I still think a `false` is better than `not true` in this case.

mm/src/audio/code_8019AF00.c
```diff
@@ -5478,6 +5478,11 @@ void Audio_StartMorningSceneSequence(u16 seqId) {
...
+                               &dayMinusOne)) {
```
balloondude2: It doesn't look like you end up using `dayMinusOne` in the enhancement file.

balloondude2 (approved):
LGTM. Left a couple of minor comments.

PR #1170, opened 2025-05-29
[Rando] Add Grass CSMC

balloondude2:
This adds custom models (courtesy of @DanaTheElf) to randomized grass. Includes both CSMC and non-CSMC models. 

grassCSMC

grassNonCSMC

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3284181016.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3284202198.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3284229851.zip)


mm/2s2h/Rando/ActorBehavior/ObjGrass.cpp
```diff
@@ -1034,7 +1038,151 @@ void SpawnGrassDrop(Vec3f pos, RandoCheckId randoCheckId) {
...
+    Matrix_SetTranslateRotateYXZ(grassElem->pos.x, grassElem->pos.y, grassElem->pos.z, &rot);
+    Matrix_Scale(objGrass->actor.scale.x, objGrass->actor.scale.y, objGrass->actor.scale.z, MTXMODE_APPLY);
+    if (grassElem->flags & OBJ_GRASS_ELEM_ANIM) {
+        ObjGrass_OverrideMatrixCurrent(&objGrass->distortionMtx[j]);
+    }
+
+    gSPMatrix(POLY_OPA_DISP++, Matrix_NewMtx(gPlayState->state.gfxCtx), G_MTX_NOPUSH | G_MTX_LOAD | G_MTX_MODELVIEW);
+    gSPDisplayList(POLY_OPA_DISP++, GetObjGrassDList(randoCheckId));
+    gSPDisplayList(POLY_OPA_DISP++, (Gfx*)gObjGrass_D_809AA9F0);
+    CLOSE_DISPS(gPlayState->state.gfxCtx);
```
Eblo, line 1139: I think this code is causing some issue with interpolation. Is this necessary? Pots for comparison only change what DL is being used with CSMC:

https://github.com/HarbourMasters/2ship2harkinian/blob/3221402a77079c2e316563bc032c23fd5c63876b/mm/2s2h/Rando/ActorBehavior/ObjTsubo.cpp#L385-L397

Eblo (commented):
I do like the models. But I am seeing interpolation issues with CSMC grass only, while normal grass is fine:

https://github.com/user-attachments/assets/0609ca52-4eb5-4df4-979e-0b89d2a69e32

mm/2s2h/Rando/ActorBehavior/ObjGrass.cpp
balloondude2: I don't recall having interpolation issues, and I get a 404 error trying to view the attachment you left in the other comment (which I assume is a video of the issue you're seeing).

For context, there are currently three different actors related to Grass Shuffle:
1. en_kusa, non-termina field bushes (and currently unshuffled regrowable grass)
2. obj_grass, a single actor that manages all of the termina field grass. obj_grass contains grassGroups, which contain grassElems, which are the individual bushes. ie, many grass, one actor
3. and obj_grass_carry, which Link actually picks up when he "picks" up one of the grass managed by obj_grass.

En_kusa and obj_grass_carry have simple vanilla draw functions that are similar to obj_tsubo, and their rando draw function were modeled after `ObjTsubo_RandoDraw()`

However, obj_grass's draw function is much more complicated and I don't think an approach similar to obj_tsubo is feasible. In the draw function, it loops through each grassElem and draws them based on flags (which are set based on distance from camera or something). I use a VB_Should to hijack the draw process at the individual grassElem portion, and essentially duplicate it here with the new DLs. 

Until I see what the issue you're seeing is, I'm not sure what will need to be done to correct it.

Eblo: Apparently GitHub doesn't like it if there's text before a video without a new line. Try now.

balloondude2: Thanks, I was able to replicate and pushed up a fix. Let me know if you still see any other issues.

Eblo (approved):

PR #1179, opened 2025-06-20
Lus Bump

louist103:
Mostly cleanups and big fixes. Biggest change is LUS excluding MPQ support by default.
Blockers:
https://github.com/Kenix3/libultraship/pull/902

List of changes:
Bump ImGui https://github.com/Kenix3/libultraship/pull/884
Update Prism https://github.com/Kenix3/libultraship/pull/887
Close O2R archive in destructor https://github.com/Kenix3/libultraship/pull/889
Properly Shutdown ImGui: https://github.com/Kenix3/libultraship/pull/888
Cleanup destructors https://github.com/Kenix3/libultraship/pull/886
MacOS cleanups https://github.com/Kenix3/libultraship/pull/890 https://github.com/Kenix3/libultraship/pull/892
Fix DX11 memory leak https://github.com/Kenix3/libultraship/pull/896
Use `pragma once` for all header guards https://github.com/Kenix3/libultraship/pull/897
Fix font double free https://github.com/Kenix3/libultraship/pull/900

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3549855143.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3549897425.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3550233054.zip)

PR #1182, opened 2025-06-23
Use registration system instead of polling for dropped files

louist103:
Moves the file drop system to a registration system handled by LUS as opposed to a polling based system handled by the port.
This PR includes commits to work with latest LUS. They can be dropped once #1179 is merged.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3838690714.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3838700939.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3838751376.zip)

### DIFF
--- a/mm/2s2h/BenPort.cpp
+++ b/mm/2s2h/BenPort.cpp
@@ -738,7 +738,6 @@ extern "C" void InitOTR() {
     std::shared_ptr<Ship::Config> conf = OTRGlobals::Instance->context->GetConfig();
     Ship::Context::GetInstance()->GetFileDropMgr()->RegisterDropHandler(BinarySaveConverter_HandleFileDropped);
     Ship::Context::GetInstance()->GetFileDropMgr()->RegisterDropHandler(SaveManager_HandleFileDropped);
-    Ship::Context::GetInstance()->GetFileDropMgr()->RegisterDropHandler(SaveManager_HandleFileDropped);
 }
 
 extern "C" void SaveManager_ThreadPoolWait() {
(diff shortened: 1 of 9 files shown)


Eblo:
> This PR includes commits to work with latest LUS. They can be dropped once https://github.com/HarbourMasters/2ship2harkinian/pull/1179 is merged.

Now that it is merged, could you update this accordingly?

mm/2s2h/BenPort.cpp
```diff
@@ -765,6 +736,9 @@ extern "C" void InitOTR() {
 #endif
 
     std::shared_ptr<Ship::Config> conf = OTRGlobals::Instance->context->GetConfig();
+    Ship::Context::GetInstance()->GetFileDropMgr()->RegisterDropHandler(BinarySaveConverter_HandleFileDropped);
+    Ship::Context::GetInstance()->GetFileDropMgr()->RegisterDropHandler(SaveManager_HandleFileDropped);
+    Ship::Context::GetInstance()->GetFileDropMgr()->RegisterDropHandler(SaveManager_HandleFileDropped);
```
Eblo: Does this handler need to be registered twice?

Eblo (commented):
Tested JSON saves, binary saves, spoiler logs, and presets on Windows. All check out. Left a question.

mm/2s2h/BenPort.cpp
louist103: No. I had it there to test something. I will remove it.

Eblo (approved):

PR #1183, opened 2025-06-24
[Rando] Change rando sword pickup fix to check if not FD

Eblo:
#1131 fixed a bug where picking up swords as FD erroneously equipped that sword immediately. This fix only changed the sword equip if the player is in human form. However, this inadvertently caused sword upgrades to only cosmetically change if picked up in any of the three other forms. The HUD icon and damage output would still match whichever sword the player had before.

This fix tweaks it to instead check if the player is not FD. FD is the only affected form by the original bug, as FD and human share equip slots. Other forms have their own equip slots, thus it is safe to acquire sword upgrades in those forms.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3395356859.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3395375711.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3395533585.zip)


balloondude2 (approved):
Third time's the charm

PR #1190, opened 2025-07-02
Restrict Rando Settings for French Vanilla and Vanilla

### LINKED ISSUE #1173: French Vanilla Logic Setting is Incompatible with Some Settings
File generation will fail with French Vanilla if options that add items to the pool are enabled, such as Boss Souls or Plentiful Items. 

Some possible solutions:

- Adjust French Vanilla to be compatible with more items
- Disable incompatible settings in the menu when French Vanilla is selected
- Update French Vanilla's tooltip
- Remove French Vanilla entirely

balloondude2:
Mostly resolves #1173 

This updates the tooltip descriptions for Vanilla and French Vanilla, clears incompatible settings when selecting those Logic Settings, and disables the incompatible settings when those Logic settings are currently selected.

This does not handle resetting or disabling Starting Items to ensure that one sword and the Hero's Shield are included. For Vanilla logic, this isn't an issue because a starting sword and shield are hardcoded in. For French Vanilla, seeds starting without a sword and shield will not have them in the pool at all. ie. players will never be able to get a Gilded Sword and have one less Hero's Shield available. In either case, choosing to start without a sword or shield for French Vanilla or Vanilla will not prevent the seed from generating or from being beatable, so this can be handled later imo. 

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3492400194.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3492413157.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3492429531.zip)


Eblo (approved):
I think this is fine for now. I'm curious if @Malkierian has any quick thoughts on this, both as the modern menu developer and as someone with extensive knowledge of SoH's randomizer.

balloondude2:
Since I was in this file, I went ahead and added a couple tooltips for some less obvious settings.

PR #1197, opened 2025-07-05
Fix incorrect backend ID for OpenGL in README.md

### LINKED ISSUE #1194: OpenGL Backend on Windows No Longer Works
Summary:
The OpenGL rendering backend no longer works on Windows, even when manually configured in 2ship2harkinian.json. The game still launches using DirectX 11.

Steps to reproduce:

Edit 2ship2harkinian.json to:

```
"Backend": {
  "id": 3,
  "Name": "OpenGL"
}
```

Launch 2ship.exe

It still uses DirectX11 instead of OpenGL.

Expected behavior:
The game should use the OpenGL backend when id: 3 is set in the config file.

Actual behavior:
DirectX11 backend is still used regardless of the setting.

Notes:
- im on windows 10

-  i use rx5500xt with 8gb vram and i7 3770 with 16g ram (i dont this is relevant )

-  The OpenGL backend works on Linux, so it seems Windows-specific

Additional Information:

-  I also tried marking the .json file as read-only — no effect.

- Using wine under Linux but it triggers crash.

-  I’m trying to test OpenGL-only behavior for development/debugging purposes .

    Could be hardcoded default to DX11 on Windows builds ?

Ammar-Sadaoui:
This PR corrects the graphics backend IDs listed in the README.md file.

Previously, the README incorrectly listed:
- OpenGL as ID 3 (should be 1)
- DirectX 11 as ID 2 (should be 0)
- Metal as ID 4 (should be 2)

Also clarified behavior when setting an invalid ID, which can cause the program to silently fall back to the default backend or crash.

The updated section now provides accurate backend ID values and keeps the original guidance on changing APIs via the settings menu or the JSON file.

Fixes #1194

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3469215212.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3469215906.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3469225595.zip)

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,13 @@ Congratulations, you are now sailing with 2 Ship 2 Harkinian! Have fun!
 | Ctrl+R | Reset |
 
 ### Graphics Backends
-Currently, there are three rendering APIs supported: DirectX11 (Windows), OpenGL (all platforms), and Metal (macOS). You can change which API to use in the `Settings` menu of the menubar, which requires a restart.  If you're having an issue with crashing, you can change the API in the `2ship2harkinian.json` file by finding the line `"Backend":{`... and changing the `id` value to `3` and set the `Name` to `OpenGL`. `DirectX 11` with id `2` is the default on Windows. `Metal` with id `4` is the default on macOS.
+Currently, there are three rendering APIs supported: DirectX 11 (Windows), OpenGL (all platforms), and Metal (macOS). You can change which API to use in the `Settings` menu of the menubar, which requires a restart.
+
+If you're having an issue with crashing, you can also change the API manually in the `2ship2harkinian.json` file by finding the `"Backend": {` section and updating the backend ID and name. Be sure to use one of the valid values:
+
+- `0` = DirectX 11 (default on Windows)
+- `1` = OpenGL
+- `2` = Metal (default on macOS)
 
 # Custom Assets
 


Eblo (approved):
For future reference in case this ever changes again, these values come from the `WindowBackend` enum in LUS: https://github.com/Kenix3/libultraship/blob/5e33d3e7cd0396f847923cd6c471eaf324e90351/src/window/Window.h#L13

Ammar-Sadaoui:
> For future reference in case this ever changes again, these values come from the `WindowBackend` enum in LUS: https://github.com/Kenix3/libultraship/blob/5e33d3e7cd0396f847923cd6c471eaf324e90351/src/window/Window.h#L13

should i make pull requests for others project that use [libultraship](https://github.com/Kenix3/libultraship) with incorrect backend ID by HarbourMasters ?

Eblo:
> should i make pull requests for others project that use [libultraship](https://github.com/Kenix3/libultraship) with incorrect backend ID by HarbourMasters ?

I'm not involved with development for other projects, but I don't see why not. If you ran into this confusion here, others might elsewhere.

amayra:
why they changed the order in the first place?

PR #1202, opened 2025-07-18
Skip Intro Persistence Fix

### LINKED ISSUE #1140: Investigate save/load and moon crash issues
When testing #970, I ran into strange issues related to saving and loading. They did not seem to be caused by the feature itself. Now that other users have since reported similar issues and that PR is not merged, this seems to be an issue with the main branch. To reiterate my findings on that PR:

> I seem to be getting these issues no matter what my time saver settings are, and even if I have actually saved. This includes playing through the entire intro up to the first cycle, or playing in subsequent cycles. This happens even with "Skip Story Cutscenes" disabled, and therefore this enhancement does not run, so that much is probably unrelated.
>
> - Various saves are just throwing me back to the start of the intro when I load them. Happens with first cycle pause saves (which may be unforeseen consequences of saving during a part of the game that wasn't built for it), but also happens with a legitimate Song of Time save from playing atop the Clock Tower. Since the moon crash isn't a factor here, again I think it's unrelated to this enhancement, but a big issue nonetheless.
> - If I moon crash during first cycle, Link will always be human in the Clock Tower interior cutscene. The Happy Mask Salesman will be invisible for most of it as well. My inventory is wiped, but maybe that's a first cycle thing?
> - If I moon crash after first cycle, the Clock Tower interior plays the "wow you got your instrument!" cutscene. Link will always be Deku as expected for that scene. I can do this multiple times in a row; it never seems to change what Happy Mask Salesman cutscene I get. This wipes some things from my inventory, including Deku Mask, Razor Sword, bottles, and other masks.

Here's a video provided by a user who encountered this issue. Context:

>  think i last saved via song of time
> went to load it back up and its starting from the intro
> on nightly tho so not suprised
> could it be because ive never used an owl before? bc of index warp
> thats my only guess but idk the specifics

https://github.com/user-attachments/assets/734c58a0-e4f1-478e-9fe9-c82e384448fd

To complicate it further, I'm having trouble now reproducing with my own steps that I outlined myself when I first ran into this.

mckinlee:
Pre-fix:
https://imgur.com/a/dzjs19O

Post Fix:
https://imgur.com/a/Iosjvsk

I was not able to reproduce all the issues reported in #970 & #1140 but this could potentially fix those as well.

Added persistence to SkipTatlInterrupt, I didn't test it, but Balloondude confirmed it fixed it as well.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3568832926.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3568836726.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3568879099.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipTatlInterrupts.cpp
+++ b/mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipTatlInterrupts.cpp
@@ -19,6 +19,7 @@ void RegisterSkipTatlInterrupts() {
         if (gSaveContext.save.entrance == ENTRANCE(SOUTH_CLOCK_TOWN, 0) && gSaveContext.save.cutsceneIndex == 0 &&
             !CHECK_WEEKEVENTREG(WEEKEVENTREG_59_04)) {
             Flags_SetWeekEventReg(WEEKEVENTREG_59_04);
+            Sram_SaveSpecialEnterClockTown(gPlayState);
         }
     });
 
--- a/mm/2s2h/Enhancements/Cutscenes/SkipIntroSequence.cpp
+++ b/mm/2s2h/Enhancements/Cutscenes/SkipIntroSequence.cpp
@@ -60,6 +60,8 @@ void RegisterSkipIntroSequence() {
             SET_WEEKEVENTREG(WEEKEVENTREG_ENTERED_WEST_CLOCK_TOWN);
             SET_WEEKEVENTREG(WEEKEVENTREG_ENTERED_NORTH_CLOCK_TOWN);
         }
+        // Persist this modified state so moon crashes without saving keep it
+        Sram_SaveSpecialEnterClockTown(gPlayState);
     });
 }
 


balloondude2 (dismissed):
Nice find!

Eblo (approved):
The crux of the issues is that `gSaveContext.save.isFirstCycle` was not being set. This gets set by `Sram_SaveSpecialEnterClockTown`, which normally gets called when entering South Clock Town for the first time. That happens with the `CS_MISC_SAVE_ENTER_CLOCK_TOWN` cutscene command, which plays when Tatl first talks to Link in South Clock Town.

It makes sense for both intro skip and that particular Tatl interrupt to call that save function. Good work tracking this down.

~~Fixes the bigger issues with #1140. I still get the issue where post-first cycle moon crashes always give the learning SoH cutscene. With this fix, that rules out any `isFirstCycle` issues as the cause.~~

Fixes #1140.

balloondude2 (approved):
For real this time 😅 

In addition to isFirstCycle, this ensure certain cycleSceneFlags and the player's form are persisted correctly.

Issue #1184, opened 2025-06-24
Asset generation stalls if multiple valid ROMs exist in folder


Eblo:
It seems that, if multiple valid ROMs exist in the 2ship folder, then nothing will happen when clicking "Yes" to generate the O2R. Users can simply move the ROMs out of the folder to get around this for now, but this should be addressed.

PR #1203, opened 2025-07-18
Add ROM Selector

mckinlee:
image
In the event the user has multiple roms, this will detect it and allow them to choose which one rather than failing. Addresses #1184 

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3566959311.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3566960846.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3567114918.zip)

### DIFF
--- a/mm/2s2h/Extractor/Extract.cpp
+++ b/mm/2s2h/Extractor/Extract.cpp
@@ -441,6 +441,17 @@ bool Extractor::Run(std::string searchPath, RomSearchMode searchMode) {
         }
     }
 
+    if (roms.size() > 1) {
+        int ret = ShowYesNoBox("Multiple ROMs Found", "Multiple ROM files were detected. Select one manually?");
+        if (ret == IDYES) {
+            if (!ManuallySearchForRomMatchingType(searchMode)) {
+                return false;
+            }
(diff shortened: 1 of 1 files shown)


mckinlee:
If someone can reproduce the issue, their assistance would be greatly appreciated. This was a quick write up to workaround it if need be, but I'd like to address the root cause. The reporter said they are on Windows 10 for additional context. As I'm not able to reproduce, that makes me assume it could be potentially related. If it is related, my assumption would be we need to revisit how we are ensuring the paths are valid.

Eblo (approved):
I have tested the issue in question, but I have no had luck in reproducing it. The behavior I see in 2.0.0 is that it will just prompt for and identify each ROM sequentially if you answer "No". The application certainly isn't doing "nothing".

I wonder if it was a path issue after all. While I cannot say whether this PR fixes that problem, I do like the option it presents as opposed to going through multiple ROMs sequentially.

PR #1213, opened 2025-07-23
Save play time, support timer display options

Eblo:
Currently, save files have a timer display that shows the naive difference between the current time and the time when the file was created. This PR improves the time tracking so that it tracks the time actually spent between saves and not real time elapsed regardless of play.

2ship needs to update the tracked time in chunks with `SavingEnhancements_AdvancePlaytime()`. This gets called when:
* Saving or speaking to an owl statue
* Defeating Majora's Wrath, i.e. beating the game

Some logic for initializing and marking game completion were handled in the timer display element. I moved that to `SavingEnhancements.cpp`.

This necessitates and adds a save migration.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4649375382.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4649400698.zip)

### DIFF
--- a/mm/include/z64save.h
+++ b/mm/include/z64save.h
@@ -397,8 +397,7 @@ typedef struct ShipSaveInfo {
     s32 pauseSaveEntrance;
     SaveType saveType;
     uint64_t fileCreatedAt;
-    uint64_t fileCompletedAt; // For now this is always Majora final blow, has the potential to be something else later on
-    uint64_t lastTimeLog; // Persisted value not relevant, but needs stored to calculate filePlaytime
+    uint64_t fileCompletedAt;
     uint64_t filePlaytime;
     char commitHash[8];
     RandoSaveInfo rando;
@@ -435,6 +434,7 @@ typedef struct DpadSaveContext {
 // See `ShipSaveInfo` for values on the SaveContext that are persisted.
 typedef struct ShipSaveContext {
     DpadSaveContext dpad;
+    uint64_t lastTimeLog;
 } ShipSaveContext;
 // #endregion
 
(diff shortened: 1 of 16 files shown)


balloondude2:
I don't think the original timer was naive, but meant to be an RTA timer. I think we shouldn't replace it, but instead add this Play Time timer as an additional option.

Eblo:
Fair enough. I think giving the display different options is a good idea. I'll return to this.

Eblo:
I reworked it so that the timer display is not a toggle, but a list of options. Supports RTA, IGT, or no display (default).

image

image

image

Caladius:
Thank you for adding this, as with https://github.com/HarbourMasters/2ship2harkinian/pull/1266 my intention here was for Speedruns specifically. The thought of just In Game Play-time was there but wasn't my priority at the time.

Eblo:

mckinlee (approved):

mm/include/z64save.h
```diff
@@ -398,6 +398,8 @@ typedef struct ShipSaveInfo {
     SaveType saveType;
     uint64_t fileCreatedAt;
     uint64_t fileCompletedAt; // For now this is always Majora final blow, has the potential to be something else later on
+    uint64_t lastTimeLog; // Persisted value not relevant, but needs stored to calculate filePlaytime
```
garrettjoecox: See `ShipSaveContext`

garrettjoecox (approved):
Move lastTimeLog to `ShipSaveContext`, aside from that LGTM. 

Clever use of timestamps to avoid incrementing a number every frame. Both that and the ability to have both RTA & gametime should be back-ported to SoH if you're feeling adventurous.

Eblo:
Moved the intermediate time tracking variable to `ShipSaveContext`. I noticed that completion time was not being properly tracked in Triforce Hunt, so I added a hook for `OnGameCompletion` that fires on completing the Triforce and defeating Majora. Given that this could do other things in the future, I left time splits alone, so they still set the completed time directly as before.

PR #1219, opened 2025-07-25
[Cheat] Climb Anywhere

mckinlee:
When enabled, it simply forces WALL_FLAG_3 at the time wall flags are fetched.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3642011822.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3642058385.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3642538130.zip)


mm/src/code/z_bgcheck.c
```diff
@@ -4207,7 +4207,11 @@ WallType SurfaceType_GetWallType(CollisionContext* colCtx, CollisionPoly* poly,
 }
 
 s32 SurfaceType_GetWallFlags(CollisionContext* colCtx, CollisionPoly* poly, s32 bgId) {
-    return sWallFlags[SurfaceType_GetWallType(colCtx, poly, bgId)];
+    s32 result = sWallFlags[SurfaceType_GetWallType(colCtx, poly, bgId)];
+    if (GameInteractor_Should(VB_BE_CLIMBABLE_SURFACE, result & WALL_FLAG_3, poly, bgId)) {
+        result |= WALL_FLAG_3;
+    }
+    return result;
```
balloondude2: Instead of `result & WALL_FLAG_3` the vanilla condition should be `false` because the behavior within the should never happens in vanilla. 
Additionally, neither `poly` nor `bgId` are used in the hook, and should be removed here. 
I'd also like the original return line to remain unchanged in the vanilla case.

```suggestion
    if (GameInteractor_Should(VB_BE_CLIMBABLE_SURFACE, false)) {
        return sWallFlags[SurfaceType_GetWallType(colCtx, poly, bgId)] | WALL_FLAG_3;
    }
    return sWallFlags[SurfaceType_GetWallType(colCtx, poly, bgId)];
```

balloondude2 (changes requested):
Worked great from my limited testing. I left a few notes on the VB implementation.

mm/src/code/z_bgcheck.c
mckinlee: Went ahead and committed your exact suggested change—early return for the cheat, vanilla return untouched otherwise. I lean DRY out of habit, but I get the case for vanilla return flow in a project like this.

Thanks for the feedback!

balloondude2 (approved):
Thanks for making that change!

Eblo (approved):

PR #1224, opened 2025-07-27
[Rando] Add setting for minimum Stray Fairies

Eblo:
Adds a setting to rando generation to set the required Stray Fairies to get the Great Fairy check. Does not change how many Stray Fairies are shuffled; this only lowers the threshold with which the Great Fairy will be restored. This does not affect the Clock Town fairy, which only needs 1 anyway.

image

I renamed the tracker display from `HAS_ALL_STRAY_FAIRIES` to `HAS_ENOUGH_STRAY_FAIRIES` to be a little clearer.

image

If this approach seems good, we can do something similar for Skulltula tokens.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3812362166.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3812396470.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3812700208.zip)


balloondude2:
For at least Woodfall and Great Bay, going in with the minimum required fairies triggers two cutscenes. This isn't noticeable with Misc Skips on, as they both get skipped.

Eblo:
Yeah, looks like that first cutscene ends up reloading the scene, which thus re-runs the actor init and causes the second cutscene to play out immediately after. I pushed a change to check the `eligible` flag in the hook to prevent that. While I was at it, I refactored the check lookup to use a map, lest I add another switch block.

balloondude2 (approved):

PR #1232, opened 2025-07-31
Check previousBankValue in custom bank thresholds

Eblo:
Fixes a bug with Lower Bank Reward Thresholds where the interest reward would supersede the second threshold reward. The hook'd condition did not also check `previousBankValue` like the vanilla condition, which caused the interest reward to be true and thus preempt the next reward.

To test this, be sure to reload the scene between thresholds. Without this fix, the bug only happens if the thresholds are not hit in the same scene load.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3654264500.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3654294584.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3654549912.zip)

### DIFF
--- a/mm/2s2h/Enhancements/DifficultyOptions/CustomBankRewardThresholds.cpp
+++ b/mm/2s2h/Enhancements/DifficultyOptions/CustomBankRewardThresholds.cpp
@@ -19,13 +19,15 @@ void RegisterCustomBankRewardThresholds() {
     COND_VB_SHOULD(VB_PASS_FIRST_BANK_THRESHOLD, CVAR, {
         EnGinkoMan* enGinkoMan = va_arg(args, EnGinkoMan*);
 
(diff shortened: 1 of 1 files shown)


balloondude2:
I tried testing the bug on a develop build to get a baseline, but I couldn't seem to replicate the issue. What are the steps to reproduce?

Eblo:
The key thing is to leave the scene and come back for each threshold. If you do it all without leaving West Clock Town, you won't hit the bug. Or maybe you do SoT between each one; I forget.

balloondude2 (approved):

PR #1238, opened 2025-08-03
Separate Stone Tower Temple save entrances

Eblo:
Separates STT and the inverted variant so that they don't both load into the normal entrance. This is consistent with game over respawns and removes an annoyance where the player would be forced to watch the flip cutscene when loading into this state and exiting the dungeon.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3676979529.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3676979892.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3676992917.zip)


balloondude2 (approved):
Would we want this on Rena? Feels kinda like an oversight, so I can see it going either way.

Eblo:
I thought about it. Since it has implications for save data, I thought it better for develop.

PR #1240, opened 2025-08-06
Save editor enhancements

Eblo:
- Add description tooltips to weekEventReg, eventInf, and owlActivation. Sourced from TCRF: https://tcrf.net/Proto:The_Legend_of_Zelda:_Majora's_Mask/Debug_Version/Event_Editor
  - weekEventReg checkboxes are color-coded based on the lifespan of the flag as indicated by that TCRF article (persistent, cycle, scene)
- Allow player to remove shield via editor, like sword
- Add ability to toggle Chateau magic

image

image

image

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3890540936.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3890543091.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3890558950.zip)


mm/2s2h/DeveloperTools/SaveEditor.h
```diff
@@ -1,6 +1,943 @@
...
+} FlagEntry;
+
+typedef struct {
+    const char* name;
+    FlagTableType flagTableType;
+    std::vector<FlagEntry> entries;
+} FlagTable;
+
+// Reference https://tcrf.net/Proto:The_Legend_of_Zelda:_Majora's_Mask/Debug_Version/Event_Editor
+// The source was last referenced on 2025-08-05 and had a last updated value of 2024-06-29
```
balloondude2, line 40: This comments implies to me that this list should always come from tcrf. Is that what you intend, or do you see these potentially being updated as 2ship contributors document them?

mm/2s2h/DeveloperTools/SaveEditor.cpp
```diff
@@ -1822,25 +1827,26 @@ void DrawFlagsTab() {
             UIWidgets::DrawFlagArray32("##collectible3", gPlayState->actorCtx.sceneFlags.collectible[3]);
             ImGui::EndGroup();
             break;
-        case 1: // weekEventReg
+        case WEEK_EVENT_REG:
```
balloondude2, line 1839: With the persistent, cycle, scene/other coloring, I think we should add some kind of legend when the flag type is weekEventReg. 

Perhaps some dummy checkboxes at the top, with tooltips explaining the color?

mm/2s2h/DeveloperTools/SaveEditor.h
Eblo: I followed SoH's similar comment for its own flagtables:
https://github.com/HarbourMasters/Shipwright/blob/7b4df9bdb2ebde2bdd6581c10ffdc1aea4ee1573/soh/soh/Enhancements/debugger/debugSaveEditor.h#L28-L29

This isn't to mean that 2ship absolutely must source the TCRF article, but rather to document what source the flagtable used for its contents. In theory, that source could be updated in the future as unknown flags become understood and documented.

mm/2s2h/DeveloperTools/SaveEditor.cpp
Eblo: A legend sounds like a good idea.

mm/2s2h/DeveloperTools/SaveEditor.h
balloondude2: Great, just wanted to make sure.

mm/2s2h/DeveloperTools/SaveEditor.h
```diff
@@ -1,6 +1,943 @@
...
+          { CYCLE_RESET, WEEKEVENTREG_91_02, "Mikau Performed His Song" },
+          { CYCLE_RESET, WEEKEVENTREG_91_04, "Sent Letter to Kafei On Time" },
+          { CYCLE_RESET, WEEKEVENTREG_91_08, "Sent Letter to Kafei Late" },
+          { CYCLE_RESET, WEEKEVENTREG_91_10, "Saw Darmani at Grave" },
+          { CYCLE_RESET, WEEKEVENTREG_91_20, "" },
+          { CYCLE_RESET, WEEKEVENTREG_91_40, "" },
+          { CYCLE_RESET, WEEKEVENTREG_91_80, "" },
+          { PERSISTENT, PACK_WEEKEVENTREG_FLAG(92, 0x01), "Started Gorman Bros. Race Once?" },
+          { CYCLE_RESET, PACK_WEEKEVENTREG_FLAG(92, 0x02), "Finished Gorman Bros. Race" },
+          { CYCLE_RESET, PACK_WEEKEVENTREG_FLAG(92, 0x04), "" },
```
balloondude2: 92_01 is not persistent and these three flags are used together to form a 3-bit state to set/determine Gorman Bros Race states.

I would change these to SCENE_RESET and "Gorman Bros. Race State" or something.

mm/2s2h/DeveloperTools/SaveEditor.h
```diff
@@ -1,6 +1,943 @@
...
+          { CYCLE_RESET, WEEKEVENTREG_23_40, "" },
+          { PERSISTENT, WEEKEVENTREG_23_80, "Obtained Beaver Bottle" },
+          { SCENE_RESET, WEEKEVENTREG_24_01, "Won Against \"Big Brother\" Beaver" },
+          { PERSISTENT, WEEKEVENTREG_24_02, "" },
+          { SCENE_RESET, WEEKEVENTREG_24_04, "Won Against \"Little Brother\" Beaver" },
+          { CYCLE_RESET, WEEKEVENTREG_24_08, "" },
+          { CYCLE_RESET, WEEKEVENTREG_24_10, "Spoke to Goron Elder Twice" },
+          { CYCLE_RESET, WEEKEVENTREG_24_20, "" },
+          { CYCLE_RESET, WEEKEVENTREG_24_40, "Goron Elder Taught Lullaby Intro" },
+          { CYCLE_RESET, WEEKEVENTREG_24_80, "Spoke to Goron in Goron Shrine (Next to Baby, Right) Once" },
```
balloondude2: 24_80 is PERSISTENT.

mm/2s2h/DeveloperTools/SaveEditor.h
```diff
@@ -1,6 +1,943 @@
...
+          { PERSISTENT, WEEKEVENTREG_50_02, "Obtained Carnival of Time (storytime) Piece of Heart" },
+          { PERSISTENT, WEEKEVENTREG_50_04, "Obtained Four Giants (storytime) Piece of Heart" },
+          { CYCLE_RESET, WEEKEVENTREG_PROMISED_MIDNIGHT_MEETING, "Promised to Meet Anju" },
+          { CYCLE_RESET, WEEKEVENTREG_50_10, "Obtained Letter to Kafei" },
+          { CYCLE_RESET, WEEKEVENTREG_HAD_MIDNIGHT_MEETING, "Obtained Letter to Kafei" },
+          { CYCLE_RESET, WEEKEVENTREG_50_40, "" },
+          { CYCLE_RESET, WEEKEVENTREG_RECEIVED_PENDANT_OF_MEMORIES, "Spoke to Kafei After Delivering Letter" },
+          { CYCLE_RESET, WEEKEVENTREG_DELIVERED_PENDANT_OF_MEMORIES, "Gave Pendant of Memories to Anju" },
+          { CYCLE_RESET, WEEKEVENTREG_51_02, "Letter to Kafei Related?" },
+          { SCENE_RESET, WEEKEVENTREG_51_04, "Hit Kafei's Bell" },
```
balloondude2, line 457: I'd call 51_04 SCENE_RESET, but thought it was interesting that this flag is set to persist.

Edit: Not asking for a change. Just pointing out something I thought was interesting

balloondude2 (changes requested):
Left a couple comments after comparing the TCRF weekEventReg flag designation against `sPersistentCycleWeekEventRegs[]` from `z_sram_NES.c`.

mm/2s2h/DeveloperTools/SaveEditor.h
Eblo: This flag gets cleared by the Kafei actor, making the persistent designation all the more perplexing.

Eblo:
Pushed an update addressing feedback. Here's the color legend for the weekeventreg flag table:

image

mm/2s2h/BenGui/UIWidgets.cpp
```diff
@@ -1147,7 +1186,8 @@ void DrawFlagArray8Mask(const std::string& name, uint8_t& flags, Colors color) {
             }
         }
         if (ImGui::IsItemHovered()) {
-            std::string label = fmt::format("0x{:02X} ({})", bitMask, flagIndex);
+            std::string label = WrappedText(flagEntry.description, 60).c_str();
+            label += fmt::format("{}0x{:02X} ({})", label.size() ? "\n" : "", flagIndex, flagIndex);
```
balloondude2: I'd like for the label to keep the bitMask, so we get something like `0x80 (7)` instead of `0x07 (7)`

```suggestion
            label += fmt::format("{}0x{:02X} ({})", label.size() ? "\n" : "", bitMask, flagIndex);
```

balloondude2 (changes requested):
These changes are looking really nice, just one last thing.

balloondude2 (approved):

PR #1254, opened 2025-08-29
[Rando] Reset cycle collectible flags on cycle save

Eblo:
Some collectible flags persist between cycles, mainly Heart Pieces. Rando does not reset those, meaning there is no collectible flag to fire when picking up certain freestanding checks on subsequent cycles, thus resulting in no item get. This PR adds `FLAG_CYCL_SCENE_COLLECTIBLE` to the list of flags to reset on a SoT save. Quickest way to test the bug and the fix is to collect the Clock Tower Heart Piece check, play the Song of Time, and collect it again. Edit: With and without going to another scene.

I also noticed the Blast Mask check was erroneously listed as `FLAG_CYCL_SCENE_COLLECTIBLE`. I changed it to `FLAG_WEEK_EVENT_REG`.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3890998807.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3891008160.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/3891068897.zip)


mm/2s2h/Rando/MiscBehavior/OnCycleSave.cpp
```diff
@@ -87,6 +87,13 @@ void Rando::MiscBehavior::AfterEndOfCycleSave() {
                         ~(1 << (randoStaticCheck.flag & 0x1F));
                 }
                 break;
+            case FLAG_CYCL_SCENE_COLLECTIBLE:
+                // Clear the flag without triggering hook
+                if ((randoStaticCheck.flag > 0) && (randoStaticCheck.flag < 0x80)) {
+                    gPlayState->actorCtx.sceneFlags.collectible[(randoStaticCheck.flag & ~0x1F) >> 5] &=
+                        ~(1 << (randoStaticCheck.flag & 0x1F));
```
balloondude2, line 94: This will work if you reset the cycle in the same scene you collected the check, but after a transition I think you will need to clear the corresponding `gSaveContext.cycleSceneFlags` and/or `gSaveContext.save.saveInfo.permanentSceneFlags`

See the Switch Flag case above (although you don't need to worry about switch0 vs switch1 in this case)

Edit: You should also have a scene check here

balloondude2 (approved):

mm/2s2h/Rando/MiscBehavior/OnCycleSave.cpp
```diff
@@ -87,6 +87,17 @@ void Rando::MiscBehavior::AfterEndOfCycleSave() {
...
+                }
+                gSaveContext.save.saveInfo.permanentSceneFlags[randoStaticCheck.sceneId].collectible &=
+                    ~(1 << (randoStaticCheck.flag & 0x1F));
+                gSaveContext.cycleSceneFlags[randoStaticCheck.sceneId].collectible &=
+                    ~(1 << (randoStaticCheck.flag & 0x1F));
+                break;
                 // most of the others are handled by the game, with the exception of PERSISTENT_CYCLE_FLAGS_SET, not
                 // sure if any of these cases affect us yet so ignoring for now
```
balloondude2: Realized this comment isn't accurate anymore. Persistent Cycle Scene Flags include: `switch0`, `switch1`, `chest`, and `collectible`, and we now account for each one except `chest`. 

There is one persistent chest flag in the game, for the Laundry Pool. If this flag is used in game, it's not for an actual chest, and we don't use it for a check.

Issue #1017, opened 2025-02-25
Soaring from Pause Menu not disabled in all situations it should be


garrettjoecox:
While this is convenient in some cases (under water), it causes issues in others (riding the swamp boat)

The rules dictating whether or not this enhancement can be used should match the Song of Soaring's in game.

rockmasterflex69:
Being able to soar in a variety of situations via pause menu (like in water) has absolutely saved me from what would otherwise be softlocks caused by other bugs (especially FD mask anywhere-related ones).

Instead of restricting this to song of soaring's rules (which are really ocarina rules?) could we just document the only times it does cause issues and block those? or make this an option?

Eblo:
Any softlock needs to be fixed regardless of this Pause Warp topic, as the player may not have SoS in a given scenario. In the event of a true softlock, the player has the nuclear option to use `void` in the developer console. On that note, the FD anywhere swimming softlock has a pending fix, and underwater ocarina is in nightly, FYI.

> Instead of restricting this to song of soaring's rules (which are really ocarina rules?) could we just document the only times it does cause issues and block those? or make this an option?

I voiced this in the linked PR, but I'm not a big fan of that approach since it's unknown how many other scenarios won't play well with being able to warp any time the pause menu is available. Even from cursory testing, I quickly found another issue with warping from the moon. SoS rules are not synonymous with ocarina rules; there are locations in-game where one can play the SoS only to get a dismissive message that nothing happened. The moon is one area, and there are others.

I'm not firmly against the idea, but it's a harder sell to me versus the safer route of simply making Pause Warp conditions match those for SoS.

rockmasterflex69:
Very fair. Maybe I flip the thinking around a little: We want it to match the rules for SoS because theres an unknowable amount of scenarios that cause issues - solid. Maybe _also_ with an option to allow it while swimming (except.. on the moon). 

Being able to warp while in the swim state is a massive, positive convenience. 
Menu Warping as a whole is pretty game changing in a massively positive way, but doing so while swimming is like a big cherry on top of it all =).

If there's no sane way to retain that behavior (eg if _certain_ swim states are still bad scenarios... like maybe warping during the bottle races?) A "use at your own risk - enable menu warp outside SoS ruleset" 'cheat' would be neat to retain.

[...]

PR #1273, opened 2025-09-13
[Fix] Clear Pictograph State on POW

mckinlee:
Addresses #1017 
This simply clears EVENTINF_41 to prevent UI state from persisting after warp. 

#1017 states "The rules dictating whether or not this enhancement can be used should match the Song of Soaring's in game". I would prefer to address edge cases as they are reported to maintain player freedom. If a player wants to warp during the pictograph tour, I don't want to dictate they can't. I'm totally down to discuss it further though should someone disagree.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4064429130.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4064436431.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4064449445.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Songs/PauseOwlWarp.cpp
+++ b/mm/2s2h/Enhancements/Songs/PauseOwlWarp.cpp
@@ -43,6 +43,9 @@ void HandleConfirmingState(PauseContext* pauseCtx, Input* input) {
                 gHorseIsMounted = false;
             }
 
+            // Clear pictograph/camera event flag to prevent UI state from persisting after warp
+            CLEAR_EVENTINF(EVENTINF_41);
+
             Interface_SetAButtonDoAction(gPlayState, DO_ACTION_NONE);
             pauseCtx->state = PAUSE_STATE_UNPAUSE_SETUP;
             sPauseMenuVerticalOffset = -6240.0f;
--- a/mm/2s2h/Enhancements/Songs/PauseOwlWarp.cpp
+++ b/mm/2s2h/Enhancements/Songs/PauseOwlWarp.cpp
@@ -20,7 +20,8 @@ extern s32 gHorseIsMounted;
 extern "C" bool PauseOwlWarp_IsOwlWarpEnabled() {
     return CVAR && CHECK_QUEST_ITEM(QUEST_SONG_SOARING) &&
            gSaveContext.save.saveInfo.playerData.owlActivationFlags != 0 &&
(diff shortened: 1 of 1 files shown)


Eblo:
> I would prefer to address edge cases as they are reported to maintain player freedom. If a player wants to warp during the pictograph tour, I don't want to dictate they can't. I'm totally down to discuss it further though should someone disagree.

Player freedom is nice, but I don't love the idea of plugging holes as they're discovered, especially since that could be avoided by matching the vanilla conditions for whether SoS is enabled. To illustrate this point, I just found a new Pause Warp bug when testing this: you can Pause Warp on the moon, which then leads to weirdness with time. That's just from cursory testing; there could be more breakages from other edge cases.

As mentioned in the linked issue, Pause Warp *is* nice for when you're underwater, but #1091 happens to address that particular case. I'm not sure of other scenarios where Pause Warp being allowed while regular SoS is not that would be more convenient.

Eblo:
After discussion and thinking on this, particularly in the context of rando, I think I'll reconsider. There are still other situations where the player may not be able to immediately use the Ocarina but would otherwise be allowed to play the SoS.

But that said, I did find the aforementioned issue with the moon. I think Pause Owl Warp at least needs a condition to match whether the SoS itself is restricted. That should cover a few more edge cases. Would you be willing to lump that into this PR?

The restriction is aptly named. If this is 1, then Song of Soaring is not allowed:
https://github.com/HarbourMasters/2ship2harkinian/blob/558f59b06f03dd1881071f6379ee31a062a8e4ea/mm/src/code/z_message.c#L5915

mckinlee:
> After discussion and thinking on this, particularly in the context of rando, I think I'll reconsider. There are still other situations where the player may not be able to immediately use the Ocarina but would otherwise be allowed to play the SoS.
> 
> But that said, I did find the aforementioned issue with the moon. I think Pause Owl Warp at least needs a condition to match whether the SoS itself is restricted. That should cover a few more edge cases. Would you be willing to lump that into this PR?
> 
> The restriction is aptly named. If this is 1, then Song of Soaring is not allowed:
> 
> https://github.com/HarbourMasters/2ship2harkinian/blob/558f59b06f03dd1881071f6379ee31a062a8e4ea/mm/src/code/z_message.c#L5915

I like that. Added.

Eblo (approved):

balloondude2 (approved):

PR #1280, opened 2025-09-28
Port fix for soil patch Skulltula sounds

Eblo:
The big tl;dr to this one is that the actor responsible for whether the soil patches should behave as if they contain a Skulltula ultimately causes a left shift by a large value. That is undefined behavior, causing certain compiled builds of 2ship to have the soil patches always play Skulltula sounds. This fix changes the bit mask to more closely mirror how the token actor itself checks its associated flag.

Unfortunately, I could not figure out a way to produce matching non-UB code that could be shared upstream. I simply changed a macro to AND by 0xFC instead of by 0x3FC, so only the relevant bits end up making it into `Flags_GetTreasure`.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4375571371.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4375571463.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4375599248.zip)

### DIFF
--- a/mm/src/overlays/actors/ovl_Bg_Kin2_Picture/z_bg_kin2_picture.h
+++ b/mm/src/overlays/actors/ovl_Bg_Kin2_Picture/z_bg_kin2_picture.h
@@ -4,7 +4,12 @@
 #include "global.h"
 
 #define BG_KIN2_PICTURE_SKULLTULA_COLLECTED(thisx) (((thisx)->params >> 5) & 1)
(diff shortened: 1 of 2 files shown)


balloondude2 (approved):

Eblo:
Pushed a corresponding fix for the two Skull Kid paintings. The problem and fix are the same as the soil patch.

PR #1281, opened 2025-09-29
Apply 3D item drops to Twinmold fight drops

Eblo:
These were not 3D because they are handled by a different actor than EnItem00. Something to note is that these particular item drops do not appear above Link's head when collected.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4573472435.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4573529470.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4573578040.zip)


Eblo:

PR #1283, opened 2025-10-01
[Enhancement] Remote Bombchu

mckinlee:
This takes the ground work that Proxy did and updates the steering logic to rotate around the bombchu's up axis.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4241277821.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4241315183.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4241501268.zip)


mm/2s2h/Enhancements/Equipment/RemoteBombchu.cpp
```diff
@@ -0,0 +1,84 @@
+#include <libultraship/bridge.h>
```
Eblo:
```suggestion
#include "public/bridge/consolevariablebridge.h"
```

mm/2s2h/Enhancements/Equipment/RemoteBombchu.cpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+#define CVAR_NAME "gEnhancements.PlayerActions.RemoteBombchu"
+#define CVAR CVarGetInteger(CVAR_NAME, 0)
+
+static bool focused = false;
+static EnBomChu* activeBombchu = nullptr;
+
+void RegisterRemoteBombchu() {
+    COND_ID_HOOK(OnActorInit, ACTOR_EN_BOM_CHU, CVAR, [](Actor* actor) { activeBombchu = (EnBomChu*)actor; });
```
Eblo, line 33:
```suggestion
    // Link can deploy multiple bombchus at once. Only assume control of the last placed one.
    COND_ID_HOOK(OnActorInit, ACTOR_EN_BOM_CHU, CVAR, [](Actor* actor) { activeBombchu = (EnBomChu*)actor; });
```

mm/2s2h/Enhancements/Equipment/RemoteBombchu.cpp
```diff
@@ -0,0 +1,84 @@
...
+    COND_ID_HOOK(OnActorDestroy, ACTOR_EN_BOM_CHU, CVAR, [](Actor* actor) {
+        Player* player = GET_PLAYER(gPlayState);
+
+        if (actor == (Actor*)activeBombchu) {
+            activeBombchu = nullptr;
+
+            if (focused) {
+                Camera_SetFocalActor(Play_GetCamera(gPlayState, player->subCamId), &player->actor);
+                player->stateFlags1 &= ~PLAYER_STATE1_20;
+                focused = false;
```
Eblo: Since these three lines are duplicated for manually returning to Link by pressing A, I prefer they be put in a function like `void ReleaseBombchuFocus()` and called as needed.

Eblo (changes requested):
This is more fun than it has any right to be.

https://github.com/user-attachments/assets/e62705b1-f8d9-4cce-90a1-41f4a6441176

This does not seem to play well with freecam. If you use the right stick, the camera focus will shift back to Link while the bombchu steering is still active. The player can correct this by Z targeting to snap out of freecam, but I find that awkward. My first thought is to just ignore freecam while this is active, although I'm not sure how hairy that may be. Thoughts?

Eblo (approved):

PR #1297, opened 2025-10-16
Convert mundane shuffles to ObjectExtension

Eblo:
Converts mundane shuffles to use ObjectExtension (mostly ActorListIndex) in lieu of identifying checks by coordinates. The following have been converted:

- Pots
- Crates
- Barrels
- Grass
- Large Snowballs that used coords because of inconsistent ActorListIndex values
- Cows
- Freestanding (need second opinion)

Also of note is that I found two unshuffled grass checks in the Woods of Mystery, so I added those in this PR.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4463475765.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4463502511.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4463963005.zip)


Eblo:
I have converted every shuffle that I wanted, so I am marking this as ready for review. I'm looking for a second opinion on my changes to freestanding items. What I did does technically use OE, but the result is longer and more complicated than simply having a coords map. The coord-based freestanding checks are from a Rupee spawner that is only used twice across the entire game. It may be simpler to just leave that as-is. Coords are only checked once at spawn before being replaced with a custom item.

mm/src/overlays/actors/ovl_Obj_Mure3/z_obj_mure3.c
```diff
@@ -84,7 +85,7 @@ void func_8098F220(ObjMure3* this, PlayState* play) {
     yRot = this->actor.world.rot.y;
 
     for (i = 0; i < 6; i++) {
-        if (!((this->unk164 >> i) & 1)) {
+        if (GameInteractor_Should(VB_OBJ_MURE_SET_CHILD_ROOM, !((this->unk164 >> i) & 1), this, i)) {
```
garrettjoecox: these two in mure3 should instead be `VB_OBJ_MURE3_DROP_COLLECTIBLE`

mm/src/overlays/actors/ovl_Obj_Mure2/z_obj_mure2.c
```diff
@@ -142,7 +143,7 @@ void ObjMure2_SpawnChildren(ObjMure2* this, PlayState* play) {
             this->actors[i] = Actor_SpawnAsChildAndCutscene(
                 &play->actorCtx, play, sActorIds[childType], pos->x, pos->y, pos->z, this->actor.world.rot.x, 0,
                 this->actor.world.rot.z, childParams, this->actor.csId, this->actor.halfDaysBits, NULL);
-            if (this->actors[i] != NULL) {
+            if (GameInteractor_Should(VB_OBJ_MURE_SET_CHILD_ROOM, this->actors[i] != NULL, this, i)) {
```
garrettjoecox: This should probably be `VB_OBJ_MURE2_SPAWN_CHILD` and be moved up to 141

Eblo: I'm good with renaming it to match the actor, but moving it to 141 would pre-empt the child actor spawn, which I don't want. I want the actor to be spawned as normal so that I can apply the RC extension to that. This is different from Mure3, where I do want to pre-empt the spawn because it will be replaced with a custom item.

garrettjoecox: Can you use a ShouldActorInit instead, and check the parent?

Eblo: The parent argument here is NULL, so I cannot.

garrettjoecox: hmm, okay. If you moved it to 141 you could just *should = false and run the spawn yourself, but that doesn't really make things more clear/less complicated, so up to you, if here I would do the rename

Eblo: After mulling it over and considering other options, I think it's simplest just to leave it as is for now, with the rename. I don't see a benefit to moving the VB up and doing the spawn myself as far as this shuffle conversion effort goes. I think such an approach would make more sense if and when we need to do something before the spawner spawns each child.

Eblo:
Pushed a commit that documents the identification process for each mundane shuffle.

PR #1299, opened 2025-10-18
Convert to ObjectExtension

Eblo:
Replaces `ActorExtension` with SoH's `ObjectExtension`.

Pros:
- Works with any pointer type, not just actors. There is a real use case in converting grass shuffle to adopt this in lieu of identifying based on a coordinate map #1297
- Much smaller footprint for each added extension type. No need to manually track extension IDs or to allocate the initial memory
- The extended data type can be a struct that contains multiple fields

Cons:
- Only one data type per extension. The current pattern is to just define a new struct for each use case. Might feel a little verbose in cases where you just need a single new field

Other:
- Must free extension memory when done. ActorExtension already does this in the actor lifecycle, but non-actors will need to be handled in some way. Probably case-by-case

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4423030179.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4423042780.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4423130730.zip)


mm/2s2h/ObjectExtension/ObjectExtension.h
```diff
@@ -0,0 +1,117 @@
+#pragma once
+
+#ifdef __cplusplus
+
+#include <any>
+#include <cassert>
+#include <limits>
+#include <stdint.h>
+#include <unordered_map>
+#include <vector>
```
mckinlee: This vector header needed?

mckinlee (commented):
This is really well done! Just one very minor observation.

mm/2s2h/ObjectExtension/ObjectExtension.h
Eblo: Good question. I'll see when I rebase and resolve conflicts.

Eblo: Removed and built successfully on Windows. I pushed the change. If macOS and Linux build, which they should, then it was indeed not needed.

mckinlee (approved):

PR #1303, opened 2025-10-23
[Rando] Shuffle Ocarina Buttons

Caladius:
Needed a break from Enemy Souls/Drops... 

This shuffles the Ocarina Notes into the pool requiring you to collect all of the buttons for a given song before you can play it.

- Models and concept taken from SoH

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5053334115.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5053421649.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5054061721.zip)


mm/src/overlays/actors/ovl_En_Toto/z_en_toto.c
```diff
@@ -645,6 +647,7 @@ s32 func_80BA4B24(EnToto* this, PlayState* play) {
 }
 
 s32 func_80BA4C0C(EnToto* this, PlayState* play) {
+    // HERE
```
mckinlee: Whats this comment for?

mm/2s2h/Rando/DrawItem.cpp
```diff
@@ -335,8 +335,23 @@ void DrawAbilityItem(RandoItemId randoItemId, Actor* actor) {
...
+        (Gfx*)gOcarinaAButtonDL,      (Gfx*)gOcarinaCDownButtonDL, (Gfx*)gOcarinaCLeftButtonDL,
+        (Gfx*)gOcarinaCRightButtonDL, (Gfx*)gOcarinaCUpButtonDL,
+    };
+
+    OPEN_DISPS(gPlayState->state.gfxCtx);
+    Gfx_SetupDL25_Opa(gPlayState->state.gfxCtx);
+
     MATRIX_FINALIZE_AND_LOAD(POLY_XLU_DISP++, gPlayState->state.gfxCtx);
-    gSPDisplayList(POLY_XLU_DISP++, (Gfx*)abilityItemModel[randoItemId - RI_ABILITY_SWIM]);
+    gSPDisplayList(POLY_XLU_DISP++, (Gfx*)ocarinaButtonModel[randoItemId - RI_OCARINA_A_BUTTON]);
```
mckinlee: Since you're using `Gfx_SetupDL25_Opa` shouldn't this be `POLY_OPA_DISP`?

mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp
```diff
@@ -248,6 +248,17 @@ void Rando::MiscBehavior::OnFileCreate(s16 fileNum) {
                     Flags_SetRandoInf(RANDO_INF_OBTAINED_SWIM);
                 }
 
+                // Shuffle Ocarina Buttons or grant the Rando INF if not enabled
+                if (RANDO_SAVE_OPTIONS[RO_SHUFFLE_OCARINA_BUTTONS] == RO_GENERIC_YES) {
+                    for (int i = RI_OCARINA_A_BUTTON; i < RI_OCARINA_C_UP_BUTTON; i++) {
```
mckinlee: Should this be `<=` as to include `RI_OCARINA_C_UP_BUTTON`?

mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp
```diff
@@ -248,6 +248,17 @@ void Rando::MiscBehavior::OnFileCreate(s16 fileNum) {
...
+                        itemPool.push_back((RandoItemId)i);
+                    }
+                } else {
+                    for (int i = RANDO_INF_OBTAINED_OCARINA_A_BUTTON; i < RANDO_INF_OBTAINED_OCARINA_C_UP_BUTTON; i++) {
```
mckinlee: Same here, `<=`?

mm/src/overlays/actors/ovl_En_Toto/z_en_toto.c
Caladius: dont worry about it :kappa:

mm/2s2h/Rando/Types.h
```diff
@@ -2381,6 +2381,11 @@ typedef enum {
     RI_MUSHROOM,
     RI_NONE,
     RI_OCARINA,
+    RI_OCARINA_A_BUTTON,
+    RI_OCARINA_C_DOWN_BUTTON,
+    RI_OCARINA_C_LEFT_BUTTON,
+    RI_OCARINA_C_RIGHT_BUTTON,
+    RI_OCARINA_C_UP_BUTTON,
```
Eblo: Personal preference, I think it's easier to read like so:
```suggestion
    RI_OCARINA_BUTTON_A,
    RI_OCARINA_BUTTON_C_DOWN,
    RI_OCARINA_BUTTON_C_LEFT,
    RI_OCARINA_BUTTON_C_RIGHT,
    RI_OCARINA_BUTTON_C_UP,
```

mm/2s2h/Rando/Types.h
```diff
@@ -2870,6 +2876,11 @@ typedef enum {
     RANDO_INF_OBTAINED_SOUL_OF_ODOLWA,
     RANDO_INF_OBTAINED_SOUL_OF_TWINMOLD,
     RANDO_INF_OBTAINED_SWIM,
+    RANDO_INF_OBTAINED_OCARINA_A_BUTTON,
+    RANDO_INF_OBTAINED_OCARINA_C_DOWN_BUTTON,
+    RANDO_INF_OBTAINED_OCARINA_C_LEFT_BUTTON,
+    RANDO_INF_OBTAINED_OCARINA_C_RIGHT_BUTTON,
+    RANDO_INF_OBTAINED_OCARINA_C_UP_BUTTON,
```
Eblo:
```suggestion
    RANDO_INF_OBTAINED_OCARINA_BUTTON_A,
    RANDO_INF_OBTAINED_OCARINA_BUTTON_C_DOWN,
    RANDO_INF_OBTAINED_OCARINA_BUTTON_C_LEFT,
    RANDO_INF_OBTAINED_OCARINA_BUTTON_C_RIGHT,
    RANDO_INF_OBTAINED_OCARINA_BUTTON_C_UP,
```

Eblo (changes requested):
Sweeping question: why the approach of evaluating at song's end whether the song is valid, along with ad hoc VBs to prevent the player from engaging with a song check if they don't have the needed notes? I think a cleaner and saner approach would be to prevent an individual note from being registered if the button has not been obtained. As-is, I think it's somewhat confusing that Link can play the notes, only for it to fail when playing the complete song or playing for a key NPC if the relevant buttons are not acquired.

For example, the player can still play any notes for Pierre or the Termina wall regardless of buttons obtained. I realize the Termina wall drops aren't shuffled now, but they might be. You could add even more VBs for those, but if you nipped this at the per-button input level, you wouldn't need to worry about every single case.

There needs to be more logic accounting for things besides core songs. While you did define whether a song is playable, not use of the Ocarina in logic was tied to a specific song. For instance, there's no logic preventing Evan from holding one of the buttons used for the jam song. Ditto Ballad of the Wind Fish.

Scarecrow's Song is a unique case where, iirc, the player needs to have at least two different notes to make a song. We have an enhancement to skip that, but it's opt-in. This shuffle now makes that distinction significant.

Caladius:
> Sweeping question: why the approach of evaluating at song's end whether the song is valid, along with ad hoc VBs to prevent the player from engaging with a song check if they don't have the needed notes? I think a cleaner and saner approach would be to prevent an individual note from being registered if the button has not been obtained. As-is, I think it's somewhat confusing that Link can play the notes, only for it to fail when playing the complete song or playing for a key NPC if the relevant buttons are not acquired.
> 
> For example, the player can still play any notes for Pierre or the Termina wall regardless of buttons obtained. I realize the Termina wall drops aren't shuffled now, but they might be. You could add even more VBs for those, but if you nipped this at the per-button input level, you wouldn't need to worry about every single case.
> 
> There needs to be more logic accounting for things besides core songs. While you did define whether a song is playable, not use of the Ocarina in logic was tied to a specific song. For instance, there's no logic preventing Evan from holding one of the buttons used for the jam song. Ditto Ballad of the Wind Fish.
> 
> Scarecrow's Song is a unique case where, iirc, the player needs to have at least two different notes to make a song. We have an enhancement to skip that, but it's opt-in. This shuffle now makes that distinction significant.

Have a place in mind for this? The approach, whether sane or not haha, was just where it worked out.

Eblo:
The Ocarina button index corresponds to this enum regardless of the input used:

https://github.com/HarbourMasters/2ship2harkinian/blob/fefee554a733f06dc60272521c19fdc76fdcc702/mm/include/z64ocarina.h#L166-L174

This point in code_8019AF00.c seems convenient to add a new VB should:

https://github.com/HarbourMasters/2ship2harkinian/blob/fefee554a733f06dc60272521c19fdc76fdcc702/mm/src/audio/code_8019AF00.c#L2669

It comes after the base ocarina button determination has been made across input types, but before anything is done with it. I'm thinking we could wrap it like so:

```c++
 if (GameInteractor_Should(VB_PLAY_OCARINA_NOTE, sOcarinaInputButtonCur, &sCurOcarinaButtonIndex, &sCurOcarinaPitch)) {} 
```

Then in the enhancement code, do something like:

```c++
COND_VB_SHOULD(VB_PLAY_OCARINA_NOTE, IS_RANDO, {
    u8* sCurOcarinaButtonIndex = va_arg(args, u8*);
    u8* sCurOcarinaPitch = va_arg(args, u8*);
    bool canPlayNote = true;

    if(*sCurOcarinaButtonIndex == OCARINA_BTN_A && !Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_A)) {
        canPlayNote = false;
    }

    // Repeat the above for all other buttons. Might be cleaner to use a map or macro for brevity

    // Player pressed a button they do not have, so clear the ocarina button state
    if(!canPlayNote) {
        *sCurOcarinaButtonIndex = OCARINA_BTN_INVALID;
        *sCurOcarinaPitch = OCARINA_PITCH_NONE;
    }
});
```

That should catch things at the per-note level and prevent Link from even playing a single note if he doesn't have it.

Caladius:
Adding a note here to add this to the Item Tracker once merged in. https://github.com/HarbourMasters/2ship2harkinian/pull/1333

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTracker.cpp
```diff
@@ -431,6 +438,7 @@ void ItemTrackerWindow::Draw() {
     uint32_t windowIndex = TRACKER_MAIN;
     for (auto* window : itemTrackerWindows) {
         if (window->empty()) {
+            windowIndex++;
```
Eblo: Will be a conflict since I addressed this bug here: #1366

mm/2s2h/Rando/ActorBehavior/Player.cpp
```diff
@@ -30,4 +39,20 @@ void Rando::ActorBehavior::InitPlayerBehavior() {
...
+        u8* sCurOcarinaButtonIndex = va_arg(args, u8*);
+        u8* sCurOcarinaPitch = va_arg(args, u8*);
+        bool canPlayNote = true;
+
+        auto findNote = ocarinaButtonToFlagMap.find(static_cast<OcarinaButtonIndex>(*sCurOcarinaButtonIndex));
+        if (findNote != ocarinaButtonToFlagMap.end()) {
+            canPlayNote = Flags_GetRandoInf(findNote->second);
+        }
+
+        if (!canPlayNote) {
```
Eblo: Is this map really needed? You should be able to just do some arithmetic:

```suggestion
        if(!Flags_GetRandoInf(((*sCurOcarinaButtonIndex) - OCARINA_BTN_A) + RANDO_INF_OBTAINED_OCARINA_BUTTON_A)) {
```

mm/2s2h/Rando/ActorBehavior/Player.cpp
```diff
@@ -30,4 +39,20 @@ void Rando::ActorBehavior::InitPlayerBehavior() {
             RespawnOnWaterTouch(GET_PLAYER(gPlayState));
         }
     });
+    COND_VB_SHOULD(VB_PLAY_OCARINA_NOTE, IS_RANDO, {
```
Eblo:
```suggestion

    COND_VB_SHOULD(VB_PLAY_OCARINA_NOTE, IS_RANDO, {
```

mm/2s2h/Rando/ActorBehavior/Player.cpp
```diff
@@ -30,4 +39,20 @@ void Rando::ActorBehavior::InitPlayerBehavior() {
...
+
+        auto findNote = ocarinaButtonToFlagMap.find(static_cast<OcarinaButtonIndex>(*sCurOcarinaButtonIndex));
+        if (findNote != ocarinaButtonToFlagMap.end()) {
+            canPlayNote = Flags_GetRandoInf(findNote->second);
+        }
+
+        if (!canPlayNote) {
+            *sCurOcarinaButtonIndex = OCARINA_BTN_INVALID;
+            *sCurOcarinaPitch = OCARINA_PITCH_NONE;
+            Audio_PlaySfx(NA_SE_SY_OCARINA_ERROR);
```
Eblo: The feedback is nice, but because the Ocarina is programmed to play a note continuously as long as a button is held, this results in this sfx playing every frame, which is jarring.

I don't know if this is the best answer, but you could have a debounce within this enhancement file and only play the sfx if enough time has elapsed since the last call.

mm/2s2h/Rando/Logic/Logic.h
```diff
@@ -136,6 +138,73 @@ inline std::string LogicString(std::string condition) {
...
+        case OCARINA_SONG_SUNS:
+            return (Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_RIGHT) &&
+                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_DOWN) &&
+                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_UP));
+        case OCARINA_SONG_WIND_FISH_DEKU:
+            return (Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_RIGHT) &&
+                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_A) &&
+                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_DOWN) &&
+                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_LEFT));
+        // case OCARINA_SONG_SCARECROW_SPAWN:
```
Eblo: Scarecrow requires you have any two different notes. That might be one ugly condition.

mm/2s2h/Rando/Logic/Logic.h
```diff
@@ -136,6 +138,73 @@ inline std::string LogicString(std::string condition) {
...
+                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_RIGHT) &&
+                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_LEFT));
+        case OCARINA_SONG_NEW_WAVE:
+        case OCARINA_SONG_ELEGY:
+            return (Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_LEFT) &&
+                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_UP) &&
+                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_RIGHT) &&
+                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_DOWN));
+        case OCARINA_SONG_OATH:
+        case OCARINA_SONG_WIND_FISH_ZORA:
```
Eblo, line 306: The Zora ballad piece requires C Left.

^ > v A ^ < > v A

mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp
```diff
@@ -60,7 +60,7 @@ void Rando::MiscBehavior::OnFileCreate(s16 fileNum) {
                 }
 
                 std::vector<RandoItemId> startingItems = convertStartingItemsToRandoItemId(
-                    CVarGetString("gRando.StartingItems", RANDO_STARTING_ITEMS_DEFAULT), ",");
+                    CVarGetString("gRando.StartingItems", RANDO_STARTING_ITEMS_DEFAULT.c_str()), ",");
```
Eblo: It seems like this was done to clarify the starting item default values. Will this at all impact #1367?

mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp
```diff
@@ -253,6 +253,18 @@ void Rando::MiscBehavior::OnFileCreate(s16 fileNum) {
                     Flags_SetRandoInf(RANDO_INF_OBTAINED_SWIM);
                 }
 
+                // Shuffle Ocarina Buttons or grant the Rando INF if not enabled
+                if (RANDO_SAVE_OPTIONS[RO_SHUFFLE_OCARINA_BUTTONS] == RO_GENERIC_YES) {
+                    for (int i = RI_OCARINA_BUTTON_A; i <= RI_OCARINA_BUTTON_C_UP; i++) {
+                        itemPool.push_back((RandoItemId)i);
```
Eblo:
```suggestion
                    for (RandoItemId i = RI_OCARINA_BUTTON_A; i <= RI_OCARINA_BUTTON_C_UP; i++) {
                        itemPool.push_back(i);
```

mm/2s2h/Rando/DrawItem.cpp
```diff
@@ -395,8 +395,23 @@ void DrawAbilityItem(RandoItemId randoItemId, Actor* actor) {
 
     Gfx_SetupDL25_Xlu(gPlayState->state.gfxCtx);
 
-    MATRIX_FINALIZE_AND_LOAD(POLY_XLU_DISP++, gPlayState->state.gfxCtx);
-    gSPDisplayList(POLY_XLU_DISP++, (Gfx*)abilityItemModel[randoItemId - RI_ABILITY_SWIM]);
+    MATRIX_FINALIZE_AND_LOAD(POLY_OPA_DISP++, gPlayState->state.gfxCtx);
+    gSPDisplayList(POLY_OPA_DISP++, (Gfx*)abilityItemModel[randoItemId - RI_ABILITY_SWIM]);
```
Eblo: Did you mean to change these from XLU to OPA? Ocarina buttons do not seem to use this. The setup function before this is still XLU.

mm/2s2h/Rando/Menu.cpp
```diff
@@ -280,6 +281,12 @@ static void DrawItemsTab() {
                                                 "into deep water will respawn Link.",
                                      .disabled = IncompatibleWithLogicSetting(RO_SHUFFLE_SWIM),
                                      .disabledTooltip = "Incompatible with current Logic Setting" } }));
+    CVarCheckbox("Shuffle Ocarina Buttons", Rando::StaticData::Options[RO_SHUFFLE_OCARINA_BUTTONS].cvar,
+                 CheckboxOptions({ { .tooltip = "Shuffles the Buttons used to play Ocarina Notes,\n"
+                                                "you will be unable to play a song until you find all\n"
```
Eblo:
```suggestion
                 CheckboxOptions({ { .tooltip = "Shuffles the Buttons used to play Ocarina Notes.\n"
                                                "You will be unable to play a song until you find all\n"
```

mm/2s2h/Rando/Logic/Logic.h
```diff
@@ -62,7 +62,9 @@ extern std::map<RandoRegionId, RandoRegion> Regions;
 #define CAN_USE_HUMAN_SWORD (GET_CUR_EQUIP_VALUE(EQUIP_TYPE_SWORD) >= EQUIP_VALUE_SWORD_KOKIRI)
 #define CAN_USE_SWORD (CAN_USE_HUMAN_SWORD || HAS_ITEM(ITEM_SWORD_GREAT_FAIRY) || CAN_BE_DEITY)
 // Be careful here, as some checks require you to play the song as a specific form
-#define CAN_PLAY_SONG(song) (HAS_ITEM(ITEM_OCARINA_OF_TIME) && CHECK_QUEST_ITEM(QUEST_SONG_##song))
+#define CAN_PLAY_SONG(song)                                                   \
+    (HAS_ITEM(ITEM_OCARINA_OF_TIME) && CHECK_QUEST_ITEM(QUEST_SONG_##song) && \
+     Rando::Logic::canPlaySong(Ship_GetOcarinaSongByQuestId(QUEST_SONG_##song)))
```
Eblo: Since only a valid QUEST_SONG can exist in this, I don't think the associated function and map are needed. This can be handled like CAN_USE_ABILITY:

```suggestion
     Rando::Logic::canPlaySong((QUEST_SONG_##song - QUEST_SONG_SONATA) + OCARINA_SONG_SONATA))
```

mm/src/overlays/actors/ovl_En_Zob/z_en_zob.c
```diff
@@ -5,6 +5,7 @@
  */
 
 #include "z_en_zob.h"
+#include "GameInteractor/GameInteractor.h"
```
Eblo: Seems unused. I assume the per-button solution obsoleted whatever you might have planned here? At any rate, Japas doesn't lock the player in like Toto, so nothing special needs done for him.
```suggestion
```

mm/2s2h/Rando/ActorBehavior/EnToto.cpp
```diff
@@ -38,4 +39,19 @@ void Rando::ActorBehavior::InitEnTotoBehavior() {
...
         Message_BombersNotebookQueueEvent(gPlayState, BOMBERS_NOTEBOOK_EVENT_MET_GORMAN);
     });
+
+    COND_VB_SHOULD(VB_TOTO_START_SOUND_CHECK, IS_RANDO && RANDO_SAVE_OPTIONS[RO_SHUFFLE_OCARINA_BUTTONS], {
+        EnToto* totoActor = va_arg(args, EnToto*);
+        if (totoActor->text->textId == 0x2B24) {
+            if (!(Rando::Logic::canPlaySong(OCARINA_SONG_WIND_FISH_HUMAN) ||
+                  Rando::Logic::canPlaySong(OCARINA_SONG_WIND_FISH_DEKU) ||
+                  Rando::Logic::canPlaySong(OCARINA_SONG_WIND_FISH_GORON) ||
+                  Rando::Logic::canPlaySong(OCARINA_SONG_WIND_FISH_ZORA))) {
```
Eblo: One problem with this approach is if the player can play at least one of the songs, but goes into the performance as a form for which they do not have the needed notes. The lockout will still happen. Since you cannot get the check without playing all parts anyway, maybe we just AND these instead and call it a day? This would also prevent the skip ballad enhancement from letting the player cheat through performances they cannot do.

```suggestion
            if (!(Rando::Logic::canPlaySong(OCARINA_SONG_WIND_FISH_HUMAN) &&
                  Rando::Logic::canPlaySong(OCARINA_SONG_WIND_FISH_DEKU) &&
                  Rando::Logic::canPlaySong(OCARINA_SONG_WIND_FISH_GORON) &&
                  Rando::Logic::canPlaySong(OCARINA_SONG_WIND_FISH_ZORA))) {
```

Eblo (changes requested):
One thing we have to come up with an answer for is the Scarecrow's Song. There is an enhancement to automatically summon the Scarecrow if the player brings out the Ocarina in range, bypassing the need to play songs. The song requires at least two different notes. I think what we can do is go into `RegisterSkipScarecrowSong` and tack on `canPlaySong(OCARINA_SONG_SCARECROW_SPAWN)` for rando, thus only allowing this enhancement to work if the player *could* play the Scarecrow's Song.

Malkierian:
I think SoH handles scarecrow very well, where anytime it checks for the ability to play, it includes a counter for adding each of the available buttons based on set flags, and if available >= 2, then scarecrow can be played. This can be handled in a helper function.

Eblo:
Shipwright's answer for reference:

https://github.com/HarbourMasters/Shipwright/blob/da6cf439d68bf9ab6a2e6bd4a527a9f62491edd1/soh/soh/Enhancements/randomizer/logic.cpp#L1063-L1066

https://github.com/HarbourMasters/Shipwright/blob/da6cf439d68bf9ab6a2e6bd4a527a9f62491edd1/soh/soh/Enhancements/randomizer/logic.cpp#L1145-L1148

mm/2s2h/Rando/DrawItem.cpp
Caladius: This is for Swim, but no probably not haha.

mm/2s2h/Rando/ActorBehavior/Player.cpp
Caladius: Maybe a hook for player inputs that flips a boolean once on press, once on release so the note only plays once and doesnt reset until that button is released?

Eblo: The debounce might actually be easier. I'm thinking you could just:
```c++
#define DEBOUNCE_OCARINA 500
uint64_t ocarinaNoteDebounce = 0;
// ...
        if (!canPlayNote) {
            *sCurOcarinaButtonIndex = OCARINA_BTN_INVALID;
            *sCurOcarinaPitch = OCARINA_PITCH_NONE;
            uint64_t currentTime = GetUnixTimestamp();
            if (currentTime > ocarinaNoteDebounce + DEBOUNCE_OCARINA) {
                Audio_PlaySfx(NA_SE_SY_OCARINA_ERROR);
                ocarinaNoteDebounce = currentTime + DEBOUNCE_OCARINA;
            }
```

Caladius: I was thinking this, but your call:

```cpp
COND_ID_HOOK(OnActorUpdate, ACTOR_PLAYER, shouldPlayerRegister, [](Actor* actor) {
    Player* player = GET_PLAYER(gPlayState);
    if (!Flags_GetRandoInf(RANDO_INF_OBTAINED_SWIM)) {
        RespawnOnWaterTouch(player);
    }

    if (player->stateFlags2 & PLAYER_STATE2_USING_OCARINA) {
        Input* input = CONTROLLER1(&gPlayState->state);

        if (CHECK_BTN_ANY(input->press.button, BTN_A | BTN_CDOWN | BTN_CLEFT | BTN_CRIGHT | BTN_CUP)) {
            if (lastOcarinaButton != input->press.button) {
                lastOcarinaButton = input->press.button;
                playErrorChime = true;
            }
        }
    }
});
```

Eblo: Go with that.

mm/2s2h/Rando/Logic/Logic.h
```diff
@@ -198,8 +209,9 @@ inline bool canPlaySong(u8 songId) {
                     Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_A) &&
                     Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_DOWN) &&
                     Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_LEFT));
-        // case OCARINA_SONG_SCARECROW_SPAWN:
-        //    break;
+        case OCARINA_SONG_SCARECROW_SPAWN:
+            return FoundOcarinaButtons() >= 2;
+            break;
```
Eblo:
```suggestion
```

mm/2s2h/Rando/Logic/Logic.h
```diff
@@ -138,6 +139,16 @@ inline std::string LogicString(std::string condition) {
     return condition;
 }
 
+inline uint8_t FoundOcarinaButtons() {
+    uint8_t foundButtons = 0;
+    for (int i = RANDO_INF_OBTAINED_OCARINA_BUTTON_A; i < RANDO_INF_OBTAINED_OCARINA_BUTTON_C_UP; i++) {
```
Eblo: This has to be inclusive. If I have two buttons and one of them is C Up, the scarecrow will not spawn.
```suggestion
    for (int i = RANDO_INF_OBTAINED_OCARINA_BUTTON_A; i <= RANDO_INF_OBTAINED_OCARINA_BUTTON_C_UP; i++) {
```

mm/2s2h/Enhancements/Songs/SkipScarecrowSong.cpp
```diff
@@ -21,7 +22,16 @@ void RegisterSkipScarecrowSong() {
...
-            *should = true;
+
+            // In Rando we may utilize Ocarina Buttons, ensure this is honored.
+            if (IS_RANDO) {
+                if (Rando::Logic::canPlaySong(OCARINA_SONG_SCARECROW_SPAWN)) {
+                    *should = true;
+                }
+            } else {
+                *should = true;
+            }
```
Eblo: This should bail early and *not* set should to true if the player has insufficient notes. As it is now, if the player uses the Ocarina in range of the scarecrow but lacks sufficient notes, they will be soft locked in the Ocarina playing state.
```suggestion
            // In Rando we may utilize Ocarina Buttons, ensure this is honored.
            if (IS_RANDO && Rando::Logic::canPlaySong(OCARINA_SONG_SCARECROW_SPAWN)) {
                    return;
            }

            *should = true;
```

mm/2s2h/Rando/ActorBehavior/Player.cpp
```diff
@@ -26,9 +29,24 @@ void RespawnOnWaterTouch(Player* player) {
...
+        Player* player = GET_PLAYER(gPlayState);
         if (!Flags_GetRandoInf(RANDO_INF_OBTAINED_SWIM)) {
-            RespawnOnWaterTouch(GET_PLAYER(gPlayState));
+            RespawnOnWaterTouch(player);
+        }
+
+        if (player->stateFlags2 & PLAYER_STATE2_USING_OCARINA) {
+            Input* input = CONTROLLER1(&gPlayState->state);
+
+            if (CHECK_BTN_ANY(input->press.button, BTN_A | BTN_CDOWN | BTN_CLEFT | BTN_CRIGHT | BTN_CUP)) {
```
Eblo: This directly checks the input source, meaning only the actual A and C buttons cause this behavior, which then means that the dpad and right stick do not play error chimes even if they trigger an unavailable note. I also notice that, sometimes when I do have a note, the correct note and then the error chime will both play.

I think the better approach would be to move this logic into `VB_PLAY_OCARINA_NOTE` and then track `sCurOcarinaButtonIndex`, comparing against `OCARINA_BTN_*` values (defaulting at `OCARINA_BTN_INVALID`). At that point you also would not need to track `playErrorChime` in a static variable, as you'd already have all the information you need at that scope.

mm/2s2h/Rando/DrawItem.cpp
Eblo: My point was moreso that this change does not seem relevant to this PR, as it affects ability items (only swim at this point) and not Ocarina Buttons. The flippers look the same either way, so I'm not sure why this change is here.

Eblo (changes requested):

mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp
Caladius: I dont know that the Locations Tab has any reference to Starting Items.

mm/2s2h/Rando/ActorBehavior/Player.cpp
Caladius: Im fine with moving it but the whole point of the static Bool was to prevent the note from repeatedly playing the error chime.

mm/2s2h/Enhancements/Songs/SkipScarecrowSong.cpp
```diff
@@ -21,7 +22,13 @@ void RegisterSkipScarecrowSong() {
          */
         if ((enKakasi->picto.actor.xzDistToPlayer < enKakasi->songSummonDist) &&
             ((BREG(1) != 0) || (gPlayState->msgCtx.ocarinaMode == OCARINA_MODE_ACTIVE))) {
+            // In Rando we may utilize Ocarina Buttons, ensure this is honored.
+            if (IS_RANDO && Rando::Logic::canPlaySong(OCARINA_SONG_SCARECROW_SPAWN)) {
```
Eblo: Oops, my suggestion did the opposite and made Pierre spawn if you don't have enough notes, and he won't spawn if you do. Minor fix:

```suggestion
            if (IS_RANDO && !Rando::Logic::canPlaySong(OCARINA_SONG_SCARECROW_SPAWN)) {
```

mm/2s2h/Rando/ActorBehavior/Player.cpp
```diff
@@ -25,9 +28,30 @@ void RespawnOnWaterTouch(Player* player) {
...
+            RespawnOnWaterTouch(player);
+        }
+    });
+
+    COND_VB_SHOULD(VB_PLAY_OCARINA_NOTE, IS_RANDO, {
+        u8* sCurOcarinaButtonIndex = va_arg(args, u8*);
+        u8* sCurOcarinaPitch = va_arg(args, u8*);
+        u8 currentOcarinaButton = *sCurOcarinaButtonIndex;
+
+        if (!Flags_GetRandoInf(((*sCurOcarinaButtonIndex) - OCARINA_BTN_A) + RANDO_INF_OBTAINED_OCARINA_BUTTON_A)) {
```
Eblo, line 43: Well, this is unfortunate. The button indices in the source code do not line up with the RI enum. C-Right and C-Left are swapped, meaning obtaining "C-Right" will actually yield C-Left and vice versa. Our options are to do a map lookup, or swap the RI order to match source. Maybe there's some clever naming we could do to maintain both alphabetical order and the intended button order? `RANDO_INF_OBTAINED_OCARINA_BUTTON_C1_RIGHT`, `RANDO_INF_OBTAINED_OCARINA_BUTTON_C2_LEFT`? A map lookup is not desirable if it can be helped.

mm/2s2h/Rando/Logic/Logic.h
```diff
@@ -138,6 +141,83 @@ inline std::string LogicString(std::string condition) {
...
+        case OCARINA_SONG_NEW_WAVE:
+        case OCARINA_SONG_ELEGY:
+            return (Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_LEFT) &&
+                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_UP) &&
+                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_RIGHT) &&
+                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_DOWN));
+        case OCARINA_SONG_OATH:
+        case OCARINA_SONG_WIND_FISH_ZORA:
+            return (Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_RIGHT) &&
+                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_DOWN) &&
```
Eblo, line 309:
```suggestion
                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_DOWN) &&
                    Flags_GetRandoInf(RANDO_INF_OBTAINED_OCARINA_BUTTON_C_LEFT) &&
```

Eblo (changes requested):
With the reorganization of VBs, this will need rebased to define and document the new VBs in `GameInteractor_VanillaBehavior.h`.

Eblo (approved):

PR #1308, opened 2025-10-26
Save Editor Minor Redesign

mckinlee:
**New UI Components**
- Card Layout: Masonry-style multi-column layout (auto-fills shortest column)
- Searchable Combobox: Filter-enabled dropdown with auto-focus
- Combobox Width Override: Optional width parameter added

**Save Editor Redesign**
- General Tab: Reorganized into 6 cards (Identity, Time, Currency, Health, Magic, Temple Progress). Time skip buttons scale evenly, banked rupees moved to currency card
- Temple Clears: 2-column layout (cleared | open/inverted)
- Player Tab: Added "Current Items" card with all equipped slots + Clear All button. Swapped speed/location card positions
- Reg Editor: Changed to 2-column table (index | slider)
- Flags/Rando: Scene and reward dropdowns now searchable
- Polish: Theme-based tab styling, extra padding on dungeon tab

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4662186495.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4663312491.zip)


mm/2s2h/BenGui/UIWidgets.hpp
```diff
@@ -630,7 +655,7 @@ bool Combobox(const char* label, T* value, const std::unordered_map<T, const cha
     ImGui::BeginGroup();
     ImGui::BeginDisabled(options.disabled);
     PushStyleCombobox(options.color);
-
+
```
Eblo:
```suggestion

```

mm/2s2h/BenGui/UIWidgets.hpp
```diff
@@ -642,7 +667,7 @@ bool Combobox(const char* label, T* value, const std::unordered_map<T, const cha
         }
     }
     float comboWidth = CalcComboWidth(longest, options.flags);
-
+
```
Eblo:
```suggestion

```

mm/2s2h/BenGui/UIWidgets.hpp
```diff
@@ -661,7 +686,7 @@ bool Combobox(const char* label, T* value, const std::unordered_map<T, const cha
             }
         }
     }
-
+
```
Eblo:
```suggestion
```

mm/2s2h/BenGui/UIWidgets.hpp
```diff
@@ -676,7 +701,7 @@ bool Combobox(const char* label, T* value, const std::unordered_map<T, const cha
         ImGui::PopStyleVar();
         ImGui::EndCombo();
     }
-
+
```
Eblo:
```suggestion
```

mm/2s2h/DeveloperTools/SaveEditor.cpp
```diff
@@ -1390,6 +1401,17 @@ void GetPlayerForm(uint32_t form) {
     }
 }
 
+void ClearAllEquippedItems() {
+    CUR_FORM_EQUIP(0) = ITEM_NONE;
```
Eblo: Whichever one of these is B will break if used on a transformation. Goron/Deku/Zora will have their B button action replaced with holding an item. FD will lose his sword, which results in human losing his sword because they share inventory slots. This lasts until turning into human and back into FD. It also seems like calling this while in a transformation will stop short of actually clearing every equip slot.

Eblo (changes requested):
I do like the cards and how easy they are on the eyes, but the lack of scrolling presents issues that are not present with the less pretty editor we have now.

These flag labels are cut off:

Image

Quite a bit is cut off here:

Image

The two unused songs are out of view here unless I give the save editor a lot of width:

Image

The small keys slider seems strangely narrow compared to before. It doesn't seem like you directly did anything, so it may be inheriting some property by being in a card:

Image

There may be more than that which I haven't caught. I wish I had any bright ideas on reconciling the lack of scrolling with the pretty cards, but I'm afraid I don't.

The combobox searching is very nice, no notes there.

mm/2s2h/DeveloperTools/SaveEditor.cpp
```diff
@@ -1186,12 +1185,22 @@ void DrawQuestStatusTab() {
...
-    }
-    for (int32_t i = QUEST_SONG_SONATA; i <= QUEST_SONG_SARIA; i++) {
-        DrawSong((QuestItem)i);
-    }
+    auto drawSongRange = [](int32_t start, int32_t end) {
+        for (int32_t i = start; i <= end; i++) {
+            float lineStartX = ImGui::GetCursorPosX();
+            ImVec2 itemSize = DrawSong(static_cast<QuestItem>(i));
+            bool isLastItem = (i == end);
+            if (!isLastItem) {
```
Eblo:
```suggestion
            if (i != end) {
```

mm/2s2h/DeveloperTools/SaveEditor.cpp
```diff
@@ -1186,12 +1185,22 @@ void DrawQuestStatusTab() {
...
+                float nextItemMaxX = lineStartX + itemSize.x + ImGui::GetStyle().ItemSpacing.x + itemSize.x;
```
Eblo: I think this might be causing the items to have a greater width than before, which makes it easier to bump the latter items down to another row. Does `itemSize.x` need to be added twice? That might be the problem.

image

mm/2s2h/DeveloperTools/SaveEditor.cpp
```diff
@@ -1402,14 +1413,40 @@ void GetPlayerForm(uint32_t form) {
...
-    DPAD_GET_CUR_FORM_BTN_ITEM(EQUIP_SLOT_D_LEFT) = ITEM_NONE;
-    DPAD_GET_CUR_FORM_BTN_ITEM(EQUIP_SLOT_D_DOWN) = ITEM_NONE;
-    DPAD_GET_CUR_FORM_BTN_ITEM(EQUIP_SLOT_D_UP) = ITEM_NONE;
+    auto& buttonItems = gSaveContext.save.saveInfo.equips.buttonItems;
+    auto& cButtonSlots = gSaveContext.save.saveInfo.equips.cButtonSlots;
+
+    for (size_t form = 0; form < ARRAY_COUNT(buttonItems); form++) {
+        for (size_t slot = 0; slot < ARRAY_COUNT(buttonItems[form]); slot++) {
+            if (slot == EQUIP_SLOT_B) {
+                if (form != 0) {
```
Eblo: A problem here is that human and FD actually share the same slots. FD *is* form 0, but then human is coded to look at the slots for form 0 (see `CUR_FORM` and `PlayerTransformation`), causing them to share slots. These factors culminate in this code still clearing the B button for FD, leaving him with an invisible sword that I imagine is the Kokiri Sword.

This very problem is why rando has to specially check for FD when acquiring progressive sword pickups. Look for uses of `if (GET_PLAYER_FORM != PLAYER_FORM_FIERCE_DEITY)`, which includes what the current save editor does.

Eblo (changes requested):
I see most of the wrapping issues are taken care of. I'm still not sure what's causing the fairy/key slider to be so narrow.

mckinlee:
> I see most of the wrapping issues are taken care of. I'm still not sure what's causing the fairy/key slider to be so narrow.

Looks like I failed to add `PushItemWidth` and `PopItemWidth` around the slider like I did for small keys. Should be easy fix... I hope lol

mm/2s2h/DeveloperTools/SaveEditor.cpp
mckinlee: Looks like after 'DrawSong' is called, the cursor has already moved to the end of the drawn item. I'll use the current cursor position instead of recalculating from `lineStartX`. That will hopefully fix it.

mm/2s2h/DeveloperTools/SaveEditor.cpp
```diff
@@ -1390,6 +1413,49 @@ void GetPlayerForm(uint32_t form) {
...
+                    // Goron/Deku/Zora expect their innate B actions; leave untouched.
+                    continue;
+                }
+
+                int swordValue = GET_CUR_EQUIP_VALUE(EQUIP_TYPE_SWORD);
+                if (swordValue == EQUIP_VALUE_SWORD_NONE) {
+                    swordValue = EQUIP_VALUE_SWORD_KOKIRI;
+                }
+
+                buttonItems[form][slot] = ITEM_SWORD_KOKIRI + swordValue - EQUIP_VALUE_SWORD_KOKIRI;
```
Eblo: Just so I understand this, the intent is to keep the sword if it exists, but give one if it does not? Is this to handle cases where the B button is something else, like with the Blast Mask? I mostly don't understand why there's the edge case of giving a sword if there isn't one, as opposed to leaving it be.

mm/2s2h/DeveloperTools/SaveEditor.cpp
```diff
@@ -1390,6 +1413,49 @@ void GetPlayerForm(uint32_t form) {
...
+        for (size_t slot = 0; slot < ARRAY_COUNT(buttonItems[form]); slot++) {
+            if (slot == EQUIP_SLOT_B) {
+                // Form 0 is shared by FD and Human via CUR_FORM
+                // FD uses Deity Sword (not in B slot), Human uses regular swords
+                if (form == 0 && GET_PLAYER_FORM == PLAYER_FORM_FIERCE_DEITY) {
+                    // FD expects innate B action; leave untouched
+                    continue;
+                }
+                if (form != 0) {
+                    // Goron/Deku/Zora expect their innate B actions; leave untouched.
```
Eblo: Although the conditions are different, the reasoning and goal are the same, so let's condense this a little bit:
```suggestion
                // Goron/Deku/Zora/FD expect their innate B actions; leave untouched.
                // Form 0 is shared by FD and Human via CUR_FORM
                if (form != 0 || GET_PLAYER_FORM == PLAYER_FORM_FIERCE_DEITY) {
```

Eblo (changes requested):
I see the changes, but the key and fairy sliders still look squished to me. Here's a direct comparison:

develop:
Image

This PR:
Image

mm/2s2h/DeveloperTools/SaveEditor.cpp
mckinlee: No, you're right. This was a oversight on my part. The intent was for the Clear All button to set B to the players default sword OR ITEM_NONE if they do not have one. I'll add a else statement to do so shortly.

Eblo (approved):

PR #1310, opened 2025-10-26
Port Display List Viewer

mckinlee:
As the title says. The tool supports viewing various graphics commands (textures, vertices, colors, etc.) and allows live editing of color-related commands.

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4654470949.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4654501860.zip)


mm/2s2h/DeveloperTools/DLViewer.cpp
```diff
@@ -0,0 +1,351 @@
...
+#include "DisplayList.h"
+#include "BenPort.h"
+
+#include <array>
+#include <bit>
+#include <map>
+#include <string>
+#include <cstdint>
+#include <algorithm>
+#include <libultraship/libultraship.h>
```
Eblo: You probably don't need this entire import.

mm/2s2h/DeveloperTools/DLViewer.cpp
```diff
@@ -0,0 +1,351 @@
...
+    }
+    UIWidgets::PopStyleInput();
+
+    if (doSearch) {
+        if (searchDebounceFrames == 0) {
+            doSearch = false;
+            PerformDisplayListSearch();
+        }
+
+        searchDebounceFrames--;
```
Eblo: What is a "frame" in this context? Is it based on refresh rate? For higher refresh rates, 30 frames may not be meaningfully long enough to debounce.

mm/2s2h/DeveloperTools/DLViewer.cpp
```diff
@@ -0,0 +1,351 @@
...
+                if (ImGui::Selectable("gsDPPipeSync")) {
+                    *gfx = gsDPPipeSync();
+                }
+                if (ImGui::Selectable("gsSPGrayscale")) {
+                    *gfx = gsSPGrayscale(true);
+                }
+                if (ImGui::Selectable("gsDPSetGrayscaleColor")) {
+                    *gfx = gsDPSetGrayscaleColor(0, 0, 0, 255);
+                }
+                ImGui::EndCombo();
```
Eblo: Could we throw in `gsSPNoOp`? I think effectively canceling out a command in place will be useful, going by previous experiences in patching graphical commands.

mm/2s2h/DeveloperTools/DLViewer.cpp
```diff
@@ -0,0 +1,351 @@
...
+                }
+                ImGui::PopItemWidth();
+            }
+            if (cmd == G_RDPPIPESYNC) {}
+            if (cmd == G_SETGRAYSCALE) {
+                bool* state = (bool*)&gfx->words.w1;
+                ImGui::SameLine();
+                UIWidgets::PushStyleCheckbox(THEME_COLOR);
+                if (ImGui::Checkbox(("state" + id).c_str(), state)) {
+                    //
```
Eblo: ?

mm/2s2h/DeveloperTools/DLViewer.cpp
```diff
@@ -0,0 +1,351 @@
...
+    { G_SETCOMBINE, "gsDPSetCombineLERP" },
+    { G_CULLDL, "gsSPCullDisplayList" },
+    { G_NOOP, "gsDPNoOp" },
+    { G_SPNOOP, "gsSPNoOp" },
+    { G_MARKER, "LUS Custom Marker" },
+};
+
+void PerformDisplayListSearch() {
+    auto result = Ship::Context::GetInstance()->GetResourceManager()->GetArchiveManager()->ListFiles(
+        "*" + std::string(searchString) + "*DL*");
```
Eblo: This search is case sensitive. Is that intended?

mm/2s2h/DeveloperTools/DLViewer.cpp
```diff
@@ -0,0 +1,351 @@
...
+            ImGui::PushItemWidth(25.0f);
+            ImGui::Text("%lu", i);
+            ImGui::PopItemWidth();
+            ImGui::SameLine();
+            ImGui::PushItemWidth(175.0f);
+
+            UIWidgets::PushStyleCombobox(THEME_COLOR);
+            if (ImGui::BeginCombo(("CMD" + id).c_str(), cmdLabel.c_str())) {
+                if (ImGui::Selectable("gsDPSetPrimColor") && cmd != G_SETPRIMCOLOR) {
+                    *gfx = gsDPSetPrimColor(0, 0, 0, 0, 0, 255);
```
Eblo: If I'm following this right, you're actually irreversibly modifying the DL commands in place, meaning they will remain modified for the remainder of runtime. I think a better approach would be to call something like this, where the last argument is what you normally do to `gfx` now:

```c++
ResourceMgr_PatchGfxByName(activeDisplayList.c_str(), ("CMD" + id + "Debug").c_str(), id, gsDPSetPrimColor(0, 0, 0, 0, 0, 255));
```

One benefit to doing this is that you could add a revert button that calls this and restores the default DL command:

```c++
ResourceMgr_UnpatchGfxByName("scenes/nonmq/Z2_00KEIKOKU/Z2_00KEIKOKU_room_00DL_00D490", ("CMD + id + "Debug").c_str());
```

Eblo (commented):
One thing that would be nice is the ability to limit the DL list to those active in the current scene. However, DLs remain persistent in memory once loaded, and I don't think there is currently tracking for what is in active use, so I would say that's out of scope for this PR.

mm/2s2h/DeveloperTools/DLViewer.cpp
mckinlee: This is a remnant from SoH. Gonna move to a time-based debounce instead. Similar to BenInputEditorWindow.

mm/2s2h/DeveloperTools/DLViewer.cpp
mckinlee: I'll remove it. You're guess is as good as mine.. this was directly copy and pasted from SoH lol

mm/2s2h/DeveloperTools/DLViewer.cpp
mckinlee: Another copy/paste from SoH.  It looks like they may have intended it to not be. Glob_match appears to be case sensitive though, so what I'll do is convert everything to lowercase before comparing.

mm/2s2h/BenGui/BenGui.cpp
```diff
@@ -133,7 +133,7 @@ void SetupGuiElements() {
     mEventLogWindow = std::make_shared<EventLogWindow>("gWindows.EventLog", "Event Log", ImVec2(520, 600));
     gui->AddGuiWindow(mEventLogWindow);
 
-    mDLViewerWindow = std::make_shared<DLViewerWindow>("gWindows.DLViewer", "Display List Viewer", ImVec2(520, 600));
+    mDLViewerWindow = std::make_shared<DLViewerWindow>("gWindows.DLViewer", "DL Viewer", ImVec2(520, 600));
```
Eblo, line 141: Any particular reason you renamed it to an abbreviation?

mm/2s2h/DeveloperTools/DLViewer.cpp
```diff
@@ -126,219 +216,563 @@ void DLViewerWindow::DrawElement() {
...
+                           ImGuiColorEditFlags_NoTooltip | ImGuiColorEditFlags_NoPicker,
+                           ImVec2(buttonSize, buttonSize));
+        ImGui::SetCursorPosY(ImGui::GetCursorPosY() - offsetY);
+    }
+}
+
+std::string GetCommandMetadata(Gfx* gfx, int cmd, size_t currentIndex, size_t totalInstructions) {
+    std::string metadata;
+
+    if (cmd == G_SETTILE) {
```
Eblo: This should probably be a switch block.

mm/2s2h/BenGui/BenGui.cpp
mckinlee: For aesthetic reasons so it fits better on the side menu.

mm/2s2h/DeveloperTools/DLViewer.cpp
mckinlee: Good call. I went ahead and refactored it. That change is now in.

Eblo (approved):

Issue #1302, opened 2025-10-22
[wishlist] Option to disable enemy proximity music


koratfood:
In most (if not all) areas, every time Link gets close to a regular enemy, the local area/dungeon music fades out, and the battle theme fades in. This is reversed once Link has defeated or moved away from all adjacent enemies.

It would be nice to have an option to disable this behaviour, instead allowing area/dungeon music to keep playing regardless of whether Link is close to an enemy.

See also HarbourMasters/Shipwright#2118 where this was previously implemented.

Eblo:
This is now in develop and will be in the next named release.

PR #1325, opened 2025-11-06
[Audio Editor] Port Audio Options from SoH Audio Editor

Glought:
I have port over Audio Options from SoH Audio Editor.

- Moved "Mute Low HP Alarm" to Audio Editor -> Audio Options
   Changed Cvar
   From: "gEnhancements.Sfx.LowHpAlarm"
   To: "gAudioEditor.LowHpAlarm"

- Moved "Mute Carpenter Sounds" to Audio Editor -> Audio Options
   Changed Cvar
   From: "gEnhancements.Sfx.MuteCarpenterSfx"
   To: "gAudioEditor.MuteCarpenterSfx"

- Moved "Mute Crying Goron Child" to Audio Editor -> Audio Options
  Changed Cvar
  From: "gEnhancements.Sfx.ChildGoronCry"
  To: "gAudioEditor.ChildGoronCry"

- Added "Disable Tatle Call Audio".
- Added "Disable Enemy Proximity Music"  -> #1302.
- Added "Randomize All Music and Sound Effects on New Scene".
- Added "Randomize All Music and Sound Effects on Randomizer Generation".
- Added "Display Sequence Name on Overlay" checkbox  and "Overlay Duration" Slider.
- Added "Enable Link's Voice Pitch Multiplier" checkbox and "Link's Voice Pitch Multiplier"  Slider.
- Added "AddSearchWidget" Functionally .

2ship audio editor - ported 2

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4608882245.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4608936959.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4609154782.zip)


mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -10,13 +10,33 @@
 #include "../../BenPort.h"
 #include <ship/utils/StringHelper.h>
 #include "../../BenGui/UIWidgets.hpp"
+#include "../../BenGui/BenMenu.h"
+#include "../../BenGui/BenGui.hpp"
```
Eblo:
```suggestion
#include "2s2h/BenGui/BenMenu.h"
#include "2s2h/BenGui/BenGui.hpp"
```

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -416,17 +436,25 @@ void DrawTypeChip(SeqType type) {
...
-void AudioEditorRegisterOnSceneInitHook() {
-    // BENTODO implement this
-    // GameInteractor::Instance->RegisterGameHook<GameInteractor::OnSceneInit>([](int16_t sceneNum) {
-    //    if (CVarGetInteger(CVAR_AUDIO("RandomizeAllOnNewScene"), 0)) {
-    //        AudioEditor_RandomizeAll();
-    //    }
-    //});
+void AudioEditorRegisterOnSceneChangeHook() {
+    GameInteractor::Instance->RegisterGameHook<GameInteractor::OnSceneChange>([](int16_t sceneNum) {
+        if (CVarGetInteger("gAudioEditor.RandomizeAllOnNewScene", 0)) {
```
Eblo: The pattern in this file is to use `CVAR_AUDIO(var)`, so for consistency's sake, let's change these CVar calls.


```suggestion
        if (CVarGetInteger(CVAR_AUDIO("RandomizeAllOnNewScene"), 0)) {
```

I realize this may be moot when it comes to addressing CVar prefixes in general, but within the scope of this PR, I'd prefer we not have "gAudioEditor.Something" 10 times in one file.

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -416,17 +436,25 @@ void DrawTypeChip(SeqType type) {
...
+            AudioEditor_RandomizeAll();
+        }
+    });
+}
+
+void AudioEditorRegisterOnGenerationCompletionHook() {
+    GameInteractor::Instance->RegisterGameHook<GameInteractor::OnGenerationCompletion>([]() {
+        if (CVarGetInteger("gAudioEditor.RandomizeAllOnRandoGen", 0)) {
```
Eblo:
```suggestion
        if (CVarGetInteger(CVAR_AUDIO("RandomizeAllOnRandoGen"), 0)) {
```

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -760,4 +746,104 @@ void AudioEditor_UnlockAll() {
...
     Ship::Context::GetInstance()->GetWindow()->GetGui()->SaveConsoleVariablesNextFrame();
-}
\ No newline at end of file
+}
+
+void RegisterAudioWidgets() {
+
+    lowHpAlarm = { .name = "Mute Low HP Alarm", .type = WidgetType::WIDGET_CVAR_CHECKBOX };
+
+    lowHpAlarm.CVar("gAudioEditor.LowHpAlarm")
```
Eblo:
```suggestion
    lowHpAlarm.CVar(CVAR_AUDIO("LowHpAlarm"))
```

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -760,4 +746,104 @@ void AudioEditor_UnlockAll() {
...
+        .Options(CheckboxOptions()
+                     .Color(THEME_COLOR)
+                     .Tooltip("Mutes the beeping alarm when you are critically low on health."));
+
+    BenGui::mBenMenu->AddSearchWidget({ lowHpAlarm, "Enhancements", "Audio Editor", "Audio Options" });
+
+    muteCarpenterSfx = { .name = "Mute Carpenter Sounds", .type = WidgetType::WIDGET_CVAR_CHECKBOX };
+
+    muteCarpenterSfx.CVar("gAudioEditor.MuteCarpenterSfx")
```
Eblo:
```suggestion
    muteCarpenterSfx.CVar(CVAR_AUDIO("MuteCarpenterSfx"))
```

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -760,4 +746,104 @@ void AudioEditor_UnlockAll() {
...
+        .Options(CheckboxOptions()
+                     .Color(THEME_COLOR)
+                     .Tooltip("Requires scene reload to take effect. Mutes the carpenter sounds coming "
+                              "from the tower in South Clock Town."));
+
+    BenGui::mBenMenu->AddSearchWidget({ muteCarpenterSfx, "Enhancements", "Audio Editor", "Audio Options" });
+
+    childGoronCry = { .name = "Mute Crying Goron Child", .type = WidgetType::WIDGET_CVAR_CHECKBOX };
+
+    childGoronCry.CVar("gAudioEditor.ChildGoronCry")
```
Eblo:
```suggestion
    childGoronCry.CVar(CVAR_AUDIO("ChildGoronCry"))
```

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -760,4 +746,104 @@ void AudioEditor_UnlockAll() {
...
+        .Options(CheckboxOptions().Color(THEME_COLOR).Tooltip("Mutes the crying Goron child inside Goron Shrine."));
+
+    BenGui::mBenMenu->AddSearchWidget({ childGoronCry, "Enhancements", "Audio Editor", "Audio Options" });
+
+    tatlCall = { .name = "Disable Tatl Call Audio", .type = WidgetType::WIDGET_CVAR_CHECKBOX };
+
+    tatlCall.CVar("gAudioEditor.DisableTatlCallAudio")
```
Eblo:
```suggestion
    tatlCall.CVar(CVAR_AUDIO("DisableTatlCallAudio"))
```

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -760,4 +746,104 @@ void AudioEditor_UnlockAll() {
...
+        .Options(CheckboxOptions().Color(THEME_COLOR).Tooltip("Disables the bell audio when Tatl calls you."));
+
+    BenGui::mBenMenu->AddSearchWidget({ tatlCall, "Enhancements", "Audio Editor", "Audio Options" });
+
+    enemyProx = { .name = "Disable Enemy Proximity Music", .type = WidgetType::WIDGET_CVAR_CHECKBOX };
+
+    enemyProx.CVar("gAudioEditor.EnemyBGMDisable")
```
Eblo:
```suggestion
    enemyProx.CVar(CVAR_AUDIO("EnemyBGMDisable"))
```

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -760,4 +746,104 @@ void AudioEditor_UnlockAll() {
...
+                     .Color(THEME_COLOR)
+                     .Tooltip("Disables the music change when getting close to enemies. Useful for hearing "
+                              "your custom music for each scene more often."));
+
+    BenGui::mBenMenu->AddSearchWidget({ enemyProx, "Enhancements", "Audio Editor", "Audio Options" });
+
+    randoMusicOnSceneChange = { .name = "Randomize All Music and Sound Effects on New Scene",
+                                .type = WidgetType::WIDGET_CVAR_CHECKBOX };
+
+    randoMusicOnSceneChange.CVar("gAudioEditor.RandomizeAllOnNewScene")
```
Eblo:
```suggestion
    randoMusicOnSceneChange.CVar(CVAR_AUDIO("RandomizeAllOnNewScene"))
```

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -760,4 +746,104 @@ void AudioEditor_UnlockAll() {
...
+        .Options(CheckboxOptions()
+                     .Color(THEME_COLOR)
+                     .Tooltip("Enables randomizing all unlocked music and sound effects when you enter a new scene."));
+
+    BenGui::mBenMenu->AddSearchWidget({ randoMusicOnSceneChange, "Enhancements", "Audio Editor", "Audio Options" });
+
+    randomAudioOnSeedGen = { .name = "Randomize All Music and Sound Effects on Randomizer Generation",
+                             .type = WidgetType::WIDGET_CVAR_CHECKBOX };
+
+    randomAudioOnSeedGen.CVar("gAudioEditor.RandomizeAllOnRandoGen")
```
Eblo:
```suggestion
    randomAudioOnSeedGen.CVar(CVAR_AUDIO("RandomizeAllOnRandoGen"))
```

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -760,4 +746,104 @@ void AudioEditor_UnlockAll() {
...
+        .Options(CheckboxOptions()
+                     .Color(THEME_COLOR)
+                     .Tooltip("Enables randomizing all unlocked music and sound effects when you generate a new "
+                              "randomizer. Respects locks already in place."));
+
+    BenGui::mBenMenu->AddSearchWidget({ randomAudioOnSeedGen, "Enhancements", "Audio Editor", "Audio Options" });
+
+    displaySeqName = { .name = "Display Sequence Name on Overlay", .type = WidgetType::WIDGET_CVAR_CHECKBOX };
+
+    displaySeqName.CVar("gAudioEditor.SeqNameNotification")
```
Eblo:
```suggestion
    displaySeqName.CVar(CVAR_AUDIO("SeqNameNotification"))
```

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -760,4 +746,104 @@ void AudioEditor_UnlockAll() {
...
+                     .Color(THEME_COLOR)
+                     .Tooltip("Displays the name of the current sequence in the corner of the screen whenever a new "
+                              "sequence "
+                              "is loaded to the main sequence player (does not apply to fanfares or enemy BGM)."));
+
+    BenGui::mBenMenu->AddSearchWidget({ displaySeqName, "Enhancements", "Audio Editor", "Audio Options" });
+
+    ovlDuration = { .name = "Overlay Duration: %d seconds", .type = WidgetType::WIDGET_CVAR_SLIDER_INT };
+
+    ovlDuration.CVar("gAudioEditor.SeqNameNotificationDuration")
```
Eblo:
```suggestion
    ovlDuration.CVar(CVAR_AUDIO("SeqNameNotificationDuration"))
```

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -760,4 +746,104 @@ void AudioEditor_UnlockAll() {
...
+        .Options(IntSliderOptions().Color(THEME_COLOR).Min(1).Max(10).DefaultValue(5).Size(ImVec2(300.0f, 0.0f)));
+
+    BenGui::mBenMenu->AddSearchWidget({ ovlDuration, "Enhancements", "Audio Editor", "Audio Options" });
+
+    voicePitch = { .name = "Link's Voice Pitch Multiplier", .type = WidgetType::WIDGET_CVAR_SLIDER_FLOAT };
+
+    voicePitch.CVar("gAudioEditor.LinkVoiceFreqMultiplier")
```
Eblo:
```suggestion
    voicePitch.CVar(CVAR_AUDIO("LinkVoiceFreqMultiplier"))
```

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -760,4 +746,104 @@ void AudioEditor_UnlockAll() {
...
+void RegisterAudioWidgets() {
+
+    lowHpAlarm = { .name = "Mute Low HP Alarm", .type = WidgetType::WIDGET_CVAR_CHECKBOX };
+
+    lowHpAlarm.CVar("gAudioEditor.LowHpAlarm")
+        .Options(CheckboxOptions()
+                     .Color(THEME_COLOR)
+                     .Tooltip("Mutes the beeping alarm when you are critically low on health."));
+
+    BenGui::mBenMenu->AddSearchWidget({ lowHpAlarm, "Enhancements", "Audio Editor", "Audio Options" });
```
Eblo: With this being identical for the 10 different calls aside from the widget argument itself maybe we could make this a function call? Something like:

```c++
void AddAudioSearchWidget(WidgetInfo& widgetInfo) {
    BenGui::mBenMenu->AddSearchWidget({ widgetInfo, "Enhancements", "Audio Editor", "Audio Options" });
}
```

And then replace each use you have here with:

```c++
AddAudioSearchWidget(lowHpAlarm);
```

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -760,4 +746,104 @@ void AudioEditor_UnlockAll() {
     }
 
     Ship::Context::GetInstance()->GetWindow()->GetGui()->SaveConsoleVariablesNextFrame();
-}
\ No newline at end of file
+}
+
+void RegisterAudioWidgets() {
```
Eblo, line 749: Can you remove the extra line breaks within each section? For example, the `lowHpAlarm` section should have no blank lines within it, but it should have one blank line after to separate it from the next `muteCarpenterSfx` section. I think it would be easier to tell at a glance how everything is grouped.

mm/2s2h/GameInteractor/GameInteractor_HookTable.h
```diff
@@ -52,3 +53,9 @@ DEFINE_HOOK(OnItemGive, (u8 item))
 DEFINE_HOOK(OnBottleContentsUpdate, (u8 item))
 
 DEFINE_HOOK(ShouldVanillaBehavior, (GIVanillaBehavior flag, bool* should, va_list originalArgs))
+
+// Audio
+DEFINE_HOOK(OnSeqPlayerInit, (int32_t playerIdx, int32_t seqId));
```
Eblo:
```suggestion
DEFINE_HOOK(OnSeqPlayerInit, (s32 playerIdx, s32 seqId));
```

mm/2s2h/GameInteractor/GameInteractor_HookTable.h
```diff
@@ -52,3 +53,9 @@ DEFINE_HOOK(OnItemGive, (u8 item))
...
+
+// Rando
+DEFINE_HOOK(OnGenerationCompletion, ());
```
Eblo: While the comment is there, I think the hook name should also clarify that this is rando-specific. Maybe `OnRandoSeedGeneration`? Generally our "On" hooks happen after something, so "completion" is not needed in the name.

mm/2s2h/GameInteractor/GameInteractor.h
```diff
@@ -726,6 +726,7 @@ void GameInteractor_ExecuteAfterInterfaceClockDraw();
 void GameInteractor_ExecuteBeforeInterfaceClockDraw();
 
 void GameInteractor_ExecuteOnSceneInit(s16 sceneId, s8 spawnNum);
+void GameInteractor_ExecuteOnSceneChange(s16 sceneId);
```
Eblo: What about `OnSceneInit` was insufficient to warrant this new hook?

mm/src/code/z_actor.c
```diff
@@ -2485,7 +2488,17 @@ void func_800B8E1C(PlayState* play, Actor* actor, f32 arg2, s16 arg3, f32 arg4)
...
         Audio_PlaySfx_AtPosWithPresetLowFreqAndHighReverb(&player->actor.projectedPos, sfxId);
+    } else if (sfxId >= NA_SE_VO_LI_SWORD_N && sfxId <= NA_SE_VO_DEMO_394 || sfxId == NA_SE_PL_TRANSFORM_VOICE) {
+
+        freqMultiplier = CVarGetFloat("gAudioEditor.LinkVoiceFreqMultiplier", 1.0);
+        if (freqMultiplier <= 0) {
+            freqMultiplier = 1;
+        }
+
+        AudioSfx_PlaySfx(sfxId, &player->actor.projectedPos, 4, &freqMultiplier, &gSfxDefaultFreqAndVolScale,
+                         &gSfxDefaultReverb);
```
Eblo: I'm not sure I like introducing a new static variable, condition flow, and inline CVar into source code for this one. I think the better route would be to wrap the normal `AudioSfx_PlaySfx` call below in a GameInteractor_Should call with a new VB, then put the custom sfx stuff in an enhancement file.

mm/src/code/z_parameter.c
```diff
@@ -5091,12 +5091,15 @@ void Interface_SetTatlCall(PlayState* play, u16 tatlCallState) {
 
     if (((tatlCallState == TATL_STATE_2A) || (tatlCallState == TATL_STATE_2B)) && !interfaceCtx->tatlCalling &&
         (play->csCtx.state == CS_STATE_IDLE)) {
-        if (tatlCallState == TATL_STATE_2B) {
-            Audio_PlaySfx(NA_SE_VO_NAVY_CALL);
-        }
-        if (tatlCallState == TATL_STATE_2A) {
-            Audio_PlaySfx_AtPosWithReverb(&gSfxDefaultPos, NA_SE_VO_NA_HELLO_2, 0x20);
+        if (!CVarGetInteger("gAudioEditor.DisableTatlCallAudio", 0)) {
```
Eblo: Ditto, I'd rather this be a VB instead.

mm/src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -12207,7 +12207,7 @@ void Player_UpdateCamAndSeqModes(PlayState* play, Player* this) {
             Camera_ChangeMode(camera, camMode);
         }
 
-        if (play->actorCtx.attention.bgmEnemy != NULL) {
+        if (play->actorCtx.attention.bgmEnemy != NULL && !CVarGetInteger("gAudioEditor.EnemyBGMDisable", 0)) {
```
Eblo: Ditto again, VB.

Eblo (changes requested):

mm/2s2h/GameInteractor/GameInteractor.h
Glought: When I tried using `OnSceneInit` the music stopped playing. From my testing the music does get randomized but 
in the file `"mm/src/code/z_play.c" line 2497:  "GameInteractor_ExecuteOnSceneInit(sceneIdAbsolute, spawnNum);"` causes the music to stop playing. When I commented it out music will play. I don't know the exact purpose of that line I didn't want to risk breaking something by changing or removing it so I made `OnSceneChange` if you have any suggestions please let me know.

Caladius:
Just wanted to say thank you for taking the time and effort to port some of these things over to 2Ship.

mm/2s2h/GameInteractor/GameInteractor.h
garrettjoecox: Could you also try `AfterRoomSceneCommands` and `OnRoomInit`? If we do need to go with this place, I'll do some digging and figure out what the right name is, I don't think SceneChange is it

mm/2s2h/BenGui/MenuTypes.h
```diff
@@ -253,6 +260,15 @@ struct SidebarEntry {
     std::vector<std::vector<WidgetInfo>> columnWidgets;
 };
 
+struct SearchWidget {
```
garrettjoecox, line 263: Is this also something ported from SoH? Is the intent that we can have searchable widgets that may not be apart of a more formal "modern menu" setup?

Glought: Yes its ported from SoH it's used by `void AddSearchWidget(SearchWidget widget);` in Menu.h. Without it the options in AudioEditor are unsearchable.

The PR on Shipwright is ` Search Exclusion and Search Widgets #5656 `  I only ported what was needed  for `AddSearchWidget` to work.

mm/2s2h/GameInteractor/GameInteractor.h
Eblo: > causes the music to stop playing. When I commented it out music will play.

It may or may not be relevant, but there is a BGM replay patch that kills the main sequence player on transitions if certain settings are active. See `VB_PLAY_SCENE_SEQUENCE`. Try using `OnSceneInit` with that setting disabled if you haven't already.

Glought: I never had that setting enabled when I tried using `OnSceneInit.`  
I just tested using `OnRoomInit` it does work and I don't need to add `GameInteractor_ExecuteOnRoomInit` to `z_play2SH.cpp: OTRPlay_SpawnScene` function.

Glought: I pushed a commit removing `OnSceneChange` and to use 'OnRoomInit' instead.

Glought:
I have updated the PR message to reflect the changes I made.

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -416,17 +437,25 @@ void DrawTypeChip(SeqType type) {
...
 
-void AudioEditorRegisterOnSceneInitHook() {
-    // BENTODO implement this
-    // GameInteractor::Instance->RegisterGameHook<GameInteractor::OnSceneInit>([](int16_t sceneNum) {
-    //    if (CVarGetInteger(CVAR_AUDIO("RandomizeAllOnNewScene"), 0)) {
-    //        AudioEditor_RandomizeAll();
-    //    }
-    //});
+void AudioEditorRegisterOnRoomInitHook() {
+    GameInteractor::Instance->RegisterGameHook<GameInteractor::OnRoomInit>([](s8 spawnNum, s8 roomNum) {
```
Eblo: With `OnRoomInit`, the audio gets shuffled every time the player enters a new *room* rather than a scene (think anytime you use a door, or traverse the Woods of Mystery). I'm not sure if that was your intent, but it feels rather overwhelming to me.

mm/src/code/z_actor.c
```diff
@@ -2486,8 +2486,12 @@ void Player_PlaySfx(Player* player, u16 sfxId) {
...
         Audio_PlaySfx_AtPosWithPresetLowFreqAndHighReverb(&player->actor.projectedPos, sfxId);
     } else {
-        AudioSfx_PlaySfx(sfxId, &player->actor.projectedPos, 4, &gSfxDefaultFreqAndVolScale,
-                         &gSfxDefaultFreqAndVolScale, &gSfxDefaultReverb);
+        if (GameInteractor_Should(VB_LINK_VOICE_PITCH_MULTIPLIER, false, &sfxId)) {
+            return;
+        } else {
+            AudioSfx_PlaySfx(sfxId, &player->actor.projectedPos, 4, &gSfxDefaultFreqAndVolScale,
+                             &gSfxDefaultFreqAndVolScale, &gSfxDefaultReverb);
+        }
```
Eblo: Let's try not to change flow if we can help it. I know there are areas that do this, but let's try and keep that to a minimum, only doing so if necessary.

```suggestion
        if (GameInteractor_Should(VB_LINK_VOICE_PITCH_MULTIPLIER, true, &sfxId)) {
            AudioSfx_PlaySfx(sfxId, &player->actor.projectedPos, 4, &gSfxDefaultFreqAndVolScale,
                             &gSfxDefaultFreqAndVolScale, &gSfxDefaultReverb);
        }
```

mm/2s2h/Enhancements/Audio/Sfx/LinksVoicePitchMultiplier.cpp
```diff
@@ -0,0 +1,37 @@
...
+
+// For Link's voice pitch SFX modifier
+static f32 freqMultiplier = 1;
+
+#define CVAR_NAME "gAudioEditor.LinkVoiceFreqMultiplier.Enable"
+#define CVAR CVarGetInteger(CVAR_NAME, 0)
+
+void RegisterLinksVoicePitchMultiplier() {
+    COND_VB_SHOULD(VB_LINK_VOICE_PITCH_MULTIPLIER, CVAR, {
+        *should = true;
```
Eblo: Paired with my suggested flow change, this would need to be set to false later, after checking the condition.

```suggestion
```

mm/2s2h/Enhancements/Audio/Sfx/LinksVoicePitchMultiplier.cpp
```diff
@@ -0,0 +1,37 @@
...
+            if (freqMultiplier <= 0) {
+                freqMultiplier = 1;
+            }
+
+            AudioSfx_PlaySfx(sfxId, &player->actor.projectedPos, 4, &freqMultiplier, &gSfxDefaultFreqAndVolScale,
+                             &gSfxDefaultReverb);
+        } else {
+            AudioSfx_PlaySfx(sfxId, &player->actor.projectedPos, 4, &gSfxDefaultFreqAndVolScale,
+                             &gSfxDefaultFreqAndVolScale, &gSfxDefaultReverb);
+        }
```
Eblo: With how `should` works, you won't have to re-implement default behavior. It will just happen by default if you don't set `should` to false.

```suggestion
        }
```

mm/2s2h/Enhancements/Audio/Sfx/LinksVoicePitchMultiplier.cpp
```diff
@@ -0,0 +1,37 @@
...
+#define CVAR CVarGetInteger(CVAR_NAME, 0)
+
+void RegisterLinksVoicePitchMultiplier() {
+    COND_VB_SHOULD(VB_LINK_VOICE_PITCH_MULTIPLIER, CVAR, {
+        *should = true;
+        Player* player = GET_PLAYER(gPlayState);
+        u16 sfxId = *va_arg(args, u16*);
+
+        if (sfxId >= NA_SE_VO_LI_SWORD_N && sfxId <= NA_SE_VO_DEMO_394 || sfxId == NA_SE_PL_TRANSFORM_VOICE) {
+
```
Eblo, line 21: At this point in your code within the conditional, we know we're processing a Link voice sfx, so we know not to do the vanilla behavior, thus we can set this here.
```suggestion
            *should = false;
```

Eblo (changes requested):
Appreciate the cleanup and organizational changes! I have some tidy up suggestions. Randomizing audio with `OnRoomInit` is an issue, as I don't think it has the intended effect you were going for. It might seem like the answer in most scenes, but scenes with multiple rooms, like dungeons, will result in somewhat frequent audio shuffling.

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
Glought: On further testing with `OnSceneInit` it semi works it doesn't play music in Clock Tower, Clock Town, Stock Pot Inn, Termina Field, Mayors Office, Great Bay, but will play music in the Milk bar, Honey & Darling's shop, Shops.

Eblo: Just to be sure, have you tried it with all of these settings off?

- Faster scene transitions
- Debug mode
- Pause save

The BGM replay thing applies if any one of those is enabled.

Glought: still doesn't work with those settings off.

Eblo: Just for the heck of it, since the VB already exists, why not try your `AudioEditor_RandomizeAll` call within a     `COND_VB_SHOULD(VB_PLAY_SCENE_SEQUENCE...`? That runs just before a scene decides to play its music. I'm not 100% sure it will work, but might as well try.

Glought: I even tested by disabling BGM Replay in code it still doesn't work.

Glought: Calling `AudioEditor_RandomizeAll ` within the `COND_VB_SHOULD` worked.

Glought: Some things I notice it doesn't get triggered when moving to other parts of clock town, termina Field. It triggers when entering shops,inn,etc,  or when leaving clock town  `clock town -> termina field`  or 
`Termina field -> Woodfall`. When calling `AudioEditor_RandomizeAll ` outside the `COND_VB_SHOULD ` it would randomize every time the scene changed.

Eblo: That is curious. What exactly does your condition look like, and what do you mean by calling the randomization outside? I know sometimes music persists between scenes like in Clock Town, but that condition is evaluated after the placement of `VB_PLAY_SCENE_SEQUENCE`.

Glought:
```c++
void AudioEditorRegisterOnSceneInitHook() {
    GameInteractor::Instance->RegisterGameHook<GameInteractor::OnSceneInit>([](s8 sceneId, s8 spawnNum) {
        COND_VB_SHOULD(VB_PLAY_SCENE_SEQUENCE, CVAR_AUDIO("RandomizeAllOnNewScene"), {
            AudioEditor_RandomizeAll();
        });
    });
}
```

Eblo: A `COND_VB_SHOULD`, as well as any other hook macro, should be called at the top level of a register func, not within another hook. So when you're calling it "outside" in this case, that's effectively no different than just directly doing it in the OnSceneInit hook. Try this, remembering to call the register func itself in the overall InitElement:

```c++
void AudioEditorRegisterPlaySceneSequenceHook() {
    COND_VB_SHOULD(VB_PLAY_SCENE_SEQUENCE, CVAR_AUDIO("RandomizeAllOnNewScene"), {
        AudioEditor_RandomizeAll();
    });
}
```

Glought: What about this
```c++
void AudioEditorRegisterRandomizeAllOnNewSceneHook() {
    COND_VB_SHOULD(VB_RANDOMIZE_ALL_ON_NEW_SCENE, CVarGetInteger(CVAR_AUDIO("RandomizeAllOnNewScene"), 0), {
        if (CVarGetInteger(CVAR_AUDIO("RandomizeAllOnNewScene"), 0)) {
            AudioEditor_RandomizeAll();
        }
    });
}
``` 
and calling it at the end of  `z_play_2SH.cpp`  in the `OTRPlay_SpawnScene` function.

```c++
extern "C" void OTRPlay_SpawnScene(PlayState* play, s32 sceneId, s32 spawn) {
    s32 pad;
    SceneTableEntry* scene = &gSceneTable[sceneId];

    scene->unk_D = 0;
    play->loadedScene = scene;
    play->sceneId = sceneId;
    play->sceneConfig = scene->drawConfig;
    std::string scenePath =
        StringHelper::Sprintf("scenes/nonmq/%s/%s", scene->segment.fileName, scene->segment.fileName);
    play->sceneSegment = OTRPlay_LoadFile(play, scenePath.c_str());
    scene->unk_D = 0;
    gSegments[2] = (uintptr_t)play->sceneSegment;
    OTRPlay_InitScene(play, spawn);
    Room_SetupFirstRoom(play, &play->roomCtx);
    
    GameInteractor_Should(VB_RANDOMIZE_ALL_ON_NEW_SCENE, false);  
}
``` 

It does work and the audio is randomized when moving to different areas in clock town  
`South Clock Town -> North Clock Town`

Eblo: When you call `COND_VB_SHOULD(VB_SOMETHING, someCondition)`, the code within will not run unless `someCondition` is true, so it's redundant to check `someCondition` inside the macro.

I checked out your changes locally to do some digging and found that my suggestion still won't work in areas like Clock Town anyway, like you've said, and the reason is that transitions within Clock Town do not reach that hook. When I try OnSceneInit, I believe it occurs too late, as I get no BGM with a display that shows the vanilla BGM name.

Then I tried `VB_PLAY_TRANSITION_CS`, which occurs on scene transitions before the actual scene load and seems to work as you intended. If no one sees any other issues with this approach, I think you can do this:

```c++
void AudioEditor::InitElement() {
    COND_VB_SHOULD(VB_PLAY_TRANSITION_CS, CVAR_AUDIO("RandomizeAllOnNewScene"), {
        AudioEditor_RandomizeAll();
    });

    GameInteractor::Instance->RegisterGameHook<GameInteractor::OnRandoSeedGeneration>([]() {
        if (CVarGetInteger(CVAR_AUDIO("RandomizeAllOnRandoGen"), 0)) {
            AudioEditor_RandomizeAll();
        }
    });
}
```

Note that I condensed the various functions into one since there didn't seem to be any need to have them outside of `InitElement`.

Glought: The reason I had a if statement inside `COND_VB_SHOULD` is because it ignored `CVAR_AUDIO("RandomizeAllOnRandoGen")` and it would trigger the code inside the macro no matter if `"Randomize All Music and Sound Effects on New Scene"` was checked or unchecked it all ways triggered. 

It didn't matter if the ` GameInteractor_Should(VB_RANDOMIZE_ALL_ON_NEW_SCENE, false);  ` was set to true or false.
Maybe being triggered from  `z_play_2SH.cpp` in the `OTRPlay_SpawnScene` function has something to do about it ignoring the condition.

Glought: I just tested `VB_PLAY_TRANSITION_CS` and it still ignores the condition and triggers the code inside the macro.
Also the music cuts out when transitioning to the different areas of clock town after loading a game save but then works fine after entering and leaving a shop or clock town then the music works like it should.

Glought: I forgot about `static RegisterShipInitFunc initFunc` so i placed 

```c++
static RegisterShipInitFunc initFuncRandomizeAllOnNewScene(AudioEditorRegisterRandomizeAllOnNewScene,
                                     { CVAR_AUDIO("RandomizeAllOnNewScene") });
``` 
at the end of AudioEditor and now it follows the condition.

Eblo: Yep, that would explain it. Without using ShipInit before, it would only evaluate the condition once and not refresh on changes.

Glought: Any suggestions on the Music stops playing on first transition after game load issue?.

Eblo: I think you should be using `VB_PLAY_TRANSITION_CS`:

```c++
void AudioEditorRegisterRandomizeAllOnNewScene() {
    COND_VB_SHOULD(VB_PLAY_TRANSITION_CS, CVAR_AUDIO("RandomizeAllOnNewScene"), { AudioEditor_RandomizeAll(); });
}

static RegisterShipInitFunc initFuncRandomizeAllOnNewScene(AudioEditorRegisterRandomizeAllOnNewScene,
                                                           { CVAR_AUDIO("RandomizeAllOnNewScene") });
```

`VB_PLAY_TRANSITION_CS` occurs just before the game loads a new scene. I think your `VB_RANDOMIZE_ALL_ON_NEW_SCENE` occurs too late, `OnSceneInit` occurs too late, and `VB_PLAY_SCENE_SEQUENCE` does not run between scenes with shared music like Clock Town.

Glought: I am using `VB_PLAY_TRANSITION_CS` during my tests today.  
Here I recorded the issue




Music stops on first transition after game load.
https://github.com/user-attachments/assets/5e6b8c97-14b7-442a-9ba5-6fc78ce139bc

Music works fine after entering and leaving a shop.
https://github.com/user-attachments/assets/1cc52b17-25fd-4cb0-9720-62aca1748235

Eblo: Go ahead and push what you have, even if it's not fully functional. Locally, I have your PR as-is but with the previous code snippet I shared, and I'm not getting that behavior when loading in South Clock Town and going west.

Glought: ok i pushed what i got

Eblo: Would you mind sanity checking the PR build? It works as intended on my end where I boot up, load into a game, appear in South Clock Town, and go to West Clock Town. With how this only rebuilds in pieces, it's possible something went stale and caused the issue for you locally. If you can confirm the PR build is working, then I'd try deleting your build folder and rebuilding from scratch to try again.

Glought: Deleting the build folder and rebuilding fixed the issue. I thought about doing that yesterday but didn't have the time till now.

Eblo (approved):

PR #1326, opened 2025-11-06
[Rando] Add Junk Customization Options

Caladius:
This adds 3 modes to RI_JUNK items:

- Default: The default cyclical behavior except you can now select which items appear in the cycle.
- Weighted: Items appear by chance, chance is set by the Item Weights that can be customized per item.
- Supply: This uses customizable thresholds to determine which item is the current Junk item.

TODO:
- [ ] Testing
- [ ] Weighted needs some review and maybe some improvements, I'm not happy with it but its functional.
- [ ] Remove Rupees from Threshold and Weighted as they are the fallback.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4591560209.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4591663918.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4592209367.zip)


mm/2s2h/Rando/Rando.h
```diff
@@ -14,13 +14,21 @@
...
 namespace Rando {
 
 void Init();
 void DrawItem(RandoItemId randoItemId, Actor* actor = nullptr);
 void GiveItem(RandoItemId randoItemId);
 void RemoveItem(RandoItemId randoItemId);
 RandoItemId CurrentJunkItem();
+void UpdateJunkOptions();
+void UpdateJunkWeights();
+void UpdateJunkThresholds();
```
mckinlee: `UpdateJunkThresholds` appears to be unused.

mm/2s2h/Rando/ConvertItem.cpp
```diff
@@ -53,46 +54,159 @@ extern GetItemEntry sGetItemTable[GI_MAX - 1];
...
+        junkSelectionList.push_back(RI_NONE);
+    }
+}
+
+void Rando::UpdateJunkWeights() {
+    junkItemWeights.clear();
+
+    for (const auto& [itemId, data] : junkCvarMap) {
+        if (CVarGetInteger(JUNK_CVAR(itemId, "Weight"), 10) > 0) {
+            junkItemWeights.emplace_back(itemId, CVarGetInteger(JUNK_CVAR(itemId, "Enabled"), 10));
```
mckinlee: Why are you grabbing `Enabled` instead of `Weight` here? Wouldn't this make every junk item wind up with weight 0, 1, or 10 rather than actually respecting the slider value?

``` suggestion
        if (CVarGetInteger(JUNK_CVAR(itemId, "Enabled"), 1) == 0) {
            continue;
        }

        uint16_t weight = CVarGetInteger(JUNK_CVAR(itemId, "Weight"), 10);
        if (weight > 0) {
            junkItemWeights.emplace_back(itemId, weight);
```

Caladius: These both needed to be "Weight". The first is to check if the Weight is greater than 0 (exclude the item if 0) and the second is adding whatever the weight is set at to the list.

The way CVar look up works is the "10" is a default value, in the event the CVar does not exist in the players json due to never interacting with the option, it will default to 10 weight.

mckinlee: Ah, that makes more sense.

mm/2s2h/Rando/Menu.cpp
```diff
@@ -301,23 +324,53 @@ static void DrawItemsTab() {
...
+                                  .Color(UIWidgets::Colors(CVarGetInteger("gSettings.Menu.Theme", 5)))
+                                  .Format("%i")
+                                  .Min(0)
+                                  .Max(maxThreshold)
+                                  .DefaultValue(10));
+            }
+            break;
+        default:
+            break;
+    }
```
Eblo: After mulling this over, I think junk options might be better placed in their own tab. It's already bigger than the hint options which have their own tab, plus these junk options would be the odd one out on this tab since they don't impact seed generation like the others.

mm/2s2h/Rando/ConvertItem.cpp
```diff
@@ -53,46 +54,159 @@ extern GetItemEntry sGetItemTable[GI_MAX - 1];
...
+        return 96;
+    } else if (randoItemId == RI_DEKU_STICKS_5) {
+        return 10;
+    } else if (randoItemId == RI_DEKU_NUTS_5) {
+        return 20;
+    } else if (randoItemId == RI_BOMBS_5 || randoItemId == RI_BOMBCHU_5) {
+        return 40;
+    } else if (randoItemId == RI_ARROWS_10) {
+        return 50;
+    }
```
Eblo: This could be a map where 500 is the fallback value if no matching RI key is found. Would be better in the long run should we ever decide to toss in more items like refills.

mm/2s2h/Rando/ConvertItem.cpp
```diff
@@ -53,46 +54,159 @@ extern GetItemEntry sGetItemTable[GI_MAX - 1];
...
+                currentWeight += weight;
+                if (weightedValue < currentWeight) {
+                    lastJunkItem = itemId;
+                    break;
+                }
+            }
+            break;
+        case RO_JUNK_TYPE_SUPPLY:
+            if (gSaveContext.save.saveInfo.playerData.health <=
+                CVarGetInteger(JUNK_CVAR(RI_RECOVERY_HEART, "Threshold"), 10)) {
```
Eblo: I don't know if this will be intuitive to the player, as the health measurement you see in-game does not match the numerical value. I don't know what code handles the conversion, but I think it would make more sense if this threshold value aligned with quarter hearts.

mm/2s2h/Rando/ConvertItem.cpp
```diff
@@ -53,46 +54,159 @@ extern GetItemEntry sGetItemTable[GI_MAX - 1];
...
+                       AMMO(ITEM_BOMBCHU) <= CVarGetInteger(JUNK_CVAR(RI_BOMBCHU_5, "Threshold"), 10)) {
+                lastJunkItem = RI_BOMBCHU_5;
+            } else {
+                if (lastRupee != gSaveContext.save.saveInfo.playerData.rupees) {
+                    std::vector<RandoItemId> rupeeList = {
+                        RI_RUPEE_GREEN,
+                        RI_RUPEE_BLUE,
+                        RI_RUPEE_RED,
+                        RI_RUPEE_PURPLE,
+                    };
```
Eblo: Could make this static since it only needs initialized once.

mm/2s2h/Rando/Menu.cpp
```diff
@@ -301,23 +324,53 @@ static void DrawItemsTab() {
...
-                            .disabled = IncompatibleWithLogicSetting(RO_SHUFFLE_BOSS_SOULS),
-                            .disabledTooltip = "Incompatible with current Logic Setting" } }));
-    CVarCheckbox("Enemy Drops", Rando::StaticData::Options[RO_SHUFFLE_ENEMY_DROPS].cvar,
-                 CheckboxOptions({ { .tooltip = "Shuffles the first drop from a non Boss Enemy." } }));
-    CVarCheckbox("Enemy Souls", "gPlaceholderBool",
-                 CheckboxOptions({ { .disabled = true, .disabledTooltip = "Coming Soon" } }));
+    UIWidgets::PushStyleCombobox();
+    UIWidgets::CVarCombobox("Junk Item Options", "gRando.Junk.ItemType", &junkTypeOptions);
+    UIWidgets::Tooltip("Default - Junk Item will cycle through all junk options.\n"
+                       "");
```
Eblo: The tooltip should explain the other options.

mm/2s2h/Rando/ConvertItem.cpp
```diff
@@ -53,46 +54,159 @@ extern GetItemEntry sGetItemTable[GI_MAX - 1];
...
+            if (junkSelectionList.size() == 0) {
+                Rando::UpdateJunkOptions();
+            }
+
+            if (gPlayState != NULL && ABS(gPlayState->gameplayFrames - lastChosenAt) > 20) {
+                lastChosenAt = gPlayState->gameplayFrames;
+                lastJunkItem = RI_UNKNOWN;
+            }
+
+            while (lastJunkItem == RI_UNKNOWN) {
```
Eblo: If I toggle various default junk options while in-game, I sometimes cause the application to freeze. It happened once when I deselected everything and then a couple of times when simply toggling some options. I think something fell into an infinite loop.

Looking at this, it is possible for `lastJunkItem` to be `RI_UNKNOWN` before hitting the loop. Depending on what happens if `junkSelectionList` is empty, it might just remain that value, thus remaining trapped in this loop. The while loop probably isn't a good idea here since the `rand()` could theoretically constantly trip something that isn't obtainable.

mm/2s2h/Rando/Menu.cpp
Caladius: I took it one step further, the code that handles Junk is now in its own JunkItem.cpp file and the Menu related UI is in its own Junk Options sidebar.

mm/2s2h/Rando/ConvertItem.cpp
Caladius: I reworked this system to be a little less like a bitch. It has a failover counter now to prevent any lockups but also shouldnt really lock up anymore with the new selection process.

mm/2s2h/Rando/ConvertItem.cpp
Caladius: I made a conversion to the slider displaying as whole hearts, so 20 max for the slider.

mm/2s2h/Rando/JunkItem.cpp
```diff
@@ -0,0 +1,241 @@
...
+                lastChosenAt = gPlayState->gameplayFrames;
+                lastJunkItem = RI_UNKNOWN;
+            }
+
+            while (lastJunkItem == RI_UNKNOWN) {
+                RandoItemId randJunkItem = std::get<0>(junkSelectionList[rand() % junkSelectionList.size()]);
+                if (Rando::IsItemObtainable(randJunkItem)) {
+                    lastJunkItem = randJunkItem;
+                }
+                if (failOver >= FAIL_OVER_MAX) {
```
Eblo, line 86:
```suggestion
                // Reached max number of attempts, default to nothing
                if (failOver >= FAIL_OVER_MAX) {
```

mm/2s2h/Rando/JunkItem.cpp
```diff
@@ -0,0 +1,241 @@
...
+    uint32_t currentWeight = 0;
+    uint16_t failOver = 0;
+
+    if (junkSelectionList.size() == 0) {
+        Rando::UpdateJunkOptions();
+    }
+
+    switch (CVarGetInteger("gRando.Junk.ItemType", (uint32_t)RO_JUNK_TYPE_DEFAULT)) {
+        case RO_JUNK_TYPE_DEFAULT:
+            if (gPlayState != NULL && ABS(gPlayState->gameplayFrames - lastChosenAt) > 20) {
```
Eblo, line 76: Is it even possible for `gPlayState` to be null in a function that relies on an Actor* argument?

mm/2s2h/Rando/JunkItem.cpp
```diff
@@ -0,0 +1,241 @@
...
+                if (weight == 0) {
+                    continue;
+                }
+                if (!Rando::IsItemObtainable(itemId)) {
+                    continue;
+                }
+
+                totalWeight += weight;
+            }
+            if (totalWeight == 0) {
```
Eblo, line 109:
```suggestion
            // No obtainable items have weight, default to nothing
            if (totalWeight == 0) {
```

mm/2s2h/Rando/JunkItem.cpp
```diff
@@ -0,0 +1,241 @@
...
+                switch (itemId) {
+                    case RI_RECOVERY_HEART:
+                        if (gSaveContext.save.saveInfo.playerData.health <= threshold * 16) {
+                            lastJunkItem = RI_RECOVERY_HEART;
+                            break;
+                        }
+                        break;
+                    case RI_MAGIC_JAR_SMALL:
+                        if (Rando::IsItemObtainable(itemId)) {
+                            if (gSaveContext.save.saveInfo.playerData.magic <= threshold) {
```
Eblo, line 135: Magic never seems to drop even if I have an empty magic bar and nothing else has reached its threshold. Looking through this code, I'm not certain why that is.

mm/2s2h/Rando/JunkItem.cpp
```diff
@@ -0,0 +1,241 @@
...
+            if (currentActor == actor) {
+                break;
+            }
+
+            currentActor = actor;
+            for (auto& [itemId, weight, threshold] : junkSelectionList) {
+                if (weight == 0) {
+                    continue;
+                }
+                if (!Rando::IsItemObtainable(itemId)) {
```
Eblo, line 103: While this does exclude the item from the weight calculation, it is still possible to get items that should not be obtainable. For example, I can still get arrows from the weighted option even if I don't have a bow.

mm/2s2h/Rando/JunkItem.cpp
```diff
@@ -0,0 +1,241 @@
...
+                }
+
+                totalWeight += weight;
+            }
+            if (totalWeight == 0) {
+                lastJunkItem = RI_NONE;
+                break;
+            }
+
+            weightedValue = rand() % totalWeight;
```
Eblo, line 114: Can you implement `lastChosenAt` for weighted as well? Depending on what configuration you have, the resulting item can rapidly change every frame.

Threshold is more or less deterministic and does not share this problem.

mm/2s2h/Rando/JunkItem.cpp
```diff
@@ -0,0 +1,241 @@
...
+                lastJunkItem = RI_UNKNOWN;
+                switch (itemId) {
+                    case RI_RECOVERY_HEART:
+                        if (gSaveContext.save.saveInfo.playerData.health <= threshold * 16) {
+                            lastJunkItem = RI_RECOVERY_HEART;
+                            break;
+                        }
+                        break;
+                    case RI_MAGIC_JAR_SMALL:
+                        if (Rando::IsItemObtainable(itemId)) {
```
Eblo, line 134: You may be able to just wrap the entire switch block with `Rando::IsItemObtainable(itemId)`. Hearts, nuts, and sticks should already just return true.

mm/2s2h/Rando/JunkItem.cpp
```diff
@@ -0,0 +1,241 @@
...
+                                break;
+                            }
+                        }
+                        break;
+                    default:
+                        break;
+                }
+                if (lastJunkItem == RI_UNKNOWN) {
+                    if (lastRupee != gSaveContext.save.saveInfo.playerData.rupees) {
+                        selectedRupee = rupeeList[rand() % rupeeList.size()];
```
Eblo, line 182: Maybe `lastChosenAt` should be implemented for threshold after all, at least for this particular case. If you have multiple Rupee junk items in view and acquire a larger Rupee, the remaining junk Rupees will rapidly flash through the different types as the Rupee counter increments.

Edit: Upon further thought, maybe just always do `lastChosenAt` at the start of `CurrentJunkItem`. No need to implement it three separate times.

Eblo (changes requested):

garrettjoecox:
Going to close this out, going with the alternative more simple refinements in #1452 

In the future if there is enough demand I could see us introducing a small sub-menu with just checkboxes per item in the junk rotation, but I'm cautious about adding the level of granularity seen here, as it will just overwhelm most people

PR #1333, opened 2025-11-11
[Enhancement] Item Tracker Rework

Caladius:
Reworking the Item Tracker to allow very fine tuned customization options.

Heavy WIP but its returned to the point where the base game is tracked. Needs testing and I welcome any PR's to the branch cleaning up code or making it flow better.

My next trek is adding full Rando Item support.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4649158280.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4649187394.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4649229320.zip)


mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTracker.cpp
```diff
@@ -0,0 +1,384 @@
...
+            case RI_TINGLE_MAP_STONE_TOWER:
+                alpha = CHECK_WEEKEVENTREG(WEEKEVENTREG_TINGLE_MAP_BOUGHT_STONE_TOWER) ? 1 : 0.4f;
+                break;
+            case RI_TRIFORCE_PIECE:
+                alpha = gSaveContext.save.shipSaveInfo.rando.foundTriforcePieces > 0 ? 1 : 0.4f;
+                break;
+            default:
+                break;
+        }
+        randoImageObject.textureColor.w = alpha;
```
Eblo: With every condition being just different enough and non-contiguous, I don't see a way to cleanly simplify this switch-case block to be more in line with what the boss souls conditions do.

One thing could be simplified is the fact that the alpha values are always 1 or 0.4f, which is the default value for alpha anyway. What if we declare something like `bool itemEnabled = false;`, and then change the case statements to be something like `itemEnabled = CHECK_WEEKEVENTREG(WEEKEVENTREG_33_01);`.

Then here, after the block, we could call `randoImageObject.textureColor.w = itemEnabled ? 1.0f : 0.4f;`

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTracker.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+        randoImageObject.textureId = Ship::Context::GetInstance()->GetWindow()->GetGui()->GetTextureByName(
+            (const char*)Rando::StaticData::GetIconTexturePath((RandoItemId)itemId));
+        randoImageObject.textureDimensions =
+            ImVec2(46.0f, itemId >= RI_OWL_CLOCK_TOWN_SOUTH && itemId <= RI_OWL_ZORA_CAPE ? 32.0f : 46.0f);
```
Eblo: I'm seeing a lot of `46.0f` in your code, which looks to be the common texture dimension. I'm going to say throw a `#define ITEM_TEXTURE_SIZE 46.0f` somewhere.

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTracker.cpp
```diff
@@ -0,0 +1,469 @@
...
+            case 1:
+                imageObject.textureColor.w = CHECK_DUNGEON_ITEM(DUNGEON_MAP, dungeonIndex) ? 1.0f : 0.4f;
+                break;
+            case 2:
+                imageObject.textureColor.w = DUNGEON_KEY_COUNT(dungeonIndex) > 0 ? 1.0f : 0.4f;
+                break;
+            case 3:
+                imageObject.textureColor.w = CHECK_DUNGEON_ITEM(DUNGEON_BOSS_KEY, dungeonIndex) ? 1.0f : 0.4f;
+                break;
+        }
```
Eblo, line 282: Same recommendation as the above regarding `isItemEnabled` and the resulting alpha value.

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTracker.cpp
```diff
@@ -0,0 +1,469 @@
...
+    std::string countStr = "";
+    if (isRandoItem) {
+        if (itemId == RI_TRIFORCE_PIECE) {
+            int16_t maxPieces = gSaveContext.save.shipSaveInfo.rando.randoSaveOptions[RO_TRIFORCE_PIECES_REQUIRED];
+            countStr = std::to_string(gSaveContext.save.shipSaveInfo.rando.foundTriforcePieces).c_str();
+            countStr += "/";
+            if (maxPieces == 1000) {
+                countStr += "1k";
+            } else {
+                countStr = std::to_string(maxPieces).c_str();
```
Eblo: If there aren't 1000 Triforce pieces, you won't actually see your current count.
```suggestion
                countStr += std::to_string(maxPieces).c_str();
```

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTracker.cpp
```diff
@@ -0,0 +1,469 @@
...
+                countStr = std::to_string(gSaveContext.save.saveInfo.playerData.rupees).c_str();
+                break;
+            case ITEM_PICTOGRAPH_BOX:
+                countStr = CHECK_QUEST_ITEM(QUEST_PICTOGRAPH) ? "1" : "0";
+                break;
+            case ITEM_SKULL_TOKEN_SWAMP:
+            case ITEM_SKULL_TOKEN_OCEAN:
+                countStr = std::to_string(Inventory_GetSkullTokenCount(
+                                              itemId == ITEM_SKULL_TOKEN_SWAMP ? SCENE_KINSTA1 : SCENE_KINDAN2))
+                               .c_str();
```
Eblo, line 326: Could we show the current/max notation here like Triforce pieces?

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTracker.cpp
```diff
@@ -0,0 +1,469 @@
...
+                break;
+            case ITEM_WOODFALL_STRAY_FAIRY:
+            case ITEM_SNOWHEAD_STRAY_FAIRY:
+            case ITEM_GREAT_BAY_STRAY_FAIRY:
+            case ITEM_STONE_TOWER_STRAY_FAIRY:
+                countStr = std::to_string(
+                               gSaveContext.save.saveInfo.inventory.strayFairies[itemId - ITEM_WOODFALL_STRAY_FAIRY])
+                               .c_str();
```
Eblo, line 336: Ditto tokens, but with the added complication of having a configurable required max in rando.

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTracker.cpp
```diff
@@ -0,0 +1,469 @@
...
+    ImGui::PushStyleVar(ImGuiStyleVar_WindowRounding, 4.0f);
+
+    std::vector<std::vector<TrackerItemListObject>*> windows = {
+        &BenGui::mItemTrackerWindow->namedItemWindows,
+        &BenGui::mItemTrackerWindow->randoItemWindows,
+    };
+
+    uint32_t windowIndex = TRACKER_MAIN;
+    for (auto* window : windows) {
+        if ((*window).empty()) {
```
Eblo: If I'm understanding right, `auto* window` clarifies that window is a pointer of some type, in which case, this might be cleaner:

```suggestion
        if (window->empty()) {
```

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTracker.h
```diff
@@ -0,0 +1,74 @@
...
+    ITEM_BOTTLE_1 = 0x100,
+    ITEM_BOTTLE_2,
+    ITEM_BOTTLE_3,
+    ITEM_BOTTLE_4,
+    ITEM_BOTTLE_5,
+    ITEM_BOTTLE_6,
+} TrackerBottleSlots;
+
+typedef enum {
+    ITEM_SKULL_TOKEN_SWAMP = 0x106,
```
Eblo: I don't think I see these enum types referenced. Since one's starting value continues from the previous, do they need to be separate enums instead of a single one?

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTrackerSettings.cpp
```diff
@@ -0,0 +1,611 @@
...
+
+void CreateMainTrackerWindow() {
+    // TrackerItemListObject initMainObject = {
+    //     .windowName = "Main",
+    //     .columnLength = 6,
+    //     .windowScale = 1.0f,
+    //     .windowOpacity = 0.5f,
+    // };
+    // BenGui::mItemTrackerWindow->namedItemWindows.push_back(initMainObject);
+}
```
Eblo: Still needed?

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTrackerSettings.cpp
```diff
@@ -0,0 +1,611 @@
...
+    UIWidgets::Tooltip(GetItemTrackerItemName(itemId, isRandoItem).c_str());
+}
+
+void DrawItemList(std::string listName, int columns) {
+    bool isRandoItem = false;
+    if (ImGui::BeginChild(listName.c_str(), ImVec2(0, 0),
+                          ImGuiChildFlags_AlwaysAutoResize | ImGuiChildFlags_AutoResizeX |
+                              ImGuiChildFlags_AutoResizeY)) {
+        if (ImGui::BeginTable(listName.c_str(), columns)) {
+            ImVec2 framePadding = ImVec2(listName == "Songs" ? 8.0f : 0, 0);
```
Eblo: I see this 8.0f used in multiple distant pieces. Meaning... `#define ITEM_SONG_PADDING 8.0f`

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTrackerSettings.cpp
```diff
@@ -0,0 +1,611 @@
...
+    { "Stray Fairies", { ITEM_WOODFALL_STRAY_FAIRY, ITEM_STONE_TOWER_STRAY_FAIRY, 4 } },
+    { "Dungeon", { ITEM_WOODFALL_DUNGEON_MAP, ITEM_STONE_TOWER_KEY_BOSS, 4 } },
+};
+
+std::map<std::string, std::tuple<int16_t, int16_t, int16_t>> randoItemLists = {
+    { "Frogs", { RI_FROG_BLUE, RI_FROG_WHITE, 4 } },
+    { "Boss Souls", { RI_SOUL_GOHT, RI_SOUL_TWINMOLD, 5 } },
+    { "Owl Statues", { RI_OWL_CLOCK_TOWN_SOUTH, RI_OWL_ZORA_CAPE, 5 } },
+    { "Tingle Maps", { RI_TINGLE_MAP_CLOCK_TOWN, RI_TINGLE_MAP_WOODFALL, 6 } },
+    { "Misc", { RI_TRIFORCE_PIECE, RI_TRIFORCE_PIECE, 1 } },
```
Eblo, line 49: With there being comparators that check the list names, think we could define the names somewhere?

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTrackerSettings.cpp
```diff
@@ -0,0 +1,611 @@
...
+    } else if (listName == "Stray Fairies") {
+        for (int j = std::get<0>(list); j <= std::get<1>(list); j++) {
+            int16_t strayFairy = ITEM_WOODFALL_STRAY_FAIRY + (j - ITEM_WOODFALL_STRAY_FAIRY);
+            itemObject.itemList.push_back(strayFairy);
+        }
+    } else if (listName == "Dungeon") {
+        for (int j = std::get<0>(list); j <= std::get<1>(list); j++) {
+            int16_t dungeonItem = ITEM_WOODFALL_DUNGEON_MAP + (j - ITEM_WOODFALL_DUNGEON_MAP);
+            itemObject.itemList.push_back(dungeonItem);
+        }
```
Eblo: These are all almost exactly the same, aside from the min and max values, which are defined in the very list referenced here. This could be condensed down to:

```suggestion
    if (listName == "Bottles" || listName == "Tokens" || listName == "Stray Fairies" || listName == "Dungeon") {
        int16_t baseValue = std::get<0>(list);
        int16_t maxValue = std::get<1>(list);
        for (int j = baseValue; j <= maxValue; j++) {
            itemObject.itemList.push_back(j);
        }
```

mm/2s2h/ShipUtils.cpp
```diff
@@ -67,12 +67,99 @@ std::array<const char*, 22> miscellaneousTextures = {
...
+    { ITEM_SONG_OATH, "Oath to Order" },
+
+    // Quest
+    { ITEM_REMAINS_ODOLWA, "Odolwa's Remains" },
+    { ITEM_REMAINS_GOHT, "Goht's Remains" },
+    { ITEM_REMAINS_GYORG, "Gyorg's Remains" },
+    { ITEM_REMAINS_TWINMOLD, "Twinmold's Remains" },
+    { ITEM_SWORD_KOKIRI, "Kokiri Sword" },
+    { ITEM_SHIELD_HERO, "Hero's Shield" },
+    { ITEM_WALLET_ADULT, "Adult Wallet" },
```
Eblo, line 148: As is, these items will always have these names even if the player has their upgraded forms. Conversely, the icons do update. Maybe these could just be generic, i.e. "Sword", "Shield", "Wallet"?

Eblo (changes requested):
Misc feedback
 
- Labels for the item name. Intentionally excluded for spacing reasons?
- The current tracker has a setting to only show when the game is paused. Lost in translation?
- Everything from frogs onward appears in a separate window from previous items

Image

Caladius:
> Misc feedback
> 
>     * Labels for the item name. Intentionally excluded for spacing reasons?
> 
>     * The current tracker has a setting to only show when the game is paused. Lost in translation?
> 
>     * Everything from frogs onward appears in a separate window from previous items
> 
> 
> Image

The Rando specific items are separated from the Vanilla item pool by design. They go into a Rando assigned window vector to ensure they are handled appropriately.

I was not porting the Pause menu support to this version, maybe another PR later.

The item labels I'll see about adding but common complaint was seeing just a wall of text over an item instead of hovering over it for a simple tooltip, I did plan on coloring the small keys but didnt like the color they showed when done.

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTrackerSettings.cpp
Caladius: At the top of ItemTrackerSettings.cpp theres a vector with the names but since Im using a map here I'm strictly using that to define an order.

mm/2s2h/ShipUtils.cpp
Caladius: Thats fine for now, may need to revisit this in the future.

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTracker.h
Caladius: i honest don't know why i separated them in the first place. There was a reason for it but maybe because Bottles were such a special case that I left it on its own island?

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTracker.cpp
Caladius: This isnt merged in yet right? When it does we can just update the maxValue to wherever  you end up defining it

Eblo: The configurable Stray Fairy max is in. Tokens are not. If this gets in first, the token max PR will handle the configurable max token tracking.

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTrackerSettings.cpp
```diff
@@ -0,0 +1,585 @@
...
+        if (itemId >= ITEM_BOTTLE_1 && itemId <= ITEM_BOTTLE_6) {
+            itemName = "Bottle";
+        } else if (itemId == ITEM_SKULL_TOKEN_SWAMP) {
+            itemName = "Swamp Token";
+        } else if (itemId == ITEM_SKULL_TOKEN_OCEAN) {
+            itemName = "Ocean Token";
+        } else if (itemId == ITEM_CLOCK_TOWN_STRAY_FAIRY) {
+            itemName = "Clock Town Stray Fairy";
+        } else if (itemId >= ITEM_WOODFALL_STRAY_FAIRY && itemId <= ITEM_STONE_TOWER_STRAY_FAIRY) {
+            itemName = dungeonPrefix[(itemId - ITEM_WOODFALL_STRAY_FAIRY) / 4] + " Stray Fairy";
```
Eblo: For some reason, this does not work. All of the dungeon stray fairies show "Woodfall Stray Fairy" even though the color and counts correctly track with the associated dungeon. Keys, maps, and compasses correctly show the dungeon name even though they do exactly the same thing as this. wut

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTrackerSettings.cpp
```diff
@@ -0,0 +1,585 @@
...
+                case ITEM_GREAT_BAY_DUNGEON_MAP:
+                case ITEM_STONE_TOWER_DUNGEON_MAP:
+                    itemName = dungeonPrefix[(itemId - ITEM_WOODFALL_DUNGEON_MAP) / 4] + " Dungeon Map";
+                    break;
+                case ITEM_WOODFALL_DUNGEON_COMPASS:
+                case ITEM_SNOWHEAD_DUNGEON_COMPASS:
+                case ITEM_GREAT_BAY_DUNGEON_COMPASS:
+                case ITEM_STONE_TOWER_DUNGEON_COMPASS:
+                    itemName = dungeonPrefix[(itemId - ITEM_WOODFALL_DUNGEON_COMPASS) / 4] + " Compass";
+                    break;
```
Eblo, line 104: This seems correct, but maps and compasses' labels are swapped in practice. wut again

Image

Image

Eblo (changes requested):
- I missed this before, but it looks like the item tracker settings are not persisted. If you close and reopen 2ship, you will have to reconfigure your tracker.
- Trade slots are a little wonky. They seem to always display the earliest item's name (Moon's Tear, Room Key, Letter to Kafei) regardless of what item you currently have, but the icon itself does appropriately update. In rando it's further complicated with how trade slots work, as the tracker only counts what is currently assigned to the slot

Image

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTrackerSettings.cpp
Caladius: fix coming up for this soon.

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTrackerSettings.cpp
Caladius: So this is a fun one, the icons are actually swapped in the vanilla code too. I swapped the icons so they lined up but forgot to figure out a text name solution lol

Caladius:
> * I missed this before, but it looks like the item tracker settings are not persisted. If you close and reopen 2ship, you will have to reconfigure your tracker.
> 
>     * Trade slots are a little wonky. They seem to always display the earliest item's name (Moon's Tear, Room Key, Letter to Kafei) regardless of what item you currently have, but the icon itself does appropriately update. In rando it's further complicated with how trade slots work, as the tracker only counts what is currently assigned to the slot
> 
> 
> Image

I have not gotten to the point of save/load. This is next once everything else is working.

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTrackerSettings.cpp
```diff
@@ -457,6 +458,90 @@ void ApplyDefaultItemPreset() {
...
 }
 
+void SaveItemTrackerLayout() {
+    std::vector<std::vector<TrackerItemListObject>*> trackerWindows = {
+        &BenGui::mItemTrackerWindow->namedItemWindows,
+        &BenGui::mItemTrackerWindow->randoItemWindows,
+    };
+
+    auto allConfig = Ship::Context::GetInstance()->GetConfig()->GetNestedJson();
+    if (allConfig.find("Window.ItemTrackerLayout") == allConfig.end()) {
```
Eblo: I don't think all of this should be under `Window`, as that is where we store pertinent LUS-level variables. What do you say about having it be "CVars.ItemTrackerLayout"? It would functionally be creating a CVar without going through the typical route.

Eblo (commented):
Upgrade/fallback names and saving process LGTM. Just have one question and still need the map/compass swap and I'd say we're golden.

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTrackerSettings.cpp
Caladius: My fear with placing it in CVars is that the Defaults preset clears the ItemTracker section of the CVars list. By placing it in Windows I was hoping to avoid presets deleting layouts.

I can move it if you prefer, just my reasoning for putting there to begin with.

Eblo: We're going go with just making this its own top-level config block, which will depend on https://github.com/Kenix3/libultraship/pull/956

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTrackerSettings.cpp
```diff
@@ -466,11 +466,14 @@ void SaveItemTrackerLayout() {
     };
 
     auto allConfig = Ship::Context::GetInstance()->GetConfig()->GetNestedJson();
-    if (allConfig.find("Window.ItemTrackerLayout") == allConfig.end()) {
-        allConfig["Window.ItemTrackerLayout"] = nlohmann::json::object();
+    if (allConfig.find("UConfs") == allConfig.end()) {
```
Eblo: Per @garrettjoecox I'm leaning more towards now just having Item Tracker be its own top level block, rather than having a top-level block for miscellaneous things.
```suggestion
    if (allConfig.find("ItemTrackerLayout") == allConfig.end()) {
```

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTrackerSettings.cpp
```diff
@@ -500,17 +503,17 @@ void SaveItemTrackerLayout() {
         windowType++;
     }
 
-    Ship::Context::GetInstance()->GetConfig()->SetBlock("Window.ItemTrackerLayout", itemTrackerConfig);
+    Ship::Context::GetInstance()->GetConfig()->SetBlock("UConfs.ItemTrackerLayout", itemTrackerConfig);
```
Eblo:
```suggestion
    Ship::Context::GetInstance()->GetConfig()->SetBlock("ItemTrackerLayout", itemTrackerConfig);
```

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTrackerSettings.cpp
```diff
@@ -0,0 +1,695 @@
...
+    auto allConfig = Ship::Context::GetInstance()->GetConfig()->GetNestedJson();
+    if (allConfig.find("UConfs") == allConfig.end()) {
+        allConfig["UConfs"] = nlohmann::json::object();
+        Ship::Context::GetInstance()->GetConfig()->SetBlock("UConfs", allConfig["UConfs"]);
+        Ship::Context::GetInstance()->GetConfig()->Save();
+    }
+    allConfig["UConfs.ItemTrackerLayout"] = nlohmann::json::object();
+
+    auto& itemTrackerConfig = allConfig["UConfs.ItemTrackerLayout"];
+    itemTrackerConfig = nlohmann::json::object();
```
Eblo: Per the most recent discussion, I think we should just have this be its own top-level item.
```suggestion
    auto& itemTrackerConfig = allConfig["ItemTrackerLayout"] = nlohmann::json::object();
```

Eblo:
```suggestion
    auto itemTrackerConfig = nlohmann::json::object();
```

mm/2s2h/Enhancements/Trackers/ItemTracker/ItemTrackerSettings.cpp
```diff
@@ -0,0 +1,695 @@
...
+    if (allConfig.find("UConfs") == allConfig.end()) {
+        return;
+    }
+
+    auto& windowConfig = allConfig["UConfs"];
+    if (windowConfig.find("ItemTrackerLayout") == windowConfig.end()) {
+        return;
+    }
+
+    auto& itemTrackerConfig = windowConfig["ItemTrackerLayout"];
```
Eblo:
```suggestion
    auto allConfig = Ship::Context::GetInstance()->GetConfig()->GetNestedJson();
    if (allConfig.find("ItemTrackerLayout") == allConfig.end()) {
        return;
    }

    auto& itemTrackerConfig = allConfig["ItemTrackerLayout"];
```

Eblo (changes requested):

Eblo (approved):

PR #1344, opened 2025-11-14
Reorganized the Close Menu, Quit, and Reset buttons to match the layout in SoH.

Glought:
I reorganized the menu buttons to match the placement and order used in SoH. 

I also port over the SohModals system and renamed it to BenModals. With this change, when the player presses the Quit button, a confirmation message will appear asking, ‘Are you sure you want to quit 2S2H?’ with two options: Quit and Cancel, like how it works in SoH.

Before:
2ship Original Menu Buttons

After:
2ship reorganized Menu Buttons

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4584058765.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4584063122.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4584118528.zip)


mm/2s2h/BenGui/BenModals.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include <libultraship/libultraship.h>
```
Eblo: Do you need this entire header? This is a rather hefty one that is not commonly used directly.

Eblo (commented):
Nice. No complaints, just one blocking question. Modals are a welcome addition as well.

mm/2s2h/BenGui/BenModals.cpp
```diff
@@ -0,0 +1,87 @@
+#include "BenModals.h"
+#include <imgui.h>
+#include <vector>
+#include <string>
+#include <libultraship/bridge.h>
+#include <libultraship/libultraship.h>
```
Eblo: Ditto, but also including the same header here is redundant when `BenModals.h` includes it as well.

Glought:
I cleaned up the unneeded `"Includes` from `BenModals.h` and `BenModals.cpp` .

Eblo (approved):

PR #1346, opened 2025-11-15
[Rando] Shuffle Tree Drops

Caladius:
This shuffles the drops from Actor Spawned Trees.

Still WIP but pushing up for visibility.

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4676070283.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4676161871.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4676684099.zip)

PR #1348, opened 2025-11-16
[Rando] Adds Man in the Tree Rupees

Caladius:
This adds both Red Rupees that spawn after committing attempted murder on the Man in the Tree in Termina Field to the shuffle pool under Freestanding Shuffle.

These rupees do not have an actorlistindex.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4582820415.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4582852159.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4582941096.zip)


Eblo (commented):
From the player's standpoint, I feel like this should belong in Tree Shuffle rather than Freestanding. In practice it's not really any different beyond this particular tree having two drops. You still bonk a tree. The item preview+boomerang aspect of Tree Shuffle might complicate that since these are not spawned in advance, but I feel like not making these Tree checks is confusing.

Caladius:
> From the player's standpoint, I feel like this should belong in Tree Shuffle rather than Freestanding. In practice it's not really any different beyond this particular tree having two drops. You still bonk a tree. The item preview+boomerang aspect of Tree Shuffle might complicate that since these are not spawned in advance, but I feel like not making these Tree checks is confusing.

Not a problem, I added these into Treesanity PR https://github.com/HarbourMasters/2ship2harkinian/pull/1346. will close this.

PR #1350, opened 2025-11-18
Reset Invadepoh state

Eblo:
Fixes an issue where the invasion would not start if re-attempted during the same play session.

`z_en_invadepoh.c` has a static variable, `sInvasionState`, for tracking the alien invasion. If the player triggers the invasion once, it will have some value besides `INVASION_STATE_NONE`, which means that the setup in `EnInvadepoh_InvasionHandler_SetInitialInvasionState` will not run on repeat attempts, thus no invasion happens.

The following should behave correctly:

- Leave the invasion in progress and come back immediately to see the invasion still in progress
- Leave the invasion in progress and come back hours later to see the empty barn
- Successfully fend off the invasion, then come back in another cycle and verify that the invasion commences orrectly
- Fail the invasion, then come back in another cycle and verify that the invasion commences correctly

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4631246652.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4631372905.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4631657756.zip)

### DIFF
--- a/mm/src/overlays/actors/ovl_En_Invadepoh/z_en_invadepoh.c
+++ b/mm/src/overlays/actors/ovl_En_Invadepoh/z_en_invadepoh.c
@@ -5490,7 +5490,7 @@ void EnInvadepoh_Reset() {
     sInvasionState = INVASION_STATE_NONE;
     sRewardFinished = false;
     sAliensExtremeThreat = false;
-    for (int i = 0; i < ALIEN_COUNT; i ++) {
+    for (int i = 0; i < ALIEN_COUNT; i++) {
         sAliens[i] = NULL;
         sAlienStateFlags[i] = 0;
     }
--- a/mm/src/overlays/actors/ovl_En_Invadepoh/z_en_invadepoh.c
+++ b/mm/src/overlays/actors/ovl_En_Invadepoh/z_en_invadepoh.c
@@ -5489,4 +5489,9 @@ void EnInvadepoh_Cremia_Draw(Actor* thisx, PlayState* play) {
 void EnInvadepoh_Reset() {
     sInvasionState = INVASION_STATE_NONE;
     sRewardFinished = false;
+    sAliensExtremeThreat = false;
+    for(int i = 0; i < ALIEN_COUNT; i ++) {
+        sAliens[i] = NULL;
+        sAlienStateFlags[i] = 0;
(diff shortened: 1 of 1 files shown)


mm/src/overlays/actors/ovl_En_Invadepoh/z_en_invadepoh.c
```diff
@@ -5483,3 +5485,8 @@ void EnInvadepoh_Cremia_Draw(Actor* thisx, PlayState* play) {
 
     CLOSE_DISPS(play->state.gfxCtx);
 }
+
+void EnInvadepoh_Reset() {
+    sInvasionState = INVASION_STATE_NONE;
+    sRewardFinished = false;
+}
```
balloondude2, line 5497: I think from my initial attempt, `sAlienStateFlags[ALIEN_COUNT]` also needs to be reset or the Aliens don't appear one at a time on the subsequent cycle. Haven't had a chance to verify that this happens on this PR though

balloondude2 (changes requested):

mm/src/overlays/actors/ovl_En_Invadepoh/z_en_invadepoh.c
Eblo: Those flags are not initialized as-is, meaning they have no determinate default state even on a first attempt. I haven't seen the described behavior when testing the bug and this fix, but who knows what flags I had when playing around.

We could ensure the flags are 0 on reset, but then the question arises of what they are on a first attempt.

Eblo: I have now learned that it depends :shipit: https://stackoverflow.com/questions/12750265/what-is-the-default-value-of-members-of-a-char-array-in-c

balloondude2 (approved):
If the dog was there for you on the subsequent cycle, that's good enough for me.

Eblo:
Just ran through the four scenarios listed and found no issues.

PR #1351, opened 2025-11-20
[Enhancement] QuickStart Menu

Caladius:
From fresh install to playing in less than 15 seconds. The QuickStart menu will ask you a few questions about what you want to do then creates a file with those options and drops you in game immediately.

The menu will also give you the option to set a curated preset that will enable a majority of the most common Quality of Life enhancements.

I hijacked the F1 hotkey to take you straight to the QuickStart menu from wherever in the Menu you are currently or even if the Menu is not opened. (This may need a change in LUS but I cannot confirm yet)

image

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4665086070.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4665165177.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4666191447.zip)


mm/2s2h/BenGui/BenGui.cpp
```diff
@@ -164,9 +165,8 @@ void SetupGuiElements() {
         "gWindows.CheckTrackerSettings", "Check Tracker Settings");
     gui->AddGuiWindow(mRandoCheckTrackerSettingsWindow);
 
-    mModalWindow = std::make_shared<BenModalWindow>("gWindows.ModalWindow", "Modal Window");
-    gui->AddGuiWindow(mModalWindow);
-    mModalWindow->Show();
+    mQuickStartMenu = std::make_shared<Ship::QuickStart>("gWindows.QuickStart.", "QuickStart Menu");
```
Eblo: The trailing period results in a blank attribute under the QuickStart config section.
```suggestion
    mQuickStartMenu = std::make_shared<Ship::QuickStart>("gWindows.QuickStart", "QuickStart Menu");
```

mm/2s2h/BenGui/QuickStart.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+namespace BenGui {
+extern std::shared_ptr<BenMenu> mBenMenu;
+extern std::shared_ptr<Ship::QuickStart> mQuickStartMenu;
+} // namespace BenGui
+
+using namespace BenGui;
+namespace Ship {
+
+std::vector<uint8_t> ConvertNameSet(const char* input) {
```
Eblo, line 66: There is no guard against special characters here. I was able to input a bunch of dollar signs `$`. The result was that the file name, in data, was a bunch of forward slashes, blank on file select, and in-game it seems to have broken dialog to be several different values: "d o h a v e a r", "c a n c i e s t", "d e a r e s e r", etc.

mm/2s2h/BenGui/QuickStart.cpp
```diff
@@ -0,0 +1,462 @@
...
+                result.push_back(D_808141F0[62]);
+                continue;
+            case '-':
+                result.push_back(D_808141F0[63]);
+                continue;
+            case ' ':
+                result.push_back(D_808141F0[64]);
+                continue;
+        }
+        result.push_back(0xFF);
```
Eblo, line 101: I assume this is the terminator character for the name string, but is this defined anywhere?

mm/2s2h/BenGui/QuickStart.cpp
```diff
@@ -0,0 +1,462 @@
...
+    if (quickStartOptions.questID == QUICK_START_NONE ||
+        (quickStartOptions.questID == SAVETYPE_RANDO && (quickStartOptions.rando.shuffleSet == QUICK_START_NONE ||
+                                                         quickStartOptions.rando.logicOption == QUICK_START_NONE))) {
+        ImGui::TextColored(TEXT_COLOR(COLOR_ORANGE), "Not Ready");
+        return;
+    }
+    CenterFullWidthSeparatorText("Ready to start?", "Half");
+    if (ImGui::BeginTable("NameEntry", 2, ImGuiTableFlags_SizingStretchSame)) {
+        ImGui::TableNextColumn();
+        UIWidgets::InputString("##playername", &fileNameEntry,
```
Eblo: There needs to be some way to enforce an 8 character limit here to match what the game allows.

mm/2s2h/BenGui/QuickStart.cpp
```diff
@@ -0,0 +1,462 @@
...
+                                                     ? COLOR_GREEN
+                                                     : COLOR_GREY })) {
+                    quickStartOptions.rando.logicOption = QUICK_START_RANDO_GLITCHLESS;
+                }
+                ImGui::TableNextColumn();
+                if (UIWidgets::Button("No Logic",
+                                      { .color = quickStartOptions.rando.logicOption == QUICK_START_RANDO_NOLOGIC
+                                                     ? COLOR_GREEN
+                                                     : COLOR_GREY })) {
+                    quickStartOptions.rando.logicOption = QUICK_START_RANDO_NOLOGIC;
```
Eblo:
```suggestion
                                      { .color = quickStartOptions.rando.logicOption == QUICK_START_RANDO_NO_LOGIC
                                                     ? COLOR_GREEN
                                                     : COLOR_GREY })) {
                    quickStartOptions.rando.logicOption = QUICK_START_RANDO_NO_LOGIC;
```

mm/2s2h/BenGui/QuickStart.h
```diff
@@ -0,0 +1,42 @@
...
+#include "UIWidgets.hpp"
+
+typedef enum {
+    QUICK_START_NONE = -1,
+    QUICK_START_PRESET_DEFAULT,
+    QUICK_START_PRESET_ENHANCED,
+    QUICK_START_RANDO_BASE,
+    QUICK_START_RANDO_ALL,
+    QUICK_START_RANDO_GLITCHLESS,
+    QUICK_START_RANDO_NOLOGIC,
```
Eblo:
```suggestion
    QUICK_START_RANDO_NO_LOGIC,
```

mm/2s2h/BenGui/BenGui.cpp
```diff
@@ -164,9 +165,8 @@ void SetupGuiElements() {
         "gWindows.CheckTrackerSettings", "Check Tracker Settings");
     gui->AddGuiWindow(mRandoCheckTrackerSettingsWindow);
 
-    mModalWindow = std::make_shared<BenModalWindow>("gWindows.ModalWindow", "Modal Window");
```
Eblo, line 167: Did you mean to completely remove the modal window?

mm/2s2h/BenGui/BenGui.cpp
```diff
@@ -75,13 +77,12 @@ void SetupGuiElements() {
     style.Colors[ImGuiCol_MenuBarBg] = UIWidgets::ColorValues.at(UIWidgets::Colors::DarkGray);
 
     mBenMenuBar = std::make_shared<BenMenuBar>(CVAR_MENU_BAR_OPEN, CVarGetInteger(CVAR_MENU_BAR_OPEN, 0));
-    gui->SetMenuBar(std::reinterpret_pointer_cast<Ship::GuiMenuBar>(mBenMenuBar));
```
Eblo, line 78: If we're replacing the menu bar entirely, then we should scrub `mBenMenuBar` here.

mm/2s2h/BenGui/QuickStart.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+Please keep the flow simple and easy for the user. If you plan to add more to this, keep simplicity in mind.
+*/
+
+// ImVec4 Colors
+#define COLOR_WHITE UIWidgets::Colors::White
+#define COLOR_GREY UIWidgets::Colors::Gray
+#define COLOR_GREEN UIWidgets::Colors::Green
+#define COLOR_ORANGE UIWidgets::Colors::Orange
+#define COLOR_RED UIWidgets::Colors::Red
```
Eblo, line 39: Do these really need macro'd? The names are the same and descriptive as-is.

mm/2s2h/BenGui/QuickStart.cpp
```diff
@@ -0,0 +1,462 @@
...
+
+#define TEXT_COLOR(color) UIWidgets::ColorValues.at(color)
```
Eblo, line 41: Ditto, more or less.

mm/2s2h/BenGui/QuickStart.cpp
```diff
@@ -0,0 +1,462 @@
...
+    CVarSetInteger(Rando::StaticData::Options[RO_HINTS_OATH_TO_ORDER].cvar, 1);
+    CVarSetInteger(Rando::StaticData::Options[RO_HINTS_HOOKSHOT].cvar, 1);
+
+    // Base Game Shuffles
+    CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_OWL_STATUES].cvar, 1);
+    CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_SHOPS].cvar, 1);
+    CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_TINGLE_SHOPS].cvar, 1);
+    CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_COWS].cvar, 1);
+    CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_GOLD_SKULLTULAS].cvar, 1);
+    CVarSetInteger(Rando::StaticData::Options[RO_MINIMUM_STRAY_FAIRIES].cvar, 1);
```
Eblo: This is not a binary toggle, but a configurable minimum. This means that any seeds generated from Quick Start will only require 1 Stray Fairy for each Great Fairy check. I don't think that makes for a good default experience.

mm/2s2h/BenGui/QuickStart.cpp
```diff
@@ -0,0 +1,462 @@
...
+    CVarSetInteger(Rando::StaticData::Options[RO_HINTS_GOSSIP_STONES].cvar, 1);
+    CVarSetInteger(Rando::StaticData::Options[RO_HINTS_BOSS_REMAINS].cvar, 1);
+    CVarSetInteger(Rando::StaticData::Options[RO_HINTS_OATH_TO_ORDER].cvar, 1);
+    CVarSetInteger(Rando::StaticData::Options[RO_HINTS_HOOKSHOT].cvar, 1);
+
+    // Base Game Shuffles
+    CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_OWL_STATUES].cvar, 1);
+    CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_SHOPS].cvar, 1);
+    CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_TINGLE_SHOPS].cvar, 1);
+    CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_COWS].cvar, 1);
```
Eblo: Does Cow Shuffle make sense as "base game"? This is a heavily subjective question, but I'd see it as more of an advanced option since it makes use of a rather obscure feature in the game.

mm/2s2h/BenGui/QuickStart.cpp
```diff
@@ -0,0 +1,462 @@
...
+    if (quickStartOptions.rando.shuffleSet == QUICK_START_RANDO_ALL) {
+        CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_POT_DROPS].cvar, 1);
+        CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_CRATE_DROPS].cvar, 1);
+        CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_BARREL_DROPS].cvar, 1);
+        CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_SNOWBALL_DROPS].cvar, 1);
+        CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_GRASS_DROPS].cvar, 1);
+        CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_FROGS].cvar, 1);
+        CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_FREESTANDING_ITEMS].cvar, 1);
+        CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_BOSS_SOULS].cvar, 1);
+    }
```
Eblo, line 182: Since this seems to be meant to be exhaustive, that would mean it would need updated with every new shuffle added. I see many (myself included) outright forgetting to update this as that happens. If every `RO_SHUFFLE` option is contiguous in the enum and is a binary setting, it might be better to just iterate them here. Granted, if the first or last one ever changes, this would need to as well, but that's probably still going to be prettier than an ever-increasing list.

Eblo (commented):
The README will need updated to reference Quick Start instead of the menu bar. Heck, it needs updated to reference the Esc menu too lol

mm/2s2h/BenGui/QuickStart.cpp
Caladius: What value would you want here?

mm/2s2h/BenGui/QuickStart.cpp
Caladius: This is not meant to be updated as new options come in. This is set and forget should reflect settings players would expect to be turned on.

As in, once this is merged this list will not be updated again. Open to changes to the initial list.

mm/2s2h/BenGui/QuickStart.cpp
Caladius: Depends on who you ask really. I dont have a preference if it stays or goes.

mm/2s2h/BenGui/QuickStart.cpp
Caladius: Was meant to reduce having to type out the whole UIWidgets::Colors prefix for the large amount of times Id be using it. Honestly, these should probably be defined as such but within ShipUtils somewhere instead of UIWidgets. Since its used in a lot of places maybe a separate PR to clean all of those locations up?

mm/2s2h/BenGui/QuickStart.cpp
Caladius: Same as my response to the colors themselves, something something move to ShipUtils and clean it all up.

mm/2s2h/BenGui/QuickStart.cpp
Eblo: After seeing you renamed it, my concerns are assuaged

mm/2s2h/BenGui/QuickStart.cpp
Eblo: I'm good with that.

mm/2s2h/BenGui/QuickStart.cpp
```diff
@@ -0,0 +1,470 @@
...
+        CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_GRASS_DROPS].cvar, 1);
+        CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_FREESTANDING_ITEMS].cvar, 1);
+        CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_BOSS_SOULS].cvar, 1);
+    }
+
+    Ship::Context::GetInstance()->GetWindow()->GetGui()->SaveConsoleVariablesNextFrame();
+}
+
+void QuickStartQuestInit(const char* fileNameEntry) {
+    static std::string cleanFileName = Ship_RemoveSpecialCharacters(fileNameEntry);
```
Eblo: The `static` keyword means that any subsequent quickstart saves I make will always have the same name in the same runtime.
```suggestion
    std::string cleanFileName = Ship_RemoveSpecialCharacters(fileNameEntry);
```

mm/2s2h/BenGui/QuickStart.cpp
```diff
@@ -0,0 +1,470 @@
...
+    std::vector<uint8_t> playerName = ConvertNameSet(cleanFileName.c_str());
+    FileSelectState* fileSelect = (FileSelectState*)gGameState;
+    SramContext* sramCtx = &fileSelect->sramCtx;
+
+    fileSelect->buttonIndex = (SelectMenuButtonIndex)SaveManager_GetOpenFileSlot() - 1;
+    if (fileSelect->buttonIndex < FS_BTN_SELECT_FILE_1) {
```
Eblo: `Ship_RemoveSpecialCharacters` does do the job of stripping out special characters, but ultimately it just passes in the sanitized string to the file and proceeds to start the game. I think a better UX might be to error out if the sanitized file name reveals a mismatch. What do you think?

```suggestion
    if (cleanFileName != fileNameEntry || fileSelect->buttonIndex < FS_BTN_SELECT_FILE_1) {
```

Eblo (changes requested):

mm/2s2h/BenGui/QuickStart.cpp
```diff
@@ -0,0 +1,470 @@
...
+        CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_GRASS_DROPS].cvar, 1);
+        CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_FREESTANDING_ITEMS].cvar, 1);
+        CVarSetInteger(Rando::StaticData::Options[RO_SHUFFLE_BOSS_SOULS].cvar, 1);
+    }
+
+    Ship::Context::GetInstance()->GetWindow()->GetGui()->SaveConsoleVariablesNextFrame();
+}
+
+void QuickStartQuestInit(const char* fileNameEntry) {
+    std::string cleanFileName = Ship_RemoveSpecialCharacters(fileNameEntry);
```
Eblo, line 188: I see that the change to make this non-static was made, yet I'm still getting the issue where subsequent files will always have the first name I made in the runtime, even if I type something else.

Eblo (changes requested):
Something bizarre happens if you try to make a new file when all 3 slots are already populated. Indeed, no file gets created as expected, but the menu gets into a weird state where it highlights two nonexistent options at the same time. If the player tries to manually create a new save in this state, no file gets created, but if rando is enabled, a seed and spoiler log do get generated albeit are not used.

As mentioned on Discord, there's a bug with using the file delete for slots 2 and 3. The correct file *is* deleted, but the file select menu acts as if slot 1 was deleted.

https://github.com/user-attachments/assets/14c8abde-5e95-4ef4-bf91-37db63ce0b1f

PR #1354, opened 2025-11-21
[Enhancement] Restore OoT tree bonk behavior

Eblo:
Unlike OoT, MM's rolls have damage flags at the end, which allows for destroying pots and grass. This was achieved by changing collision flags during certain frames in a roll. However, the flags are directly assigned instead of OR'd, meaning that if a collision occurs during this phase of the roll (which sets `OC1_HIT`), it will be cleared before it is checked. In effect, trees reached at the end of a roll are not bonked and are moreso slid around instead.

This enhancement changes the damage collision to be OR'd instead, which makes tree bonks feel more OoT-like while maintaining the damaging properties of rolls. This restoration will go well with #1346

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4642344582.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4642442908.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4642854086.zip)


Caladius:
Tested this out, this feels identical to OoT. Bonking trees from a distance is now much more consistent.

mckinlee (approved):

PR #1367, opened 2025-11-24
Locations Tab Enhancements

mckinlee:
Renamed "Locations" tab to "Check Filter" and updated labels ("Included" to "Normal", "Excluded" to "Forced Junk").
Added bulk actions: "Junk All", "Junk Filtered", "Remove Filtered", "Remove All". Buttons change based on filter state.
Code cleanup:
- Removed unnecessary `.c_str()` calls on `std::to_string()` results
- Switched to binary search since the exclusion list is sorted
- Maintain sorted order when adding checks (use `lower_bound` + `insert` instead of `push_back`)
- Fixed iterator usage in the excluded checks loop

Refactored filter UI into a reusable helper so both columns use the same layout.

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4666712038.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4666770674.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4667637934.zip)

### DIFF
--- a/mm/2s2h/Rando/Menu.cpp
+++ b/mm/2s2h/Rando/Menu.cpp
@@ -90,14 +90,13 @@ void SaveExcludedChecks() {
     SortExcludedChecks();
 
     for (auto& data : checkExclusionList) {
-        excludedString += std::to_string(data).c_str();
+        excludedString += std::to_string(data);
         excludedString += ",";
     }
     CVarSetString("gRando.ExcludedChecks", excludedString.c_str());
 }
 
 void LoadExcludedChecks() {
-    std::vector<RandoCheckId> sortedExclusionList;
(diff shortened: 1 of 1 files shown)


mckinlee:
The `.c_str` removal in `DrawStartingItemsTab` wasn't necessary for this PR, but removing it since I was already in the file. `std::to_string()` returns a `std::string`, and `std::string::operator+=` accepts both `const char*` and `const std::string&`. Calling `.c_str()` adds an unnecessary conversion.

Eblo (approved):

PR #1368, opened 2025-11-25
[Rando] Add Check Type Filters to Check Tracker

Caladius:
Various tweaks to the check tracker UX and settings
- Moved check tracker to its own tab
- Introduced check type filter buttons
  - Tweaks from cal's original changes:
    - Only actively shuffled check type buttons will show now, instead of all
    - Amount of buttons per row adjusts to tracker size
- Added "mode" combobox to different check states, hidden, normal, or colored. The significance of normal over colored (white) is for the out of logic checks it will not even run the calculations which may improve performance on lower end systems.

Screenshot 2026-01-17 at 11 55 03 PM

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4677613403.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4677681564.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4677771593.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1781,13 +1781,6 @@ void BenMenu::AddEnhancements() {
                      .Tooltip("Enables the Cosmetic Editor window, allowing you to modify various colors in the game.")
                      .Size(Sizes::Inline));
 
-    // Item Tracker Settings
-    path = { "Enhancements", "Item Tracker", SECTION_COLUMN_1 };
-    AddSidebarEntry("Enhancements", "Item Tracker", 1);
-    AddWidget(path, "Popout Settings", WIDGET_WINDOW_BUTTON)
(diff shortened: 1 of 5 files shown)


mckinlee:
I really like this. I have two recommendations:
-Add hover tooltips for the filter icons. I have a hard time telling which filter is which.
-You do `(RCTYPE_MAX - 1) / 2` to achieve a 10 column layout. You could add instead of subtracting the 1 and then this would get you 2 rows instead of 3 which I feel is more visually appealing. This is purely a opinion, entirely up to you.

Eblo:
From my own testing, works with no fuss. I second the feedback above, especially for tooltips.

garrettjoecox:
Updated this PR & included some cleanup around the check tracker UX. (updated the description to reflect such)

PR #1376, opened 2025-12-02
[Enhancement] Difficulty options to tweak Swamp Boat Archery minigame

JordanLongstaff:
Split the archery minigame hooks up by their appropriate CVars, and added settings for the swamp boat archery minigame's high score as well. It's usually set to 19 by default, but the hook will set it to whatever the CVar value is whenever a new cycle begins.

To cover the bases, I've also made the "Swamp Boat Skip" setting recognize this new CVar.

Also added two more CVars, one for Koume's health and another to make her invincible.

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4817051947.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4817083681.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4817769633.zip)


mm/src/overlays/actors/ovl_En_Tru_Mt/z_en_tru_mt.c
```diff
@@ -337,7 +338,7 @@ void func_80B76924(EnTruMt* this) {
 void func_80B76980(EnTruMt* this, PlayState* play) {
     Player* player = GET_PLAYER(play);
 
-    if (gSaveContext.minigameHiddenScore >= 10) {
+    if (gSaveContext.minigameHiddenScore >= CVarGetInteger("gEnhancements.Minigames.BoatArcheryHealth", 10)) {
```
Eblo: We're trying to keep CVars out of source whenever possible. Let's make this use a VB instead:

```suggestion
    if (GameInteractor_Should(VB_FAIL_BOAT_ARCHERY, gSaveContext.minigameHiddenScore >= 10)) {
```

Then in Archery.cpp, do something like:
```c++
#define HEALTH_CVAR_NAME "gEnhancements.Minigames.BoatArcheryHealth"
#define HEALTH_CVAR CVarGetInteger(HEALTH_CVAR_NAME, 10)

COND_VB_SHOULD(VB_FAIL_BOAT_ARCHERY, HEALTH_CVAR != 10, {
    *should = gSaveContext.minigameHiddenScore >= HEALTH_CVAR;
});
```

mm/2s2h/Enhancements/Minigames/Archery.cpp
```diff
@@ -13,8 +13,10 @@ void EnSyatekiMan_Town_RunGame(EnSyatekiMan* enSyatekiMan, PlayState* play);
 #define SWAMP_CVAR CVarGetInteger(SWAMP_CVAR_NAME, 2180)
 #define TOWN_CVAR_NAME "gEnhancements.Minigames.TownArcheryScore"
 #define TOWN_CVAR CVarGetInteger(TOWN_CVAR_NAME, 50)
+#define BOAT_CVAR_NAME "gEnhancements.Minigames.BoatArcheryScore"
+#define BOAT_CVAR CVarGetInteger(BOAT_CVAR_NAME, 20)
 
-void RegisterArchery() {
+static void RegisterSwampArchery() {
```
Eblo, line 23: Why the change to `static`?

mm/2s2h/Enhancements/Minigames/Archery.cpp
```diff
@@ -51,4 +55,10 @@ void RegisterArchery() {
     });
 }
 
-static RegisterShipInitFunc initFunc(RegisterArchery, { SWAMP_CVAR_NAME, TOWN_CVAR_NAME });
+static void RegisterBoatArchery() {
+    COND_HOOK(AfterEndOfCycleSave, BOAT_CVAR != 20, []() { HS_SET_BOAT_ARCHERY_HIGH_SCORE(BOAT_CVAR - 1); });
+}
```
Eblo, line 78: `AfterEndOfCycleSave` means that this will not apply until the player SoT saves after configuring this setting. Therefore, boat archery in the first cycle will still obey the value that was already set, which can be confusing. It would be better to follow the pattern of the other archery scores and simply run a ShouldActorUpdate hook that pre-empts the vanilla behavior:

```suggestion
static void RegisterBoatArchery() {
    COND_ID_HOOK(ShouldActorUpdate, ACTOR_EN_TRU_MT, BOAT_CVAR != 20, [](Actor* actor, bool* should) {
        EnTruMt* enTruMt = (EnTruMt*)actor;

        if (enTruMt->actionFunc == func_80B76A64) {
            if (gSaveContext.minigameScore >= BOAT_CVAR) {
                HS_SET_BOAT_ARCHERY_HIGH_SCORE(20);
                SET_EVENTINF(EVENTINF_37);
            }
        }
    });
}
```

mm/2s2h/BenGui/BenMenu.cpp
```diff
@@ -1588,6 +1588,22 @@ void BenMenu::AddEnhancements() {
                               "- Skip: Instantly win the race.\n")
                      .DefaultIndex(GoronRaceDifficultyOptions::GORON_RACE_DIFFICULTY_VANILLA)
                      .ComboVec(&goronRaceDifficultyOptions));
+    AddWidget(path, "Swamp Boat Archery High Score", WIDGET_CVAR_SLIDER_INT)
+        .CVar("gEnhancements.Minigames.BoatArcheryScore")
+        .Options(IntSliderOptions()
+                     .Tooltip("Sets the initial target score of the Swamp Boat Archery minigame. The high score will "
+                              "be set to this value minus one at the start of each cycle.")
+                     .Min(10)
+                     .Max(40)
```
Eblo: Any particular reason for these values? Setting the min to 1 would be a nice convenience.

Eblo (changes requested):
TIL Koume has a limit

mm/2s2h/Enhancements/Minigames/Archery.cpp
JordanLongstaff: Because Visual Studio advises me to make these functions that are not externed in any way static. They're just messages, not warnings or errors, but making these functions static ensures they won't be externed unless we change them.

mm/2s2h/BenGui/BenMenu.cpp
JordanLongstaff: I could set the minimum to 1. I don't want to set the maximum _too_ high - maybe 50?

mm/src/overlays/actors/ovl_En_Tru_Mt/z_en_tru_mt.c
JordanLongstaff: I hear you on this, and I'm making the change as we speak. I wondered if an exception would be allowed here just to not make the default comparison redundant.

mm/2s2h/Enhancements/Minigames/Archery.cpp
JordanLongstaff: This actually needs a major rethink if we want the score to update live like that. First of all, TIL that Koume tells you that "your best score" is 19 hits even though you never hit that mark. Second, this needs to be compatible with the Swamp Boat Timesaver hook. Third, this isn't just a "perfect score" like the other archery minigames - it's a running high score that requires preservation. Fourth, it's possible to _increase_ the target value, in which case just comparing it to that setting isn't going to work.

mm/2s2h/BenGui/BenMenu.cpp
JordanLongstaff: Actually, 50's pretty high. At my best, I can't even hit 40. You'd have to be an archery god to hit 50.

mm/2s2h/Enhancements/Minigames/Archery.cpp
```diff
@@ -51,4 +59,31 @@ void RegisterArchery() {
...
+    COND_ID_HOOK(OnActorInit, ACTOR_EN_DNH, BOAT_CVAR != 20, ResetBoatArcheryScore);
+}
+
+static bool ShouldFailBoatArchery() {
+    return gSaveContext.minigameHiddenScore >= BOAT_HEALTH_CVAR;
+}
+
+static void RegisterKoumeHealth() {
+    REGISTER_VB_SHOULD(VB_FAIL_BOAT_ARCHERY, { *should = ShouldFailBoatArchery(); });
+}
```
mckinlee, line 78: I'd use `COND_VB_SHOULD`, conditional on `BOAT_HEALTH_CVAR != 10` since that's the vanilla default. 
```suggestion
static void RegisterKoumeHealth() {
    COND_VB_SHOULD(VB_FAIL_BOAT_ARCHERY, BOAT_HEALTH_CVAR != 10, { *should = ShouldFailBoatArchery(); });
}
```

mm/2s2h/Enhancements/Minigames/Archery.cpp
```diff
@@ -51,4 +59,31 @@ void RegisterArchery() {
...
+
+static void RegisterKoumeInvincible() {
+    COND_VB_SHOULD(VB_KOUME_TAKE_DAMAGE, BOAT_NO_DAMAGE_CVAR, { *should = false; });
+}
+
+static RegisterShipInitFunc initFunc_Swamp(RegisterSwampArchery, { SWAMP_CVAR_NAME });
+static RegisterShipInitFunc initFunc_Town(RegisterTownArchery, { TOWN_CVAR_NAME });
+
+static RegisterShipInitFunc initFunc_Boat(RegisterBoatArchery, { BOAT_CVAR_NAME });
+static RegisterShipInitFunc initFunc_Health(RegisterKoumeHealth);
```
mckinlee: And then make the update path for this `BOAT_HEALTH_CVAR_NAME`. That way Koume's health gets updated immediately.
```suggestion
static RegisterShipInitFunc initFunc_Health(RegisterKoumeHealth, { BOAT_HEALTH_CVAR_NAME });
```

mm/2s2h/Enhancements/Minigames/Archery.cpp
JordanLongstaff: Because the comparison is made inside the hook, we don't have to re-register this hook every time the CVar changes, which could be a lot of times because it's controlled by a slider.

mckinlee: You make a valid point, but the trade off is this still runs every frame during the minigame even when its set to the vanilla default. Your concern during slider drag I don't disagree with, but the cost is negligible since its only during the use of the slider. The overhead at default is my main concern, but I'm not going to say you have to change it. It just goes against the established pattern in most cases I've seen in the codebase.

mckinlee (approved):

Eblo (approved):

PR #1383, opened 2025-12-04
Add option to skip the soaring cutscene when warping

JordanLongstaff:
null

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4819907799.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4819951542.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4820009239.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1265,6 +1265,9 @@ void BenMenu::AddEnhancements() {
     AddWidget(path, "Skip Song of Time cutscenes", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.Songs.SkipSoTCutscenes")
         .Options(CheckboxOptions().Tooltip("Skips the cutscenes when playing any of the Song of Time songs."));
+    AddWidget(path, "Skip Soaring cutscene", WIDGET_CVAR_CHECKBOX)
+        .CVar("gEnhancements.Songs.SkipSoaringCutscene")
+        .Options(CheckboxOptions().Tooltip("Skips the cutscene when using the Song of Soaring to warp."));
 
     // Time Savers
     path = { "Enhancements", "Time Savers", SECTION_COLUMN_1 };
--- a/mm/2s2h/Enhancements/Songs/PauseOwlWarp.cpp
+++ b/mm/2s2h/Enhancements/Songs/PauseOwlWarp.cpp
@@ -1,6 +1,7 @@
 #include <libultraship/bridge/consolevariablebridge.h>
 #include "2s2h/GameInteractor/GameInteractor.h"
 #include "2s2h/ShipInit.hpp"
+#include "2s2h/ShipUtils.h"
 
 extern "C" {
 #include "overlays/kaleido_scope/ovl_kaleido_scope/z_kaleido_scope.h"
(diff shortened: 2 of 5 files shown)


Caladius:
iirc work on Skipping this had been started when Song skips were being implemented but this specific scene was excluded as there was too much going on at the time with Rando to stop and investigate it.

Nice to see it finally finished. Thanks!

mm/2s2h/BenGui/BenMenu.cpp
```diff
@@ -1265,6 +1265,9 @@ void BenMenu::AddEnhancements() {
     AddWidget(path, "Skip Song of Time cutscenes", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.Songs.SkipSoTCutscenes")
         .Options(CheckboxOptions().Tooltip("Skips the cutscenes when playing any of the Song of Time songs."));
+    AddWidget(path, "Skip Soaring cutscene", WIDGET_CVAR_CHECKBOX)
```
Eblo, line 1268: For consistency's sake
```suggestion
    AddWidget(path, "Skip Song of Soaring cutscene", WIDGET_CVAR_CHECKBOX)
```

mm/2s2h/Enhancements/Songs/PauseOwlWarp.cpp
```diff
@@ -18,9 +18,10 @@ extern s32 gHorseIsMounted;
 
 /*
  * This became a static var in decomp, so we cannot extern it. It is already redundantly defined in z_sram_NES.c and
- * z_en_test7.c, so let's just define it a third time for now instead of modifying source.
+ * z_en_test7.c, so let's just define it a third time for now instead of modifying source. At least we can extern it
+ * here for other hooks to use.
  */
```
Eblo: With the change to rename this, the comment that was there is redundant. This might be better moved to ShipUtils, as now it is something used by multiple different places in enhancement code.

```suggestion
/*
 * This is identical to `sOwlWarpEntrances` in decomp code, which is a static variable defined in multiple places and
 * cannot be externed. We redundantly define it here and extern it, for use in port enhancements.
 */
```

mm/2s2h/Enhancements/Songs/SkipSoaringCutscene.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+static void SkipSoaringCutscene(Actor* actor, bool* should) {
+    s16 ocarinaMode = OWL_WARP_CS_GET_OCARINA_MODE(actor);
+    if (ocarinaMode == ENTEST7_ARRIVE) {
+        return;
+    }
+
+    *should = false;
+
+    if (gPlayState->sceneId == SCENE_SECOM) {
```
Eblo, line 22: Why this edge case? Does `OCARINA_MODE_WARP_TO_ENTRANCE` not cover this?

Eblo (changes requested):
Works great! Have some questions and minor feedback.

mm/2s2h/BenGui/BenMenu.cpp
JordanLongstaff: I want it to be clear that it's the actual soaring that gets skipped, not the cutscene (if you can call it that) where you learn the song.

mm/2s2h/Enhancements/Songs/SkipSoaringCutscene.cpp
JordanLongstaff: That's just copy-pasted from the vanilla code in `z_en_test7.c`. Don't ask me why that edge case exists, because I don't know.

JordanLongstaff: Oh wait, that scene is Sakon's Hideout. It must warp you back outside to the canyon.

Eblo (approved):

PR #1384, opened 2025-12-04
Skip Hungry Goron's forced dialogue

### LINKED ISSUE #1379: Cannot skip Hungry Goron's forced dialogue
If you walk past the Hungry Goron in Mountain Village, the game forces you to stop and talk to him, even if all skips are turned on. There is currently no option to skip his forced dialogue like there is in SoH.

Rather than treating this as a "Misc Interaction", I think this requires a new settings to skip forced bits of dialogue.

JordanLongstaff:
Closes #1379 

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4766204457.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4766384883.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4766415766.zip)


Eblo (approved):
LGTM. I think the reason this slipped through the cracks is because rando always skips this by setting the same flag on init, so it's only ever a problem in vanilla.

PR #1390, opened 2025-12-09
Organize `GIVanillaBehavior` enum

### LINKED ISSUE #1385: Organize VB enum
The VB enum desperately needs sorting and documentation a la SoH.

JordanLongstaff:
Closes #1385

Moved the enum to its own header file, sorted its members lexicographically, and ported over the documentation syntax from SoH. Also fixed some spelling mistakes.

Note: a lot of these hooks don't use all of their arguments, and a few of them are missing either any registered hooks, callers, or both.

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4843914079.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4843988349.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4844658716.zip)


mm/2s2h/GameInteractor/GameInteractor_VanillaBehavior.h
```diff
@@ -0,0 +1,1953 @@
...
+    // - `s32` (bgId, unused)
+    VB_BE_HOOKSHOT_SURFACE,
+
+    // #### `result`
+    // ```c
+    // fabsf(playerPosRelToDoor.z) < 50.0f
+    // ```
+    // #### `args`
+    // - `*f32`
+    VB_BE_NEAR_DOOR,
```
Eblo, line 145: While what it's doing is basically the same between the two cases, the variables are just a teensy bit different (straight f32 vs resolving position.z):

```suggestion
    // #### `result`
    // #### In `EnDoor_Idle`, for `*EnDoor`:
    // ```c
    // fabsf(playerPosRelToDoor.z) < 50.0f
    // ```
    // #### `args`
    // - `*f32`
    // #### In `func_808A0E28`, for `*DoorShutter`:
    // ```c
    // fabsf(temp_f0) < 50.0f
    // ```
    // #### `args`
    // - `*f32`
    VB_BE_NEAR_DOOR,
```

Eblo (changes requested):
Checks out to me. I caught one VB that has different args between its multiple uses, and there may be others I missed, but overall this is some much appreciated organization.

mm/2s2h/GameInteractor/GameInteractor_VanillaBehavior.h
JordanLongstaff: Good catch.

Eblo (approved):

Eblo:
The recent merges understandably threw conflicts at this. Resolving those may entail documenting the new VBs as well. If we can get those taken care of, I'd like this to get in before anything else.

JordanLongstaff:
Yup, I knew there'd be conflicts. Thanks for the merges. 👍

PR #1393, opened 2025-12-12
Add build folder to `.gitignore`

### LINKED ISSUE #1392: `.gitignore`: Add `out/build`
Whenever I build the 2SH solution with Visual Studio, it outputs a bunch of files in the `out/build/x64-Debug` directory, which show up as unstaged, added files in Git. That doesn't happen with SoH; I believe it does output those files, but SoH's `.gitignore` is properly configured to exclude them. I can't tell what the difference is specifically that would rectify 2SH's added files.

JordanLongstaff:
Closes #1392

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4846358358.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4846412239.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4846463746.zip)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,5 @@ _packages/
 /mm/src/boot/build.c
 /mm/windows/properties.h
 /clang-format.exe
+
+build/


.gitignore
garrettjoecox, line 61: Is this not satisfied by this line? I’ve never had issues with the build folder not being ignored

JordanLongstaff: Clearly not, because I am having that exact issue.

Malkierian: Also not having this issue. Maybe it's a git version difference?

JordanLongstaff: I don't have this issue with SoH, but I do have it here, so I'd say it has to be an issue with this repo.

But if it helps, my Git version is `git version 2.52.0.windows.1`.

Eblo: I'm not getting this issue on my end as I don't build with VS, but I did a quick and dirty set difference between the two projects' .gitignore lists. SoH has `build/` while 2ship does not. I wonder if `/build*` is only resolving root folders on some setups while `build/` resolves subfolders.

I don't see a problem in adding it.

Eblo (approved):

PR #1396, opened 2025-12-14
Fix cow softlock when cows are not shuffled

Eblo:
Fixes a softlock when playing Epona's Song for cows in a rando seed where cows are *not* shuffled

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4866666987.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4866719277.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4866902903.zip)

### DIFF
--- a/mm/2s2h/Rando/ActorBehavior/EnCow.cpp
+++ b/mm/2s2h/Rando/ActorBehavior/EnCow.cpp
@@ -66,13 +66,15 @@ void Rando::ActorBehavior::InitEnCowBehavior() {
             *should = false;
             return;
         }
+
+        ((EnCow*)actor)->flags |= EN_COW_FLAG_WONT_GIVE_MILK;
+
         RandoCheckId randoCheckId = IdentifyCow(actor);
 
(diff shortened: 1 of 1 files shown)


mm/2s2h/Rando/ActorBehavior/EnCow.cpp
garrettjoecox: Should we not just replace the IS_RANDO here?

Eblo: With the RO for shuffling cows? That would work. Should we go ahead and change all the VBs in this file to follow that pattern?

garrettjoecox: yea, though if they were to disable this specific check from being shuffled in the future we'll probably want both fixes.

PR #1407, opened 2025-12-22
[Enhancement] Ammo Buyback

### LINKED ISSUE #1400: Feature request: Curiosity Shop buys more items
It's always bothered me that there's no way to sell off things like arrows, bombs, Powder Kegs, Deku Nuts etc. The only things the Curiosity Shop accepts for buyback are bottled items.

mckinlee:
Adds a enhancement, allowing players to sell various ammunition types to the Curiosity Shop man. It enables selling Arrows, Bombs, Bombchus, Deku Sticks, Deku Nuts, Magic Beans, and Powder Kegs.

## Optional Todo
- [x] ~Bypass input textbox when selling powder keg (you can only ever sell 1 at a time).~

Resolves #1400 

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4976746572.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4976768676.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4976942837.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -57,6 +57,12 @@ static const std::vector<const char*> cremiaRewardOptions = {
     "Rupee",   // CREMIA_REWARD_ALWAYS_RUPEE
 };
 
+static const std::vector<const char*> ammoBuybackOptions = {
+    "Vanilla",    // AMMO_BUYBACK_VANILLA
+    "Full Price", // AMMO_BUYBACK_FULL_PRICE
+    "Half Price", // AMMO_BUYBACK_HALF_PRICE
+};
+
 static const std::vector<const char*> gibdoTradeSequenceOptions = {
     "Vanilla",  // GIBDO_TRADE_SEQUENCE_VANILLA
     "MM3D",     // GIBDO_TRADE_SEQUENCE_MM3D
@@ -1020,6 +1026,16 @@ void BenMenu::AddEnhancements() {
                               "-Hug: Get the hugging cutscene\n"
                               "-Rupee: Get the rupee reward")
                      .ComboVec(&cremiaRewardOptions));
+    AddWidget(path, "Ammo Buyback Options", WIDGET_CVAR_COMBOBOX)
+        .CVar("gEnhancements.Items.AmmoBuyback")
+        .Options(ComboboxOptions()
+                     .Tooltip("Choose whether to allow selling ammo items (arrows, bombs, bombchus, deku sticks, deku "
+                              "nuts, magic beans, powder keg) "
+                              "to the Curiosity Shop owner for rupees.\n"
+                              "-Vanilla: Ammo items cannot be sold\n"
+                              "-Full Price: Sell at full value\n"
+                              "-Half Price: Sell at half value (rounded up)")
+                     .ComboVec(&ammoBuybackOptions));
     AddWidget(path, "Accessibility", WIDGET_SEPARATOR_TEXT);
     AddWidget(path, "Disable Screen Flash for Enemy Kills", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.A11y.NoScreenFlashForEnemyKill")
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1029,9 +1029,9 @@ void BenMenu::AddEnhancements() {
     AddWidget(path, "Ammo Buyback Options", WIDGET_CVAR_COMBOBOX)
         .CVar("gEnhancements.Items.AmmoBuyback")
         .Options(ComboboxOptions()
-                     .Tooltip("Choose whether to allow selling ammo items (arrows, bombs, bombchus, deku sticks, deku "
-                              "nuts, magic beans, powder keg) "
-                              "to the Curiosity Shop owner for rupees.\n"
+                     .Tooltip("Choose whether to allow selling ammo items (Arrows, Bombs, Bombchus, Deku Sticks, Deku "
+                              "Nuts, Magic Beans, Powder Keg) "
+                              "to the Curiosity Shop owner for Rupees.\n"
                               "-Vanilla: Ammo items cannot be sold\n"
                               "-Full Price: Sell at full value\n"
                               "-Half Price: Sell at half value (rounded up)")
--- a/mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
+++ b/mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
@@ -63,21 +63,21 @@ static s16 GetPricePerUnit(ItemId itemId) {
 static const char* GetNameForDialogue(ItemId itemId) {
     switch (itemId) {
         case ITEM_BOW:
-            return "arrows";
+            return "%rArrows%w";
         case ITEM_BOMB:
-            return "bombs";
+            return "%rBombs%w";
         case ITEM_BOMBCHU:
-            return "bombchus";
+            return "%rBombchus%w";
         case ITEM_DEKU_STICK:
-            return "deku sticks";
+            return "%rDeku Sticks%w";
         case ITEM_DEKU_NUT:
-            return "deku nuts";
+            return "%rDeku Nuts%w";
         case ITEM_MAGIC_BEANS:
-            return "magic beans";
+            return "%rMagic Beans%w";
         case ITEM_POWDER_KEG:
-            return "powder kegs";
+            return "%rPowder Kegs%w";
         default:
-            return "items";
+            return "%ritems%w";
     }
 }
 
@@ -153,12 +153,22 @@ static void HandleStart(EnFsn* enFsn, PlayState* play) {
     if (GetPricePerUnit(buttonItem) > 0 && IsItemAvailable(buttonItem) && AMMO(buttonItem) > 0) {
         sAmmoSale.actor = enFsn;
         sAmmoSale.itemId = buttonItem;
-        sAmmoSale.quantity = 1;
-        sAmmoSale.price = 0;
-        // Last Rupees selected starts at 10 (1.0 or 01)
-        sAmmoSale.lastRupeesSelected = 10;
-        sAmmoSale.isInputActive = true;
-        play->msgCtx.rupeesSelected = 10;
+        sAmmoSale.lastRupeesSelected = 10; // Default for input mode
+
+        // If player only has 1, skip inputs
+        s16 maxAllowed = MIN(AMMO(buttonItem), MAX_AMMO_SELL_QUANTITY);
+
+        if (maxAllowed == 1) {
+            sAmmoSale.quantity = 1;
+            sAmmoSale.price = GetSalePrice(buttonItem, 1);
+            sAmmoSale.isInputActive = false;
+            enFsn->price = sAmmoSale.price;
+        } else {
+            sAmmoSale.quantity = 1;
+            sAmmoSale.price = 0;
+            sAmmoSale.isInputActive = true;
+            play->msgCtx.rupeesSelected = 10;
+        }
 
         enFsn->actionFunc = EnFsn_MakeOffer;
         player->actor.textId = TEXT_ID_FSN_OFFER;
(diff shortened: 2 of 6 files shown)


mm/2s2h/BenGui/BenMenu.cpp
```diff
@@ -1020,6 +1026,16 @@ void BenMenu::AddEnhancements() {
...
                      .ComboVec(&cremiaRewardOptions));
+    AddWidget(path, "Ammo Buyback Options", WIDGET_CVAR_COMBOBOX)
+        .CVar("gEnhancements.Items.AmmoBuyback")
+        .Options(ComboboxOptions()
+                     .Tooltip("Choose whether to allow selling ammo items (arrows, bombs, bombchus, deku sticks, deku "
+                              "nuts, magic beans, powder keg) "
+                              "to the Curiosity Shop owner for rupees.\n"
+                              "-Vanilla: Ammo items cannot be sold\n"
+                              "-Full Price: Sell at full value\n"
+                              "-Half Price: Sell at half value (rounded up)")
```
Eblo:
```suggestion
                     .Tooltip("Choose whether to allow selling ammo items (Arrows, Bombs, Bombchus, Deku Sticks, Deku "
                              "Nuts, Magic Beans, Powder Keg) "
                              "to the Curiosity Shop owner for Rupees.\n"
                              "-Vanilla: Ammo items cannot be sold\n"
                              "-Full Price: Sell at full value\n"
                              "-Half Price: Sell at half value (rounded up)")
```

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
```diff
@@ -0,0 +1,384 @@
...
+        case ITEM_BOMBCHU:
+            return "bombchus";
+        case ITEM_DEKU_STICK:
+            return "deku sticks";
+        case ITEM_DEKU_NUT:
+            return "deku nuts";
+        case ITEM_MAGIC_BEANS:
+            return "magic beans";
+        case ITEM_POWDER_KEG:
+            return "powder kegs";
```
Eblo: Will need to do another pass on these to match how the game normally stylizes them. At the very least I'm sure that Deku should be capitalized.

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
```diff
@@ -0,0 +1,384 @@
...
+            return 4;
+        case ITEM_DEKU_STICK:
+            return 10;
+        case ITEM_DEKU_NUT:
+            return 3;
+        case ITEM_MAGIC_BEANS:
+            return 10;
+        case ITEM_POWDER_KEG:
+            return 50;
+        default:
```
Eblo, line 58: Non-blocking question, curious what these values are based on.

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
```diff
@@ -0,0 +1,384 @@
...
+    s16 maxAllowed = MIN(AMMO(sAmmoSale.itemId), MAX_AMMO_SELL_QUANTITY);
+
+    // Enforce cursor on visible digits (0=Tens, 1=Units)
+    if (msgCtx->unk120C2 > 1)
+        msgCtx->unk120C2 = 1;
+
+    // Vertical Input (Quantity)
+    bool inputUp = CHECK_BTN_ALL(input->cur.button, BTN_DUP) || input->rel.stick_y >= STICK_THRESHOLD;
+    bool inputDown = CHECK_BTN_ALL(input->cur.button, BTN_DDOWN) || input->rel.stick_y <= -STICK_THRESHOLD;
+    s16 delta = (msgCtx->unk120C2 == 0) ? 10 : 1;
```
Eblo, line 203: What is the rationale behind 10 and 1 here?

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
Eblo: Some of the capitalization may seem weird, but some of these items are always capitalized by style, and shop inventory is capitalized anyway.

```suggestion
        case ITEM_BOW:
            return "Arrows";
        case ITEM_BOMB:
            return "Bombs";
        case ITEM_BOMBCHU:
            return "Bombchus";
        case ITEM_DEKU_STICK:
            return "Deku Sticks";
        case ITEM_DEKU_NUT:
            return "Deku Nuts";
        case ITEM_MAGIC_BEANS:
            return "Magic Beans";
        case ITEM_POWDER_KEG:
            return "Powder Kegs";
```

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
```diff
@@ -0,0 +1,384 @@
...
+        case ITEM_DEKU_STICK:
+            return 10;
+        case ITEM_DEKU_NUT:
+            return 3;
+        case ITEM_MAGIC_BEANS:
+            return 10;
+        case ITEM_POWDER_KEG:
+            return 50;
+        default:
+            return 0;
```
Eblo, line 59: There is a bug where, if the setting is not vanilla, then actual vanilla items will not sell. Normally Link gets an item Rupee instead of a counter (e.g. Huge Rupee for gold dust). I think the simplest fix for this is to just account for normally sellable items with the numerical prices, though that might be weird with half price. The other option is to check if it's a non-ammo item first, and if so, proceed with vanilla behavior.

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
```diff
@@ -0,0 +1,384 @@
...
+    if (sAmmoSale.isInputActive) {
+        // Custom Input Textbox
+        CustomMessage::Entry entry;
+        entry.textboxType = 6;
+        entry.textboxYPos = 1;
+        entry.msg = (char)MESSAGE_QUICKTEXT_ENABLE;
+        entry.msg += std::string("How many ") + GetNameForDialogue(sAmmoSale.itemId) + "? " + (char)MESSAGE_NEWLINE +
+                     (char)MESSAGE_INPUT_BANK + (char)MESSAGE_NEWLINE + "Set the amount with " +
+                     (char)MESSAGE_CONTROL_PAD + " and" + (char)MESSAGE_NEWLINE + "press " + (char)MESSAGE_BTN_A +
+                     " to decide." + (char)MESSAGE_END;
```
Eblo, line 338: Out of scope for this PR I'd say, but it may behoove us to define these char conversions in CustomMessage somewhere so we're not using `\x` all the time.

mm/2s2h/GameInteractor/GameInteractor_VanillaBehavior.h
```diff
@@ -2008,6 +2008,22 @@ typedef enum {
...
+    VB_MSG_LOAD_RUPEES_TEXT,
+
+    // #### `result`
+    // ```c
+    // true
+    // ```
+    // #### `args`
+    // - None
+    VB_MSG_PLAY_INPUT_COUNT_SOUND,
+
```
Eblo: These should be moved to their appropriate alphabetical positions, just before `VB_MSG_SCRIPT_DEL_ITEM`.

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
```diff
@@ -0,0 +1,384 @@
...
+
+    COND_VB_SHOULD(VB_MSG_LOAD_RUPEES_TEXT, CVAR != AMMO_BUYBACK_VANILLA, {
+        if (sAmmoSale.isInputActive)
+            *should = false;
+    });
+
+    COND_VB_SHOULD(VB_MSG_PLAY_INPUT_COUNT_SOUND, CVAR != AMMO_BUYBACK_VANILLA, {
+        if (sAmmoSale.isInputActive)
+            *should = false;
+    });
```
Eblo: This affects everything using numerical input, not just the the new dialog you've added. I don't mind having a setting to disable this sound, but it does not make sense for me to have that bundled in with an ammo buyback feature.

Eblo (changes requested):
Ahoy RPG economy mechanics

> Bypass input textbox when selling powder keg (you can only ever sell 1 at a time).

If you go that route, it might make sense to just blanket apply that to anything if you only have 1.

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
mckinlee: They are what @JordanLongstaff recommended on the linked issue. He may be able to offer more information. I'm unsure.

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
mckinlee: Compared to changing the whole number by a fixed step every time, it changes the behavior so the tens place moves in 10s, ones place moves in 1s. It's to give some quality of life.

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
mckinlee: That was not my intention. I will need to tighten this to only effect this specific dialogue. Imo though, a player would find it odd to hear this sound effect. It is not a rupee counter in this context. My preference would be to keep it with some modifications.

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
mckinlee: Oof, yeh I'll fix that.

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
mckinlee: Yeh that's fine, but I might change them to red as well?

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
JordanLongstaff: Powder Keg can be bought in town for 50 Rupees.
Magic Beans can be bought in two different places for 10 Rupees apiece.
Deku Sticks can be bought one at a time in town for 10 Rupees apiece.
And so on and so forth. The other items tend to be sold in packs at the rates I recommended.
Basically, these are the buyback values at the same price that they can be bought.

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
Eblo: That would make sense.

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
mckinlee: That is a great idea! But do agree its out of scope for this PR. Marking this as resolved, but lets keep the idea in the back of our minds for later.

mm/2s2h/Enhancements/Items/AmmoBuyback.cpp
mckinlee: Leaving this unresolved for further discussion. I think I misunderstood what you meant initially. I believe your concern is that this new VB is generic, encompassing all the standard input calls. Bank, bombers code, etc. 

The enhancement itself safeguards against this by checking sAmmoSale.isInputActive. 

I can change that to only be for the bank input if that is what you wish and rename it something like VB_MSG_BANK_INPUT_SFX. Is this your preference? 

Let me know how you want to proceed.

Eblo: VBs should be named after what the source code does, so since this point of code is for *all* numerical inputs, the current name should stand. With the guard of `sAmmoSale.isInputActive`, it is thus restricted to this use case, so I think that works.

That does give rise to another concern. If the player were to reset while in this state and then go to one of the other numerical inputs, they might be silent. What do you think about adding an `OnActorDestroy` for `ACTOR_EN_FSN` that calls your `Reset`? That might clean up any other undiscovered issues as well.

mckinlee: Agreed.

Eblo (approved):

JordanLongstaff (approved):
I just tested this. You did an amazing job. I can't thank you enough. 😁

PR #1412, opened 2025-12-25
Fierce Deity Flipbook Support

Jameriquiah:
This adds flipbook support to custom Fierce Deity models, otherwise normally the game would just crash upon loading a model with flipbooks set up on FD.

Obviously this has no effect on the vanilla FD Link model so he will still look exactly the same, its just for custom models.

Uses a new array for eyes and mouth, **sFDEyesTextures** and **sFDMouthTextures**.
Texture names for each eye and mouth frame models will need to use are:
gLinkFierceDeityEyesOpenTex
gLinkFierceDeityEyesHalfTex
gLinkFierceDeityEyesClosedTex
gLinkFierceDeityEyesRightTex
gLinkFierceDeityEyesLeftTex
gLinkFierceDeityEyesUpTex
gLinkFierceDeityEyesDownTex
gLinkFierceDeityEyesWincingTex

gLinkFierceDeityMouthClosedTex
gLinkFierceDeityMouthHalfTex
gLinkFierceDeityMouthOpenTex
gLinkFierceDeityMouthSmileTex

Textures also don't need to be resized to 64x64 as they aren't replacing any vanilla assets.

I also added support for Deku and Goron, only added Goron Mouth as the eyes already have flipbooks, just not as many frames. Important to note that Deku and Goron mouth do not animate at all so this literally serves no function pretty much, so I can remove it I don't mind. However with custom animations now being possible, it is possible for mod creators to assign these flipbook frames into their animations if they want to. Deku eyes seem to animate as you would expect.

New arrays for Deku are **sDekuEyesTextures** and **sDekuMouthTextures**.
Texture names are as follows:
gLinkDekuEyesOpenTex
gLinkDekuEyesHalfTex
gLinkDekuEyesClosedTex
gLinkDekuEyesRightTex
gLinkDekuEyesLeftTex
gLinkDekuEyesUpTex
gLinkDekuEyesDownTex
gLinkDekuEyesWincingTex

gLinkDekuMouthClosedTex
gLinkDekuMouthHalfTex
gLinkDekuMouthOpenTex
gLinkDekuMouthSmileTex

For Goron **sGoronMouthTextures**.
Textures are:
gLinkGoronMouthClosedTex
gLinkGoronMouthHalfTex
gLinkGoronMouthOpenTex
gLinkGoronMouthSmileTex

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4984491686.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4984544983.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/4984590601.zip)


mm/src/code/z_player_lib.c
```diff
@@ -2057,6 +2058,7 @@ void Player_DrawImpl(PlayState* play, void** skeleton, Vec3s* jointTable, s32 dL
 
     // Only Human, Zora, and Goron will read the eye textures in the head limb display list.
     // Fierce Deity and Deku will point this segment to garbage data, but it will be unread from.
+    PlayerFDFlipbooks_Patch();
```
Eblo: I don't think this patch method for modding belongs in the source code. Even for patches that fix how vanilla things are supposed to look, the code lives solely in enhancement and port files.

I think what you should do is move the `PlayerFDFlipbooks` files into the `GfxPatcher` folder or some new folder that makes sense for mod support. Then call `PlayerFDFlipbooks_Patch` in `BenPort.cpp`, probably right after the call for `GfxPatcher_ApplyNecessaryAuthenticPatches()`.

mm/2s2h/Enhancements/PlayerFDFlipbooks.cpp
```diff
@@ -0,0 +1,63 @@
...
+
+    for (s32 i = 0; i < PLAYER_EYES_MAX; i++) {
+        sPlayerEyesTextures[PLAYER_FORM_FIERCE_DEITY][i] = (TexturePtr)sFDEyesTextures[i];
+    }
+
+    for (s32 i = 0; i < PLAYER_MOUTH_MAX; i++) {
+        sPlayerMouthTextures[PLAYER_FORM_FIERCE_DEITY][i] = (TexturePtr)sFDMouthTextures[i];
+    }
+
+    sFDFacePatchState = 2;
```
Eblo: I can tell 0 means "not patched", but what's the significance between 1 or 2?

mm/2s2h/Enhancements/PlayerFDFlipbooks.h
```diff
@@ -0,0 +1,13 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+void PlayerFDFlipbooks_Patch(void);
+
+#ifdef __cplusplus
+}
+#endif
```
Eblo: With the suggested change to move the call into `BenPort.cpp`, the C++ conditions here can just be removed.

mm/2s2h/Enhancements/GfxPatcher/PlayerFDFlipbooks.cpp
```diff
@@ -0,0 +1,63 @@
+#include "PlayerFDFlipbooks.h"
+#include "z64player.h"
+#include "2s2h/BenPort.h"
+
+extern "C" {
+extern TexturePtr sPlayerEyesTextures[PLAYER_FORM_MAX][PLAYER_EYES_MAX];
+extern TexturePtr sPlayerMouthTextures[PLAYER_FORM_MAX][PLAYER_MOUTH_MAX];
+uint8_t ResourceMgr_FileExists(const char* resName);
```
Eblo: Is this line needed with the include for `BenPort.h`?

Eblo (changes requested):
I'm on board with this, but I think the source footprint is unneeded. Do you have a custom FD mod that you can share so that I can test this?

mm/2s2h/Enhancements/PlayerFDFlipbooks.cpp
Jameriquiah: patch state 1 if for when no flipbooks are set, tells the game not to try again. patch state 2 is when it does detect flipbooks and also... actually does the same thing lol i probably dont need both in retrospect

mm/src/code/z_player_lib.c
Jameriquiah: ah ok interesting, i wouldnt have touched that file at all i just couldnt figure out how to get it to work any other way lol. i can try this suggestion tho

mm/2s2h/Enhancements/GfxPatcher/PlayerFDFlipbooks.cpp
Jameriquiah: are u talking about just the `extern "C" {` or all 4 of those lines? probably not. i forget but likely a remnant from earlier on when i was trying stuff.

Eblo: Just the last line. `ResourceMgr_FileExists` is defined in `BenPort.h`, so including that header, which you've done, should remove the need to declare the method here.

Jameriquiah: ok just took a look. `ResourceMgr_FileExists` is in `BenPort.h` but it doesnt get used from there in `PlayerFDFlipbooks.cpp` bc of the `#ifndef __cplusplus` further up in `BenPort.h`. defined it again here since i wanted to touch as little files as possible

garrettjoecox (commented):
While we're here, it seems reasonable to just add support for deku & goron no? Goron is a bit tricky though with it's abuse of PLAYER_EYES_WINCING

Jameriquiah:
> While we're here, it seems reasonable to just add support for deku & goron no? Goron is a bit tricky though with it's abuse of PLAYER_EYES_WINCING

i thought about it but it since flipbooks are tied to animations, i dont think they would really use them, maybe it would make deku blink or something idk lol. goron also prolly wouldnt use the mouth animations at all. could still be useful for custom animations tho since u could change the flipbook data to whatever u want

Jameriquiah:
implemented suggestions. `z_player_lib.c` is no longer touched

mm/src/code/z_player_lib.c
Jameriquiah: calling it after `GfxPatcher_ApplyNecessaryAuthenticPatches()` was too late, so the model loaded in before the code could be ran which crashed, moved it so its called sooner

mm/2s2h/Enhancements/GfxPatcher/PlayerFDFlipbooks.cpp
```diff
@@ -0,0 +1,63 @@
+#include "PlayerFDFlipbooks.h"
+#include "z64player.h"
```
Eblo: The `z64player.h` header should be moved down into the `extern "C"` block below, as it's a C header.

Eblo (changes requested):

Jameriquiah:
Updated it to include more forms, can remove Deku and Goron mouth if y'all would like me to.
Also fixed a critical oversight where the patch fails if it only detects one flipbook and not the other, which caused the game to crash, tweaked the code to check for eyes and mouth independently instead of together. This will allow mod creators to use either eyes or mouth and not the other if they choose.

Eblo (approved):

PR #1427, opened 2026-01-01
[Rando] Added hidden owl to logic + item tracker

pocable:
Adds the hidden owl to the item tracker, and adjusts the logic to account for the hidden owl if the user gets it. Also fixes a small logic issue with tingle in north clock town as you can use a sword to jump slash the balloon down.

I didn't add the hidden owl as a check since I didn't think that made sense. Moreso added it into logic so I can see what checks I can do when I have it.

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5004265806.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5004292161.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5004521836.zip)


Eblo:
The main obstacle to this getting in is that this pertains entirely to a glitch, but randomizer does not have glitches in logic at the moment. As-is, this can result in "glitchless" seeds that require index warping to beat. There's also a small bit of allowing for sword slashing Clock Town Tingle's balloon, which seems unrelated to the main intent here and is a trick of its own.

We can circle back to this once 2ship has some system for glitch/trick logic in place. Index warping is a prevalent glitch in MM, so I think this PR is worth considering once we actually have a way for players to choose whether they want to incorporate such a glitch into their logic.

pocable:
I never noticed that the logic in the tracker was glitchless. Well that clarifies a lot for me haha.

garrettjoecox:
Going to close this for now until we get around to adding tricks and glitches to the randomizer logic. This should be a good starting point when we get to that point thank you!

PR #1441, opened 2026-01-06
Refactor InitializeFileClocks

mckinlee:
#1431 introduced a pattern of decoupling logic from the global `gSaveContext` by passing `RandoSaveInfo&` to functions like `GeneratePools`. This allows the logic to be used for both actual save file creation and "dry run" metric calculations in the menu.

This refactor applies that same pattern to enable `RefreshMetrics` functionality for Shuffle Time.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5050796398.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5050905309.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5051010494.zip)

### DIFF
--- a/mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
+++ b/mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
@@ -3,7 +3,6 @@
 #include "Rando/Logic/Logic.h"
 #include "2s2h/GameInteractor/GameInteractor.h"
 #include "2s2h/CustomMessage/CustomMessage.h"
-#include <libultraship/bridge/consolevariablebridge.h>
 #include "2s2h/ShipUtils.h"
 
 extern "C" {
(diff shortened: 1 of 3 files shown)


mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
```diff
@@ -122,6 +121,16 @@ namespace ClockShuffle {
...
 // ============================================================================
 
+// Set a rando inf flag on the provided save info
+static void SetRandoInfFlag(RandoSaveInfo& saveInfo, RandoInf flag) {
+    if (&saveInfo == &gSaveContext.save.shipSaveInfo.rando) {
+        Flags_SetRandoInf(flag);
+        return;
+    }
+
+    saveInfo.randoInf[flag >> 4] |= static_cast<u16>(1 << (flag & 0xF));
```
Eblo: Could you leave a comment explaining what these two different cases are handling? Also, instead of an `if` with an early return, I'd prefer an `else` here.

Eblo (approved):

PR #1447, opened 2026-01-09
[Difficulty Option] Shuffle Town Archery Octorok Position

mckinlee:
Wave order is randomized. Each new wave of octoroks will have their positions randomized. Logic in place to prevent blue Octorok's from blocking red ones in center column. 

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5101401012.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5101515582.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5101607610.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1582,6 +1582,10 @@ void BenMenu::AddEnhancements() {
                      .Min(1)
                      .Max(50)
                      .DefaultValue(50));
+    AddWidget(path, "Randomize Shooting Gallery Octoroks", WIDGET_CVAR_CHECKBOX)
+        .CVar("gEnhancements.Minigames.RandomizeShootingGalleryOctoroks")
+        .Options(CheckboxOptions().Tooltip("Randomizes the positions of Octoroks in the Town Shooting Gallery minigame "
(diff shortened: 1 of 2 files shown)


sitton76:
This sounds evil...I love it.

Eblo (commented):
What guarantees that a blue Octorok will not block a red one? That never happened in my testing, which is good, but I cannot tell if that is guaranteed. Similarly, what guarantees that the shuffle is beatable?

mckinlee:
> What guarantees that a blue Octorok will not block a red one? That never happened in my testing, which is good, but I cannot tell if that is guaranteed. Similarly, what guarantees that the shuffle is beatable?

Good point. There likely are no guarantees. We  both just got lucky when testing. I'll revisit this after while and see if I can't come up with a solution.

mckinlee:
Logic has been added to prevent a blue octo from being placed in front of a red one for the center column. I decided to additionally shuffle the wave order to spice things up further. You still have the same score point potential as vanilla so this should be guaranteed beatable every time now.

Eblo (approved):

PR #1448, opened 2026-01-10
Tweaks to how starting items works and is stored and interacted with

garrettjoecox:
- Starting items is now stored differently across the 3 places we need it (save, spoiler, and config)
  - For config & spoiler, we are storing it as a dynamically sized string of arrays `["RI_OCARINA"...]`
  - For save we are storing it as a static sized array (256) of u16's (because we are in C land here)
- Prevent items from showing on the starting items view if they aren't in the pool (like boss souls)
- Prevent items from being added more than once or their max times based on progression (eg 3 for bow)
- Fixed sorting of starting items sections (using map instead of unordered map)
- Unrelated tweaks
  - Fix default value on boot to warp point
  - Fix height of hints tab

We are not attempting to migrate anything here, peoples existing starting item configurations will be lost and will be defaulted to the original starting set

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5086912551.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5086939597.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5087051194.zip)

### DIFF
--- a/mm/2s2h/DeveloperTools/WarpPoint.cpp
+++ b/mm/2s2h/DeveloperTools/WarpPoint.cpp
@@ -113,7 +113,6 @@ void RenderWarpPointSection() {
         "Boot to Warp Point on Launch", WARP_POINT_CVAR "BootToWarpPoint",
         UIWidgets::CheckboxOptions({ { .disabled = skipToFileSelect,
                                        .disabledTooltip = "Incompatible with Skip to File Select enhancement" } })
-            .DefaultValue(true)
             .Color(THEME_COLOR)
             .Tooltip(
                 "If enabled, the game will boot directly to the saved warp point with the debug save when launching "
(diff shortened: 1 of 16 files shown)


mm/2s2h/DeveloperTools/WarpPoint.cpp
```diff
@@ -113,7 +113,6 @@ void RenderWarpPointSection() {
         "Boot to Warp Point on Launch", WARP_POINT_CVAR "BootToWarpPoint",
         UIWidgets::CheckboxOptions({ { .disabled = skipToFileSelect,
                                        .disabledTooltip = "Incompatible with Skip to File Select enhancement" } })
-            .DefaultValue(true)
```
Eblo, line 116: Did you mean to include this change? It doesn't seem relevant to the PR as a whole.

Eblo (commented):
LGTM, just one question

mm/2s2h/DeveloperTools/WarpPoint.cpp
garrettjoecox: mentioned it in the PR description

> Unrelated tweaks
> - Fix default value on boot to warp point
> - Fix height of hints tab

Eblo (approved):

PR #1462, opened 2026-01-16
[WIP] Fix/Document compatibility with 0th/4th day glitches

garrettjoecox:
My main goal here is to get the 0th/4th day glitches working. To pull off 0th day you have to [store song of time](https://www.zeldaspeedruns.com/mm/tech/song-storage), and there are two caveats to know about on 2Ship
- `Faster Song Playback` must be off, as it skips the playback you need to void out during.
- Methods 1 & 2 of [ocarina dive](https://www.zeldaspeedruns.com/mm/tech/ocarina-dive) do not work. We are investigating why and would like to fix them but for now you'll have to use methods 3 & 4.

Summary of code changes:
- Remove enhancement toggle for save delay. Save delay will always be disabled unless we determine the player is trying to pull off the 0th day glitch
- Fix a few places where we were setting isOwlSave to false blindly which breaks 0th day
- Fix UB inconsistency with hardware, ensuring saving at the the Unused Owl Statue (in west clock town) will load you into the Mayor's Residence.
- Fix UB inconsistency with hardware for SceneTitleCard on 0th/4th days, and add our own messages for them (so it will render "Dawn of the Zeroth Day"

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5235610378.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5235721577.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5237301596.zip)


mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
```diff
@@ -246,6 +254,7 @@ void RegisterSavingEnhancements() {
     });
 
     // Owl statue prompt
+    // What statue prompt? Is this not the same as BeforeEndOfCycleSave?
```
garrettjoecox: cc @Eblo you may be able to answer this?

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
```diff
@@ -257,50 +266,82 @@ void RegisterSavingEnhancements() {
...
     });
 
-    GameInteractor::Instance->RegisterGameHook<GameInteractor::BeforeMoonCrashSaveReset>([]() { DeleteOwlSave(); });
-
-    GameInteractor::Instance->RegisterGameHook<GameInteractor::OnSaveLoad>(loadRespawnData);
-}
+    // This should probably be conditional?
+    COND_HOOK(BeforeMoonCrashSaveReset, true, []() { DeleteOwlSave(); });
+
+    // This should probably be conditional?
```
garrettjoecox: cc @Eblo should this be conditional?

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
```diff
@@ -121,7 +122,14 @@ void DeleteOwlSave() {
 
     // Set it to not be an owl save so after reloading the save file it doesn't try to load at the owl's position in
     // clock town
-    gSaveContext.save.isOwlSave = false;
+
+    // gSaveContext.save.isOwlSave = false;
```
garrettjoecox: This line here is the main incompatibility with the 0th day glitch. `DeleteOwlSave` is called from the `BeforeEndOfCycleSave` which is always registered.

I'm not sure that hook _should_ always be registered, but even if we conditionally register it, I think that would make 0th day incompatible with the persistent owl saves enhancement maybe? Maybe that's fine? Idk there's a lot of moving pieces here, considering pause saves are "always persistent"

Simply commenting it out then causes an issue I documented below.

Eblo: > I'm not sure that hook *should* always be registered

I was just looking at that when implementing Moon Crash save. No, it should not be.

> but even if we conditionally register it, I think that would make 0th day incompatible with the persistent owl saves enhancement maybe? Maybe that's fine

I think it's understandable and acceptable that a port enhancement steps on a glitch. We could mitigate this by adding disclaimers to the tooltips, e.g. "Note: Disabling this will prevent the ability to perform certain glitches."

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
Eblo: Any owl statue, asking the player if they want to save their game. It is not the same as BeforeEndOfCycleSave because it is not an end of cycle save. The text hook was just a convenient and non-intrusive spot to advance play time before owl saving.

mm/src/code/z_sram_NES.c
```diff
@@ -2119,10 +2124,11 @@ void Sram_UpdateWriteToFlashOwlSave(SramContext* sramCtx) {
                 sramCtx->status = 4;
             }
         }
-    } else if (OSTIME_TO_TIMER(osGetTime() - sramCtx->startWriteOsTime) >=
-               SECONDS_TO_TIMER(CVarGetInteger("gEnhancements.Saving.DisableSaveDelay", 0) ? 0 : 2)) {
-        // 2S2H [Port] Some tricks require a save delay so we can't just force it to zero
-        // Finished status is hardcoded to 2 seconds instead of when the task finishes
+        // 2S2H [Port] An arbitrary 2 second delay here, we have removed this above in Sram_UpdateWriteToFlashDefault
+        // but we are leaving it here for Owl saves because it's necessary to pull of certain 0th Day glitches
```
Eblo:
```suggestion
        // but we are leaving it here for Owl saves because it's necessary to pull off certain 0th Day glitches
```

Eblo: Just so I'm clear, is the result of this that 2ship saves will never have the save delay, except Owl statue saves, which will now always have it?

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
Eblo: `BeforeMoonCrashSaveReset` was unconditional before I touched anything. Based on the description in `DeleteOwlSave`, it seems like this could be made conditional for persistent owl saves.

I believe I made `OnSaveLoad` unconditional to account for the scenario where the player begins play without "Remember Save Location" and then enables it and pause saves. It should be fine to make it conditional.

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
garrettjoecox: Right, just brain fog moment disregard

mm/2s2h/Enhancements/Saving/SavingEnhancements.cpp
garrettjoecox: I added a comment in code, but adding it here as well, because of the different ways an owl save can be present when it shouldn't be, it's fair game for this hook to always be registered and attempting to delete the _owl_ save here.

```
When resetting the time cycle or letting the moon crash, an owl save would normally not even
be present because they are deleted when loading. However, with persistent owl saves or
pause/auto saves, we need to explicitly delete the owl save here, to let the player load
into the new cycle or lose their progress on moon crash.
```

mm/src/code/z_sram_NES.c
garrettjoecox: It's a bit confusing, I've tweaked the comments and implementation.

Basically the only time we ever want a delay is when they trying to pull off this trick

garrettjoecox:
Alright I'd like some playtesting on this prior to merging, but I think it's in a good state.

Mostly just want heavy testing of existing save functionality to make sure I didn't break anything (Remember save location, auto save, pause saving)

Also feel free to try 0th day stuff. You can easily void out while playing SoT by typing void into developer console while it's playing back.

PR #1465, opened 2026-01-17
[Enhancement] Curiousity Shop Sell Refills

mckinlee:
This allows players to purchase specific refill items (Seahorse, Gold Dust, Chateau Romani) after their primary checks or conditions have been met. 

This is a PoC so please scrutinize it. We also need to discuss if we want to move forward with the three I've chosen or if there are others that would be better.

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5182288282.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5182344008.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5182748317.zip)


mckinlee:
Also need discussion around prices.

garrettjoecox (commented):
I don't really love the idea of mixing rando checks with this "shop item" concept. Is it unreasonable to just handle them entirely separately and copy what we need from the rando code?

mckinlee:
Okay this has been reworked to be a standalone enhancement that can work in both rando and vanilla. Please continue to scrutinize the implementation. Still need to settle on what top 3 highest impact items for the shop to sell and their prices.

garrettjoecox (approved):
Code looks good, would be good to have one person QA validating it works in and out of rando

PR #1478, opened 2026-01-20
Emit notification on seed generation failure

garrettjoecox:
Also adds a spot for running some validation checks, and made shuffling ocarina, SoT, and ocarina buttons incompatible with glitchless for the time being.
Screenshot 2026-01-20 at 9 18 37 AM
Screenshot 2026-01-20 at 9 19 10 AM

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5192845882.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5192940442.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5193039756.zip)

### DIFF
--- a/mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp
+++ b/mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp
@@ -6,6 +6,7 @@
 #include <libultraship/bridge/consolevariablebridge.h>
 #include "ClockShuffle.h"
 #include <spdlog/spdlog.h>
+#include "2s2h/BenGui/Notification.h"
 
 extern "C" {
 #include "functions.h"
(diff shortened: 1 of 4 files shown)


mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp
```diff
@@ -124,15 +125,32 @@ void Rando::MiscBehavior::OnFileCreate(s16 fileNum) {
...
                     }
                 }
 
                 // Grant the starting stuff
                 Rando::GrantStartingItems();
 
+                // Run prelim compatibility/validation checks before attempting to place items
+
+                // Ocarina and Song of Time are currently required for glitchless logic. This
+                // requirement may be lifted once we have an option for not losing progress on moon crash
```
Eblo: Given that #1480 now exists, where do we want to land with this?

Eblo (commented):
Notifications work nicely imo. Now there's the question of how we want to handle ability-to-play-SoT settings with this now that #1480 exists.

mm/2s2h/Rando/MiscBehavior/OnFileCreate.cpp
garrettjoecox: Just removed it entirely. I think in the future we can revisit this decision but for now it's more simple to just force your new option on for glitchless and we can hear how many people complain.

Issue #1474, opened 2026-01-20
Soft lock when granny forwards to morning you don't have


garrettjoecox:
Reproduction steps:
- Load [this spoiler file](https://github.com/user-attachments/files/24727531/grandma2.json)
- Go to granny
- Listen to long story (forwards to day 2, which you don't have yet)
- answer correctly
- softlock

garrettjoecox:
Fixed with #1479

PR #1479, opened 2026-01-20
[Fix] Shuffle Time Granny Softlock

mckinlee:
Addresses #1474 

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5193669917.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5193793425.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5195174168.zip)

### DIFF
--- a/mm/src/overlays/actors/ovl_En_Nb/z_en_nb.c
+++ b/mm/src/overlays/actors/ovl_En_Nb/z_en_nb.c
@@ -7,6 +7,8 @@
 #include "z_en_nb.h"
 #include "objects/object_nb/object_nb.h"
 
+#include "GameInteractor/GameInteractor.h"
+
 #define FLAGS                                                                                  \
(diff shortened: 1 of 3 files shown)


mm/src/overlays/actors/ovl_En_Nb/z_en_nb.c
```diff
@@ -497,7 +499,9 @@ s32 func_80BC01DC(Actor* thisx, PlayState* play) {
         case ENNB_BEHAVIOUR_5:
             if (!CHECK_EVENTINF(EVENTINF_42)) {
                 gSaveContext.save.time = CLOCK_TIME(8, 0);
```
Eblo: It doesn't practically matter since the enhancement code will set time after this, but maybe this should be moved to within the conditional just so it's clear that this code will be ignored in Time Shuffle.

PR #1480, opened 2026-01-20
[Enhancement] Save Game on Moon Crash

Eblo:
With this enabled, game data will be SoT-saved on moon crash instead of having owl data wiped. In this first iteration, this will always be enabled for glitchless seeds since it will help ensure beatability for in seeds where SoT, Ocarina, and/or Ocarina buttons are shuffled. It can be toggled at will for vanilla or other logic settings.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5196079496.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5196165207.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5196273927.zip)


garrettjoecox:
Works as documented, tested glitchless, no logic (setting off & on), vanilla (setting off & on)

PR #1485, opened 2026-01-21
add zora boomerang trail color to cosmetics editor

Jameriquiah:
with custom models being a thing now i thought itd be cool to be able to change the trail color for these. i dont like yellow

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5202980390.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5203082428.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5203167504.zip)


garrettjoecox:
For blurs specifically, it might be better to do this within EffectBlure, then we can handle each of the different blurs like sword trails, chus and boomerangs

Jameriquiah:
> For blurs specifically, it might be better to do this within EffectBlure, then we can handle each of the different blurs like sword trails, chus and boomerangs

i actually tried this but i couldnt figure out how to get the trails to work seperately

Jameriquiah:
> For blurs specifically, it might be better to do this within EffectBlure, then we can handle each of the different blurs like sword trails, chus and boomerangs

ok looking into this now, question tho, where do chu's use trails? my memory may be off but i dont remember

Jameriquiah:
oh u were talking about bombchus... im silly

PR #1504, opened 2026-01-26
Periodic cherry-pick from upstream

Eblo:
DO NOT SQUASH

This brings us up to date with [ff5b972](https://github.com/zeldaret/mm/commit/ff5b97216d5f6f6a43cb0677b2248f6c2bb65c1d).

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9805701410.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9805632260.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9805587181.zip)


mm/assets/objects/object_po_composer/object_po_composer.h
```diff
@@ -69,41 +69,41 @@ static const ALIGN_ASSET(2) char gPoeComposerFlatHeadDL[] = dgPoeComposerFlatHea
 #define dgPoeComposerBurnDL "__OTR__objects/object_po_composer/gPoeComposerBurnDL"
 static const ALIGN_ASSET(2) char gPoeComposerBurnDL[] = dgPoeComposerBurnDL;
 
-#define dobject_po_composer_Tex_007E80 "__OTR__objects/object_po_composer/object_po_composer_Tex_007E80"
-static const ALIGN_ASSET(2) char object_po_composer_Tex_007E80[] = dobject_po_composer_Tex_007E80;
+#define dgPoeComposerArmTex "__OTR__objects/object_po_composer/gPoeComposerArmTex"
+static const ALIGN_ASSET(2) char gPoeComposerArmTex[] = dgPoeComposerArmTex;
```
garrettjoecox, line 73: Will these break existing texture packs?

Eblo:
Updated this to reflect the latest documentation commit. Still need to investigate potential mod breakage.

mm/assets/objects/object_po_composer/object_po_composer.h
Eblo: I can confirm that these break existing texture packs. How should we proceed from here? I did inadvertently break compatibility before with the last decomp cherry-pick. We could maintain compatibility by reverting the relevant renames, though we do lose some documentation that way. Has SoH ever had the same dilemma?

JordanLongstaff: I'm no authority on this subject, but my opinion is that we'll end up renaming this at some point sooner or later anyway, and the problem won't go away. Those texture pack mods will just have to update their paths.

JordanLongstaff:
What needs to happen before this gets merged? Is it just checking mods? Review-wise, it looks fine to me; I skipped reviewing any of the XML files, just taking it for granted that they're correct, but my Python script made reviewing the C files much easier.

garrettjoecox:
> What needs to happen before this gets merged?

Last we chatted, IIRC the goal is to either have an automated process built in or external that can be run to convert mods from an older format to newer, targetting GhostlyDark’s reloaded as a baseline.

Eblo:
On that note, I adapted the SoH equivalent script for 2ship and have been keeping asset renames updated as I added to this PR: https://github.com/Eblo/2S2HTexturePackUpdater

Although, Reloaded in particular just streamlined how it exports for 2ship, requiring only a couple of files to have to be updated to handle asset renames. I have a fork with changes ready to go that I tested for some of the renamed textures, and it seems to check out okay.

With all of that in tow, I am happy to merge this PR if there are no further objections.

PR #1521, opened 2026-01-29
Fix Moon Crash Save Crash

balloondude2:
Moves the BeforeMoonCrashSaveReset hook to an earlier point in the moon crash process, so the save happens while `interfaceCtx` still exists. 

Not sure if this is the best approach. The saving now happens while the HUD is visible, so you can see the inventory changes happen (like the bottle of water changing to an empty bottle). Also not sure if this has any other unforeseen impacts. 

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5310624388.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5310727534.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5311514494.zip)


sitton76:
Cannot reproduce the crash on this, so that is good.

Kinda with you in your trepidation, but tested in a few times(5 or 6 attempts) and could not cause any weird behavior while the HUD shows back up. Still might be good to poke it some more to be sure.

garrettjoecox (approved):
balloon saves the day

PR #1538, opened 2026-02-02
[Enhancement] Require only one bug to spawn a Soft Soil Skulltula

Patrick12115:
Only require one bug, instead of all three, to have to burrow into a soft soil patch to spawn the Golden Skulltula, which matches the behavior in OoT. This allows for quickly rebottling one of the three bugs dropped, so you don't have to keep getting new ones.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5338506287.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5338557746.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5338625426.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1545,6 +1545,11 @@ void BenMenu::AddEnhancements() {
         .CVar("gEnhancements.Restorations.BonkCollision")
         .Options(
             CheckboxOptions().Tooltip("Corrects rolls to allow bonking trees near the end of the roll, as in OoT."));
+    AddWidget(path, "Soft Soil Skulltulas only require one bug", WIDGET_CVAR_CHECKBOX)
+        .CVar("gEnhancements.Restorations.OneBugSoftSoil")
+        .Options(CheckboxOptions().Tooltip(
+            "Soft Soil Golden Skulltulas only require one bug to spawn, matching OoT behavior.\n\n"
+            "This allows rebottling one of the bugs while still spawning the Skulltula."));
     AddWidget(path, "Simulated Input Lag", WIDGET_CVAR_SLIDER_INT)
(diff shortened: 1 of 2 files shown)


Eblo:
I'm not sure how to handle this. Unlike OoT, the bug burrowing in MM is normally a cutscene, rendering this enhancement ineffective on its own. It becomes useful only if One Point Cutscene Skips are enabled, which leads to the scenario where bottling any of the three bugs will prevent the Skulltula spawn.

I'm curious if others have thoughts on how to approach this enhancement.

mckinlee:
> I'm not sure how to handle this. Unlike OoT, the bug burrowing in MM is normally a cutscene, rendering this enhancement ineffective on its own. It becomes useful only if One Point Cutscene Skips are enabled, which leads to the scenario where bottling any of the three bugs will prevent the Skulltula spawn.
> 
> I'm curious if others have thoughts on how to approach this enhancement.

Maybe we could just hook ObjBean's update instead of intercepting the bugs abort function. When `unk_1E4 == 2` and `unk_1E0 >= 1` (one bug has successfully burrowed), force `unk_1E0 = 3`., maybe?

In theory, the Skulltula would spawn early, the cutscene would end naturally after just one bug, and the remaining two bugs are freed back into the world for the player to rebottle. 

Regardless though, I'd prefer this logic to live inside its own enhancement file and use a proper hook.

Xelrog:
Just restore the bottled bug every time the dirt cutscene triggers. There’s no scenario where the person turning this on isn’t using it to rebottle the bugs, as has been said. It’s exactly the same functionally to just keep the bottled bug when you use them on a dirt patch.

Eblo:
Closing in favor of #1756. Thanks for this initial PR.

PR #1541, opened 2026-02-05
Time logic tweaks

garrettjoecox:
null

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5396523362.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5396647592.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5397976535.zip)

### DIFF
--- a/mm/2s2h/Rando/Logic/Logic.cpp
+++ b/mm/2s2h/Rando/Logic/Logic.cpp
@@ -9,7 +9,6 @@ namespace Logic {
 
 std::map<RandoRegionId, RandoRegion> Regions = {};
 
-// Thread-local storage for current region time during check evaluation
 uint64_t gCurrentRegionTime = 0;
 
 RandoRegionId GetRegionIdFromEntrance(s32 entrance) {
--- a/mm/2s2h/Rando/Logic/Logic.cpp
+++ b/mm/2s2h/Rando/Logic/Logic.cpp
@@ -10,7 +10,7 @@ namespace Logic {
 std::map<RandoRegionId, RandoRegion> Regions = {};
 
 // Thread-local storage for current region time during check evaluation
-thread_local uint64_t gCurrentRegionTime = 0;
+uint64_t gCurrentRegionTime = 0;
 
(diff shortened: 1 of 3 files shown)


mm/2s2h/Rando/Logic/Logic.cpp
```diff
@@ -10,7 +10,7 @@ namespace Logic {
 std::map<RandoRegionId, RandoRegion> Regions = {};
 
 // Thread-local storage for current region time during check evaluation
```
Eblo: This comment should probably be changed or removed now that this is no longer thread-local.

mm/2s2h/Rando/Logic/Logic.h
```diff
@@ -101,7 +101,7 @@ struct RegionTimeState {
 };
 
 // Thread-local current region time for check evaluation
```
Eblo: This comment should probably be changed or removed now that this is no longer thread-local.

Eblo (commented):
I think I understand what's been changed here and LGTM, though it wouldn't hurt if this had input from time guru @mckinlee as well.

mckinlee (approved):
Alright so I re-compared this against OOTMM's pathfind.ts and I think we are looking good. MergeTimeStates and TimeStateCovers are basically 1:1 with their mergeAreaData and coveringAreaData. This should fix where multi path time states weren't being merged. If you could reach a region via two different paths at two different times, only the first time was recorded. Milk road being a example. LGTM.

mm/2s2h/Rando/Logic/Logic.cpp
garrettjoecox:
```suggestion
```

mm/2s2h/Rando/Logic/Logic.h
garrettjoecox:
```suggestion
```

PR #1546, opened 2026-02-06
Clean up ClockShuffle.cpp

mckinlee:
Slapping a do not merge on this because it will need thoroughly vetted.

- Added BeforePlayInit hook for time correction before scene initialization
- Consolidated time enforcement into EnforceOwnedTime() which handles all spawn scenarios (file load, scarecrow dance, granny story, etc.)
- Added some methods from the save editor to properly sync things like weather, enemies, etc.
- Fixed transition cutscene detection, should resolve infinite loops.
- Properly sync Interface and Environment state on night transitions
- And did a good spring cleaning of the file

I will continue to test these changes and make adjustments as needed. Just need more eyes on this.

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5643663973.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5643740316.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5645227316.zip)

### DIFF
--- a/mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
+++ b/mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
@@ -250,7 +250,7 @@ static bool CheckAndSkipUnownedTime(Actor* timeActor) {
             enTest4->daytimeIndex = 0;
             gSaveContext.save.day--;
         } else {
-            enTest4->daytimeIndex = 1;
+            enTest4->daytimeIndex = IsNight(time) ? 0 : 1;
             Interface_NewDay(gPlayState, gSaveContext.save.day);
             Environment_NewDay(&gPlayState->envCtx);
 
--- a/mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
+++ b/mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
@@ -360,7 +360,7 @@ static void ProcessClockShuffleMessage(u16* textId, bool* loadFromMessageTable, 
 
     if (isSongOfTime) {
         entry.msg = "Save and return to " + destinationText + "?\n%gYes\nNo\xC2";
-    } else {
+    } else { // Song of Double Time
         entry.msg = "Time moves strangely...\nProceed to " + destinationText + "?\n%gYes\nNo\xC2";
     }
 
--- a/mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
+++ b/mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
@@ -368,7 +368,7 @@ static void ProcessClockShuffleMessage(u16* textId, bool* loadFromMessageTable, 
     *loadFromMessageTable = false;
 }
 
-// Called at BeforePlayInit. Validates time at cycle start or when a day telop
+// Called at OnFileLoad and OnPlayDestroy. Validates time at cycle start or when a day telop
 // transition is pending (respawnFlag -4). Jumps to first owned half-day if needed.
 static void EnforceOwnedTime() {
     bool isCycleStart =
@@ -429,7 +429,11 @@ static void EnforceOwnedTime() {
 void Rando::ClockShuffle::OnFileLoad() {
     bool shouldRegister = IS_RANDO && RANDO_SAVE_OPTIONS[RO_CLOCK_SHUFFLE];
 
-    COND_HOOK(BeforePlayInit, shouldRegister, []() { EnforceOwnedTime(); });
+    if (shouldRegister) {
+        EnforceOwnedTime();
+    }
+
+    COND_HOOK(OnPlayDestroy, shouldRegister, []() { EnforceOwnedTime(); });
 
     COND_ID_HOOK(ShouldActorUpdate, ACTOR_EN_TEST4, shouldRegister, [](Actor* actor, bool* should) {
         // Skip time checks if a transition cutscene is in progress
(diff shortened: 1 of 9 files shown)


mckinlee:
Decided to remove the `SceneHasDayOnlyActors()` in favor of methods found in the save editor. The following should now properly sync:

- Weather
- Bell Times
- Scene sequences
- Screen scale
- Clock tower stairs

I think this is good to go, but would like at least 2 approvals before we merge.

mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
```diff
@@ -63,410 +54,267 @@ RandoItemId GetClockItemFromHalfDayIndex(int halfDayIndex) {
...
-    switch (sceneId) {
-        case SCENE_BOWLING: // Honey & Darling - minigame mode changes based on day/time
-            return true;
-        case SCENE_CLOCKTOWER: // South Clock Town - Clock Tower platform appears at midnight Night 3
-            return CURRENT_DAY == 3;
-        default:
-            return false; // Most scenes handle time changes without reload
-    }
+static bool SceneNeedsReloadForTimeSkip(s16 sceneId) {
+    return sceneId == SCENE_BOWLING;
```
Eblo, line 170: Is Honey & Darling now the only scene that needs a reload? I see in practice that South Clock Town's tower reloads correctly without this. Was it unneeded before?

mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
```diff
@@ -63,410 +54,267 @@ RandoItemId GetClockItemFromHalfDayIndex(int halfDayIndex) {
...
+            } else {
+                EnTest4_GetBellTimeAndShrinkScreenBeforeDay3(enTest4, gPlayState);
+            }
+        }
 
-    // Skip if eventInf bit 0x0f is set (dog race)
-    if (gSaveContext.eventInf[0] & 0x0F) {
-        return;
-    }
+        if (gPlayState->sceneSequences.ambienceId != AMBIENCE_ID_13) {
```
Eblo, line 268: What does this represent, going from night to day?

mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
```diff
@@ -481,133 +329,119 @@ std::string GetTimeDescriptionForMessage(s32 day, u16 time) {
...
+            destinationText += "Third";
         }
+        destinationText += " Day%w";
     }
 
-    // Use different message format for each song
     if (isSongOfTime) {
         entry.msg = "Save and return to " + destinationText + "?\n%gYes\nNo\xC2";
-    } else { // Song of Double Time
+    } else {
```
Eblo: I'd prefer this comment remain:
```suggestion
    } else { // Song of Double Time
```

mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
mckinlee: It was needed in the prior version because of the stairs leading to the top of the clock tower not updating on transition to terminal state. Now we manually set actor.draw = ObjTokeiStep_DrawOpen and call ObjTokeiStep_SetupOpen which properly starts the opening sequence just like the Save Editor would do in UpdateGameTime. This negates the need for the forced reload.

mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
mckinlee: Save editor does this as well. I believe it's to prevent the earthquake rumbling and such from suddenly cutting to silence and restarting during the final hours.

mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
```diff
@@ -552,62 +368,80 @@ void ProcessClockShuffleMessage(u16* textId, bool* loadFromMessageTable, bool is
...
+    if (isPendingDayTelop) {
+        CLEAR_EVENTINF(EVENTINF_TRIGGER_DAYTELOP);
+        gSaveContext.respawnFlag = -8;
+    }
+}
+
+void Rando::ClockShuffle::OnFileLoad() {
+    bool shouldRegister = IS_RANDO && RANDO_SAVE_OPTIONS[RO_CLOCK_SHUFFLE];
+
+    COND_HOOK(BeforePlayInit, shouldRegister, []() { EnforceOwnedTime(); });
```
garrettjoecox: Could you try `OnPlayDestroy` here? 

My biggest concern is this looks like someone would accidentally reach for it, and it fires before `gPlayState` is available which pretty much everything everywhere uses

garrettjoecox (approved):
Would be good to get one playtest on this, lots of changes sprinkled throughout but looks good

mm/2s2h/Rando/MiscBehavior/ClockShuffle.cpp
mckinlee: As I mentioned in discord, we might could use OnPlayDestroy here. There are potential regressions it would introduce tho. One thing I know for sure it would break is corrections on save load. Not sure what else. How would you like to proceed? OnPlayDestroy + Readding the initial save load correction? Or leave as is and document that gPlayState can be null when the hook fires?

PR #1550, opened 2026-02-10
[Rando] Revise kill/drop logic for some enemies

Eblo:
Being able to swim does not equal being able to kill this Octorok

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5575902197.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5575998620.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5577061110.zip)


garrettjoecox:
> Being able to swim does not equal being able to kill this Octorok

Can you explain the change? I think as is it's "Can you kill the octorok" + "can you reach the item it dropped" (which seems right I think?)

Eblo:
The previous condition only represented the ability to reach the Octorok, but the check is for the enemy drop. You have to be able to kill the Octorok from a certain range.

- ❌ Human form alone cannot stand within the Octorok's active range and thus cannot damage the Octorok
- ❌ Goron form cannot damage the Octorok from a distance
- ✔ Deku form can get on the flowers and use magic to defeat the Octorok
- ✔ Zora form can walk within the active range and use shield or boomerangs to defeat the Octorok

Xelrog:
The issue is being able to actually go out and collect the item the Octorok drops, isn't it? Can you collect it with Deku bounces? Otherwise you would need the ability to traverse the water or swim, Deku alone wouldn't be enough.

garrettjoecox:
I would say you need `CAN_TRAVERSE_WAIST_DEEP_WATER` to pick up the item, and inaccuracies with the actual killing of the octorok should be handled inside `CanKillEnemy(ACTOR_EN_OKUTA)`

Eblo:
> The issue is being able to actually go out and collect the item the Octorok drops, isn't it?

Both being able to kill the Octorok and collect its drop. Which makes me realize that, indeed, Deku cannot actually pick up the item. Oops.

Eblo:
Did another pass on this to revise kill logic vs can-pick-up-enemy-drop logic. Also fixed some other enemies while I was at it, such as Snappers failing to logically require souls

PR #1558, opened 2026-02-18
Picto-box on C-up

OtherBlue:
Adds option to open picto-box with C-up once acquired

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5635264008.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5635448371.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5636818216.zip)

### DIFF
--- a/mm/2s2h/GameInteractor/GameInteractor_VanillaBehavior.h
+++ b/mm/2s2h/GameInteractor/GameInteractor_VanillaBehavior.h
@@ -2212,6 +2212,14 @@ typedef enum {
     // #### `args`
     // - `*f32` (speed)
     VB_ZTARGET_SPEED_CHECK,
+
+    // #### `result`
+    // ```c
+    // true
+    // ```
+    // #### `args`
+    // - None
+    VB_FIRST_PERSON_CAMERA,
 } GIVanillaBehavior;
 
 #endif
--- a/mm/2s2h/GameInteractor/GameInteractor_VanillaBehavior.h
+++ b/mm/2s2h/GameInteractor/GameInteractor_VanillaBehavior.h
@@ -651,6 +651,14 @@ typedef enum {
     // true
     // ```
     // #### `args`
+    // - None
+    VB_FIRST_PERSON_CAMERA,
+
+    // #### `result`
+    // ```c
+    // true
+    // ```
+    // #### `args`
     // - `*EnFish2`
     VB_FISH2_SPAWN_HEART_PIECE,
 
@@ -2212,14 +2220,6 @@ typedef enum {
     // #### `args`
     // - `*f32` (speed)
     VB_ZTARGET_SPEED_CHECK,
-
-    // #### `result`
-    // ```c
-    // true
-    // ```
-    // #### `args`
-    // - None
-    VB_FIRST_PERSON_CAMERA,
 } GIVanillaBehavior;
 
 #endif
--- a/mm/src/overlays/actors/ovl_player_actor/z_player.c
+++ b/mm/src/overlays/actors/ovl_player_actor/z_player.c
@@ -8229,8 +8229,7 @@ s32 Player_ActionHandler_0(Player* this, PlayState* play) {
         this->stateFlags2 |= PLAYER_STATE2_200000;
     } else if (GameInteractor_Should(VB_FIRST_PERSON_CAMERA, true)) {
         if (this->tatlTextId == 0 && !Player_CheckHostileLockOn(this) &&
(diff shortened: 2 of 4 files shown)


mm/2s2h/GameInteractor/GameInteractor_VanillaBehavior.h
```diff
@@ -2212,6 +2212,14 @@ typedef enum {
...
     // - `*f32` (speed)
     VB_ZTARGET_SPEED_CHECK,
+
+    // #### `result`
+    // ```c
+    // true
+    // ```
+    // #### `args`
+    // - None
+    VB_FIRST_PERSON_CAMERA,
```
Eblo: We're trying to maintain alphabetical order here, so this should go between `VB_FD_ALWAYS_WIELD_SWORD` and `VB_FISH2_SPAWN_HEART_PIECE`.

mm/src/overlays/actors/ovl_player_actor/z_player.c
```diff
@@ -8227,10 +8227,11 @@ s32 Player_ActionHandler_0(Player* this, PlayState* play) {
...
         this->stateFlags2 |= PLAYER_STATE2_200000;
-    } else if ((this->tatlTextId == 0) && !Player_CheckHostileLockOn(this) &&
-               CHECK_BTN_ALL(sPlayerControlInput->press.button, BTN_CUP) &&
-               !func_80831814(this, play, PLAYER_UNKAA5_1)) {
-        Audio_PlaySfx(NA_SE_SY_ERROR);
+    } else if (GameInteractor_Should(VB_FIRST_PERSON_CAMERA, true)) {
+        if (this->tatlTextId == 0 && !Player_CheckHostileLockOn(this) &&
+            CHECK_BTN_ALL(sPlayerControlInput->press.button, BTN_CUP) && !func_80831814(this, play, PLAYER_UNKAA5_1)) {
+            Audio_PlaySfx(NA_SE_SY_ERROR);
+        }
```
Eblo: Instead of nesting the vanilla condition under a new VB condition, have the vanilla condition be the boolean argument for the GameInteractor call (clang-format might complain about this, reformat as needed):
```suggestion
    } else if (GameInteractor_Should(VB_FIRST_PERSON_CAMERA, (this->tatlTextId == 0) && !Player_CheckHostileLockOn(this) && CHECK_BTN_ALL(sPlayerControlInput->press.button, BTN_CUP) &&
               !func_80831814(this, play, PLAYER_UNKAA5_1))) {
        Audio_PlaySfx(NA_SE_SY_ERROR);
```

mckinlee (approved):

Eblo (changes requested):

mm/2s2h/GameInteractor/GameInteractor_VanillaBehavior.h
OtherBlue: can't believe I forgot to change this

mm/src/overlays/actors/ovl_player_actor/z_player.c
OtherBlue: done

Eblo:
The GitHub actions build seems to have an issue with this setting where pressing C-Up with the Pictograph Box acquired will result in a double hit that starts to go into first person but immediately backs out. However, if I pull these changes in and build locally, I do not get that issue. I'm not sure why that is; the code seems correct.

OtherBlue:
weird, just tried it out and I don't get the issue you're saying
only weird thing with my picto box is it auto deactivates at 6:30 and 6:50 of the first day for some reason, but I also tested in release and it happens there

Eblo:
Can confirm there are two issues:

- Auto deactivating at those times (related to Kafei checking mail? Kafei's actor is known to mess with player state)
- The instant deactivate if time speed is 0. This may affect certain areas of the game

OtherBlue:
but none of those are related to this feature, I get them even on Keiichi Bravo
should fixes to those even come in this PR?

Eblo (approved):
After digging further, I think the first issue is actually vanilla behavior, or at least not related to this PR. It can be encountered even without this setting by having the pictograph box active normally at those times.

The second issue seems specific to this, but I don't think the time speed variable can ever normally be 0 aside from the save editor. Even in areas where time does not flow, that variable remains the same and does not present this issue. This seems good to me.

PR #1560, opened 2026-02-21
Replace our Z-Targeting fix with recomp's

garrettjoecox:
I tried many times to understand what this was doing, and port it to a hook approach but ended up with more questions than answers. 

Full credit to danielryb for submitting this to recomp (Co-authored on commit)

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5599832536.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5599866771.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5600197994.zip)

### DIFF
--- a/mm/src/code/z_camera.c
+++ b/mm/src/code/z_camera.c
@@ -2722,6 +2722,13 @@ s32 Camera_Parallel1(Camera* camera) {
     s16 phi_a0;
     s32 phi_a0_2;
 
+    // #region 2S2H [Enhancement] - FixTargettingCameraSnap
+    static bool prevTargetingHeld = false;
+    bool isChargingDekuFlowerDive = !!(((Player*)camera->focalActor)->stateFlags3 & PLAYER_STATE3_100);
+    s16 timer4 = rwData->timer2 >> 5;
+    rwData->timer2 &= 0x001F;
+    // #endregion
+
     if (!RELOAD_PARAMS(camera)) {
     } else {
         CameraModeValue* values = sCameraSettings[camera->setting].cameraModes[camera->mode].values;
@@ -2785,6 +2792,9 @@ s32 Camera_Parallel1(Camera* camera) {
                 rwData->timer2 = 20;
             } else {
                 rwData->timer2 = 6;
+                // #region 2S2H [Enhancement] - FixTargettingCameraSnap
+                timer4 = 6;
+                // #endregion
             }
 
             if ((camera->focalActor == &GET_PLAYER(camera->play)->actor) && (camera->mode == CAM_MODE_CHARGE)) {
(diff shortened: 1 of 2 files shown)


danielryb:
@garrettjoecox Hey, so I noticed this PR because github awarded me an achievement for co-authoring a merged commit. lol

Anyways, the patch consists of 2 layers of vannila fixes and an additional fix for a bug introduced by my changes.

- The first issue, addressed in https://github.com/danielryb/MMRecompCameraFixes/commit/317c54cc0a2a9baa5eef5d40bdf66f76a6324172, is camera not updating the input direction during transition (first 6 frames, counted by timer2) into targeting mode:
https://github.com/zeldaret/mm/blob/0b51a16453b0dcf136bf694206ee603adac67eaf/src/code/z_camera.c#L3035-L3039
https://github.com/zeldaret/mm/blob/0b51a16453b0dcf136bf694206ee603adac67eaf/src/code/z_camera.c#L7548-L7552
Same code in oot: https://github.com/zeldaret/oot/blob/e0e0e93644ca9055a56892a939800c1723c9a1ba/src/code/z_camera.c#L2297-L2299
This causes input direction to be tied to the position of the camera at the beggining of the transition, and causes Link to go in the opposite way if you try to rotate the camera 180* and hold up, or the weird movement direction if the transition occurs when Link does a ledge grab. You can see those cases in [Nerrel's showcase of my mod](https://youtu.be/NtSDwwZqJ5A?si=dtxPMwtnlhygSUBy&t=617). The solution is to simply always update that direction:
https://github.com/danielryb/MMRecompCameraFixes/blob/317c54cc0a2a9baa5eef5d40bdf66f76a6324172/src/camera_patches.c#L403-L404

- The next 2 issues, addressed in https://github.com/danielryb/MMRecompCameraFixes/commit/0d2044a403b4719f046e7903fcd3d2a57024d62b, are [camera following player's rotation after releasing Z-target](https://cdn.discordapp.com/attachments/1269806866674876522/1293313553801154642/fix-z-quickturn.mp4?ex=69a4a59b&is=69a3541b&hm=c5c5dc16877c5d2f420981454d95b0e8a1b8e04af50390029632f6d764f8aa06&) and [camera rotating immediately if you press-release-press Z-target in a small time span](https://cdn.discordapp.com/attachments/1269806866674876522/1293582790713151508/2024-10-09_10-34-13.mp4?ex=69a44eda&is=69a2fd5a&hm=c1148de46567d3f117a670592dbbe86bc19b47eb87a08b192d96d00f34746fc1&). Both of these bugs are present in oot, but they are much harder to perform, since the transition duration is much shorter (or at least timer2 is being set to a smaller value on a transition which, if I remember correctly, is 3 while in mm it's 6).
My solution to the first problem was allowing the desired rotation to be overwritten only on frame where Z-press is detected and disabling overwriting it on any subsequent Z-hold-frames:
https://github.com/danielryb/MMRecompCameraFixes/blob/0d2044a403b4719f046e7903fcd3d2a57024d62b/src/camera_patches.c#L224
https://github.com/danielryb/MMRecompCameraFixes/blob/0d2044a403b4719f046e7903fcd3d2a57024d62b/src/camera_patches.c#L245-L247
The latter problem could be solved by [resetting timer2 ](https://github.com/danielryb/MMRecompCameraFixes/blob/0d2044a403b4719f046e7903fcd3d2a57024d62b/src/camera_patches.c#L223)(which is used during [slerp calculation to the desired rotation](https://github.com/zeldaret/mm/blob/0b51a16453b0dcf136bf694206ee603adac67eaf/src/code/z_camera.c#L2969C94-L2969C100)) on any new Z-presses during the transition. However, timer2 is also used to store the remaining duration of the transition, so I had to separate that part of it's functionality into timer4. It eventually got embedded into unused bits of timer2 because in recomp you can't extend existing structs and Wiseguy didn't like it being a static variable, but these limitations can probably be omitted in 2ship.
- Finally, last commit https://github.com/danielryb/MMRecompCameraFixes/commit/f179d9f5ca742db1e3394b46e59e03bf55c65204 fixes an issue, which was caused by above changes, where diving into a deku flower would result in a small camera wiggle. It basically reverts the behaviour to vanilla if player is diving into a flower.

There are some very minor coding mistakes in commits above (namely '&!' instead of '&& !' in one place and '!!' in a flower-dive-check expression), but I hope the information I provided will clear most of your confusion.

PR #1496, opened 2026-01-24
[Enhancement] Bomb Arrows

mckinlee:
It works, but still needs work. Putting this out here for visibility. Someone slap a do not merge on this please.

## TODO

- [x] ~Draw funcs: make the bomb appear on the arrow actor during use (currently looks like a regular arrow)~
- [ ] Better icon, maybe?
- [x] ~Arrow cycle support~
- [x] ~Unequip item enhancement needs to consider it a bow~
- [ ] (Out of scope?) Possibly shuffle into the rando, no logic likely needed
- [x] ~Add a variant of the magic equip animation in the pause menu~
- [x] ~Prevent the bomb equip white border draw and use the bow’s instead~
- [x] ~Add explosion timer~
- [x] ~equipping regular bombs unequips bomb arrows, needs fixed~
- [ ] More to come?

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6755956765.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6755924416.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6755890486.zip)

PR #1566, opened 2026-02-21
[Enhancement] Song Items

mckinlee:
Allows the player to equip songs as items from the quest status screen of the pause menu to C/D buttons. Song automatically plays on use. All songs can be equipped with the exception of Sun's and Saria's. 

Note:
#1496 introduces a new VB (VB_KALEIDO_DRAW_EQUIP_ANIM_ICON) which is also used here.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5785981608.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5785322892.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5785094515.zip)


mm/2s2h/Enhancements/Songs/SongItems.cpp
```diff
@@ -0,0 +1,408 @@
...
+        Interface_SetHudVisibility(HUD_VISIBILITY_ALL);
+    };
+
+    updateButtonStatus((isHoveringSong && play->msgCtx.msgLength == 0) ? BTN_ENABLED : BTN_DISABLED);
+
+    if (!isHoveringSong) {
+        return;
+    }
+
+    ItemId songItem = (ItemId)(ITEM_WALLET_GIANT + cursorPoint);
```
Eblo: Why is `ITEM_WALLET_GIANT` the base value here?

mm/2s2h/Enhancements/Songs/SongItems.cpp
```diff
@@ -0,0 +1,408 @@
...
+        gDPSetCombineLERP(POLY_OPA_DISP++, PRIMITIVE, ENVIRONMENT, TEXEL0, ENVIRONMENT, TEXEL0, 0, PRIMITIVE, 0,
+                          PRIMITIVE, ENVIRONMENT, TEXEL0, ENVIRONMENT, TEXEL0, 0, PRIMITIVE, 0);
+
+        gDPSetEnvColor(POLY_OPA_DISP++, 60, 70, 40, 0);
+        gDPSetPrimColor(POLY_OPA_DISP++, 0, 0, 255, 255, 255, pauseCtx->alpha);
+
+        auto drawBorder = [&](ItemId item) {
+            if (!IsSongItem(item))
+                return;
+            s32 questIndex = item - ITEM_WALLET_GIANT;
```
Eblo: See other comment re: `ITEM_WALLET_GIANT`

mm/2s2h/Enhancements/Songs/SongItems.cpp
```diff
@@ -0,0 +1,408 @@
...
+        return;
+    }
+
+    s16 cursorPoint = pauseCtx->cursorPoint[PAUSE_QUEST];
+    bool isHoveringSong = false;
+
+    bool isDpadEquipsEnabled = CVarGetInteger("gEnhancements.Dpad.DpadEquips", 0);
+
+    if ((cursorPoint >= QUEST_SONG_SONATA && cursorPoint <= QUEST_SONG_SUN) ||
+        cursorPoint == QUEST_SONG_LULLABY_INTRO) {
```
Eblo, line 219: Because Lullaby Intro is removed from the other contiguous song values, it has to be specially handled as you've done here and in other places. However, I think something is not right, as I can never equip Lullaby Intro, but I can equip the other songs, including the full lullaby.

Eblo (commented):
Some thoughts:

- I cannot actually play any equipped song. They are always disabled even if I have the ocarina
- All of the gray-colored songs are hard to tell apart. There are no other cases of two different items having identical icons. It may be easy to forget what's what, although I don't know how that can be helped
- This may not mesh well with Sun's Song and Saria's Song, which cannot be selected in the menu

Eblo (approved):

PR #1571, opened 2026-02-23
Add slower swim speed for Zora Link

### LINKED ISSUE #1557: Feature Request: Zora slow swimming
The 3DS version makes Zora Link swim slower than in the N64 version. The normal speed is still attainable, but only by magic shielding. Meanwhile, the slower swim speed just doesn't exist in the N64 version. I'd like it to be added as a setting.
The Z button is disabled while swimming, as it turns out, so maybe it could be used, unless someone wants to add an option to enable it. The B button overrides the swim and makes Zora Link descend, which I'd rather it didn't; I'd prefer to come out of swimming myself and then descend.
Whichever button gets used, possible options for this setting could be:
- Fast swim by default, slow by holding the button
- Vice versa
- Button toggles it

JordanLongstaff:
Closes #1557

The idea was to help players not bang their heads so much when trying to swim in confined spaces (e.g. Great Bay Temple), but the game gives you better control than that; you can just tilt the stick and Zora Link will swim in that direction. Still, it might be an appreciated feature.

Anyway, you can now slow down Zora Link's swim speed using the Z button.

zora-slow-swim

Let's compare. Using the side stroke method, Zora Link swims past this platform in a bit over 3 seconds:

https://github.com/user-attachments/assets/0834af79-f93a-4a5d-be2f-30d74916cb09

When dash swimming at his normal speed, he clocks in at around 1.3 seconds:

https://github.com/user-attachments/assets/6c72745c-cde3-4109-bc58-b2f3f6af5d98

And at the reduced swim speed, he's a bit over 2 seconds, right in the middle:

https://github.com/user-attachments/assets/512fe39c-7e6d-4646-a222-d1f9b35fcc65

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5612394749.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5612463145.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5613222341.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -63,6 +63,12 @@ static const std::vector<const char*> ammoBuybackOptions = {
     "Half Price", // AMMO_BUYBACK_HALF_PRICE
 };
 
+static const std::vector<const char*> slowZoraSwimOptions = {
+    "Off",              // SLOW_ZORA_SWIM_OFF
+    "Hold to slow",     // SLOW_ZORA_SWIM_HOLD_SLOW
+    "Hold to speed up", // SLOW_ZORA_SWIM_HOLD_FAST
+};
+
 static const std::vector<const char*> gibdoTradeSequenceOptions = {
     "Vanilla",  // GIBDO_TRADE_SEQUENCE_VANILLA
     "MM3D",     // GIBDO_TRADE_SEQUENCE_MM3D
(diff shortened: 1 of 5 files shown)


Jameriquiah:
as much as i hated slow swimming in mm3d, what about a way to load a different animation while slow swimming? prolly couldnt include it in 2ship.o2r and i dont think theres any vanilla animations that could fit with this but like if a mod can make animations for it they can use it, if its not present fall back to what u have now? prolly of of the scope for this but just throwin ideas

JordanLongstaff:
> what about a way to load a different animation while slow swimming?

There might be a way to add animations, but my experience does not include that.

Eblo (approved):

Glought:
I have a suggestion instead of it being hard coded to the Z button maybe use the `BtnSelector` widget from `UIWidgets.cpp` so the player can change it to a different button like maybe one of the modifier buttons and the Z button can be set as default.

Look at PR #1464 for a example of it in use since `BtnSelector` was added in that PR.

JordanLongstaff:
> I have a suggestion instead of it being hard coded to the Z button maybe use the `BtnSelector` widget from `UIWidgets.cpp` so the player can change it to a different button like maybe one of the modifier buttons and the Z button can be set as default.

I like this idea. I'm going to consider it a blocker.

JordanLongstaff:
I'm going to actually close this and redo it later with the `BtnSelector` suggestion. I just don't want this to be merged as it is.

JordanLongstaff:
> what about a way to load a different animation while slow swimming?

I looked at the animation in MM3D, and it kinda resembles the underwater swimming animation from Twilight Princess/Skyward Sword. Maybe one of them could be imported.

mckinlee:
@JordanLongstaff Slapped a do not merge tag on there. Up to you, but you're fine if you wanna reopen this. We'll wait.

Glought:
You could of marked as draft instead of closing it

PR #1572, opened 2026-02-23
Better owl warp menu

### LINKED ISSUE #1556: Better owl warp navigation
The ordering of owl warps when choosing is unintuitive and slow. You can only choose by tilting the stick left/right in essentially a linked list where the index of each owl's placement in the list does not conform to its place on the map. I want an option to making choosing owl warps more intuitive, using all four directions with the stick and mapping choices intuitively to each direction.

JordanLongstaff:
Closes #1556

Control Stick to move the cursor is made to be more natural in all directions, rather than just left/right to move through a predetermined order that is not intuitive. Made to work no matter how many Owl Statues have been unlocked.

https://github.com/user-attachments/assets/1e8c4935-9519-4553-ba01-5f6f1b2580a8

Made to work with Pause Owl Warp too:

https://github.com/user-attachments/assets/3ba8a2a9-189c-4ef8-b844-87bbffa5731d

https://github.com/user-attachments/assets/0272ab87-e891-459c-b37b-693dd8f197df

### Build Artifacts
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5658644991.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5658769932.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5660244461.zip)


JordanLongstaff:
> Unfortunately, I just discovered that Pause Owl Warp is now broken (regardless of this new setting) if not all statues are unlocked. **DO NOT MERGE UNTIL THIS IS FIXED!!!**

Fixed now.

mm/2s2h/GameInteractor/GameInteractor_VanillaBehavior.h
```diff
@@ -1460,6 +1460,14 @@ typedef enum {
     // - `*EnOwl`
     VB_OWL_TELL_ABOUT_SHRINE,
 
+    // #### `result`
+    // ```c
+    // pauseCtx->state == PAUSE_STATE_OWL_WARP_SELECT
+    // ```
+    // #### `args`
+    // - None
+    VB_OWL_WARP_MENU_BE_INTUITIVE,
```
mckinlee: VB names are typically descriptive of the vanilla behavior. What do you think of the name `VB_OWL_WARP_USE_LINEAR_CURSOR`?

mm/2s2h/Enhancements/Songs/BetterOwlWarpMenu.cpp
```diff
@@ -0,0 +1,602 @@
...
+#include "2s2h/ShipInit.hpp"
+
+extern "C" {
+#include "src/overlays/kaleido_scope/ovl_kaleido_scope/z_kaleido_scope.h"
+#include "z64scene.h"
+
+extern PlayState* gPlayState;
+extern u16 sOwlWarpPauseItems[];
+
+extern bool Ship_HandleConsoleCrashAsReset();
```
mckinlee: `Ship_HandleConsoleCrashAsReset` appears to be unused. This needed?

mm/2s2h/Enhancements/Songs/BetterOwlWarpMenu.cpp
```diff
@@ -0,0 +1,602 @@
...
+    for (i = 0; i < NAV_BLOCK_SIZE; i++) {
+        tmpCursorPoint = navigationMap[*cursorPoint * NAV_BLOCK_SIZE + i];
+        if (pauseCtx->worldMapPoints[tmpCursorPoint]) {
+            *cursorPoint = tmpCursorPoint;
+            break;
+        }
+    }
+
+    return true;
+}
```
mckinlee, line 575: `BetterOwlWarp_NextCursorPoint` uses the cursor value as an array index with no validation. However, in BetterOwlWarp_IsCutoffOnSide it does have validation. Out of curiosity, why is that?

mm/2s2h/Enhancements/Songs/BetterOwlWarpMenu.cpp
```diff
@@ -0,0 +1,602 @@
...
+extern PlayState* gPlayState;
+extern u16 sOwlWarpPauseItems[];
+
+extern bool Ship_HandleConsoleCrashAsReset();
+}
+
+#define CVAR_NAME "gEnhancements.Songs.BetterOwlWarpMenu"
+#define CVAR CVarGetInteger(CVAR_NAME, 0)
+
+static constexpr s16 NAV_BLOCK_SIZE = OWL_WARP_MAX - 2;
```
mckinlee: Making a assumption here, but the intent is number of warp points minus one, correct? `OWL_WARP_STONE_TOWER - OWL_WARP_GREAT_BAY_COAST` would make it a bit clearer, no?

mm/2s2h/Enhancements/Songs/BetterOwlWarpMenu.cpp
```diff
@@ -0,0 +1,602 @@
...
+    PauseContext* pauseCtx = &gPlayState->pauseCtx;
+    s16 nextCursorPoint = pauseCtx->cursorPoint[PAUSE_WORLD_MAP];
+
+    if (BetterOwlWarp_NextCursorPoint(&nextCursorPoint, pauseCtx)) {
+        pauseCtx->cursorShrinkRate = 4.0f;
+
+        if (pauseCtx->cursorPoint[PAUSE_WORLD_MAP] != nextCursorPoint) {
+            pauseCtx->cursorPoint[PAUSE_WORLD_MAP] = nextCursorPoint;
+            Audio_PlaySfx(NA_SE_SY_CURSOR);
+        }
```
mckinlee, line 587: In vanilla the shrink, from what I recall, is always paired with movement, no? If there is no point to move to, this in practice would still do the shrink animation.

mckinlee (changes requested):
Looking good thus far. I did have a few questions.

mm/2s2h/Enhancements/Songs/BetterOwlWarpMenu.cpp
JordanLongstaff: I tested vanilla, and it shrinks even if there isn't a point to move to.

mm/2s2h/Enhancements/Songs/BetterOwlWarpMenu.cpp
JordanLongstaff: You're right. Good catch.

mm/2s2h/GameInteractor/GameInteractor_VanillaBehavior.h
JordanLongstaff: Sure, I like it.

mm/2s2h/Enhancements/Songs/BetterOwlWarpMenu.cpp
JordanLongstaff: I can add validation, but I think in practice it's only needed for Pause Owl Warp. I found that without the validation, if the cursor is on one of the page turn buttons, it skips over to the other one instead of touching any of the warp points.

mm/2s2h/Enhancements/Songs/BetterOwlWarpMenu.cpp
mckinlee: Ah cool. Learning something new. I see KaleidoScope_UpdateWorldMapCursor does the same thing.

mm/2s2h/Enhancements/Songs/BetterOwlWarpMenu.cpp
mckinlee: Ah, that makes sense. Was just curious. If it works as intended without it, all good.

JordanLongstaff: Even Pause Owl Warp has its own pseudo-validation before calling this function. Not quite the same AFAICT, but this function only gets called if `pauseCtx->cursorSpecialPos == 0`.

mckinlee (approved):

PR #1596, opened 2026-03-06
Add Reset function to enRailSkb

balloondude2:
There is a static variable in enRailSkb that facilitates spawning the three circling stalchildren. This looks like it should always be reset to 0 normally, but there have been reports of softlocks occurring because only one stalchild is spawning. 

I was able to replicate the softlock by artificially preventing the variable from resetting. 

This PR adds a Reset function to the actor to ensure this variable is always reset to 0. While I haven't been to naturally replicate, I'm confident this will prevent the softlock.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5792775985.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5792655538.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5792563014.zip)


Eblo (approved):

Eblo:
LGTM, but can this target `develop-keiichi`?

balloondude2:
Yeah, I'll probably just open another PR targeting keiichi instead. I think I'll have time to do that tonight

PR #1597, opened 2026-03-06
[Rando] Tycoon's Wallet

mckinlee:
I'm not married to the idea of the capacity being 5000. It would be less intrusive to do 999 like Ship does, but I'll leave that up for debate. If we decide to go with 999, easy change.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5977743211.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5977412199.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5977321400.zip)


Eblo (commented):
This checks out to me for the most part. Some thoughts:

- This combined with max Rupees on wallet upgrade means 2ship really ought to port over SoH's fast Rupee accumulator. 5000 Rupees can take a long time to fill to capacity and is rather grating to hear. I think that is out of scope for this PR but worth mentioning
- Fast Bank Deposit balks at the new max Rupee count. I'm not sure what the best way of addressing this might be, as the bank only has a three digit selection. Maybe just capping at 999 for the Tycoon's wallet is fine?
- Merge conflicts need addressed

Eblo (approved):

PR #1600, opened 2026-03-08
[Enhancement] Extra Powder Kegs

mckinlee:
Allows the player to carry up to 3 powder kegs in total. NPC Dialogue will reflect the new max.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6085407104.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6085349519.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6085276857.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Items/ExtraPowderKegs.cpp
+++ b/mm/2s2h/Enhancements/Items/ExtraPowderKegs.cpp
@@ -1,0 +1,59 @@
+#include <libultraship/bridge/consolevariablebridge.h>
+#include "2s2h/GameInteractor/GameInteractor.h"
+#include "2s2h/CustomMessage/CustomMessage.h"
+#include "2s2h/ShipInit.hpp"
+
+#define CVAR_NAME "gEnhancements.Items.ExtraPowderKegs"
+#define CVAR CVarGetInteger(CVAR_NAME, 0)
+#define MAX_POWDER_KEGS 3
+
+void RegisterExtraPowderKegs() {
+    // When giving a powder keg, increment ammo instead of setting to 1
+    COND_VB_SHOULD(VB_POWDER_KEG_SET_AMMO_ON_GIVE, CVAR, {
+        *should = false;
+        if (AMMO(ITEM_POWDER_KEG) < MAX_POWDER_KEGS) {
+            AMMO(ITEM_POWDER_KEG)++;
+        }
+    });
+
+    // When clamping ammo, use our max instead of 1
+    COND_VB_SHOULD(VB_POWDER_KEG_CAP_AMMO, CVAR, {
+        *should = false;
+        if (AMMO(ITEM_POWDER_KEG) > MAX_POWDER_KEGS) {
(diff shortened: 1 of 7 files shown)


mm/2s2h/Enhancements/Items/ExtraPowderKegs.cpp
```diff
@@ -0,0 +1,59 @@
...
+        CustomMessage::LoadCustomMessageIntoFont(entry);
+        *loadFromMessageTable = false;
+    });
+    COND_ID_HOOK(OnOpenText, 0x0673, CVAR, [](u16* textId, bool* loadFromMessageTable) {
+        auto entry = CustomMessage::LoadVanillaMessageTableEntry(*textId);
+        CustomMessage::Replace(&entry.msg, "one", "three");
+        CustomMessage::LoadCustomMessageIntoFont(entry);
+        *loadFromMessageTable = false;
+    });
+}
```
Eblo: We want to shy away from the pattern of replacing words in messages because that won't mesh well when other versions (PAL/JP) are supported, numerals being an exception. I recommend instead just creating entirely new dialog reflecting the higher max count, e.g. "Be careful, you can only carry up to three Powder Kegs at a time!"

Eblo (changes requested):
Works well in practice, just some notes on the dialog changes

mm/2s2h/Enhancements/Items/ExtraPowderKegs.cpp
mckinlee: Done. The dialogue now matches vanilla and I used the message viewer to grab the same control codes vanilla uses.

mm/2s2h/Enhancements/Items/ExtraPowderKegs.cpp
```diff
@@ -40,41 +40,43 @@ void RegisterExtraPowderKegs() {
...
+        entry.msg += " at a time. Use some up\x11"
+                     "and then come back.\xbf";
         CustomMessage::LoadCustomMessageIntoFont(entry);
         *loadFromMessageTable = false;
     });
     COND_ID_HOOK(OnOpenText, 0x0673, CVAR, [](u16*, bool* loadFromMessageTable) {
         CustomMessage::Entry entry;
         entry.autoFormat = false;
-        entry.msg = "\x1e\x38\xfc\x17Oh, but you already have \x01";
+        entry.msg = "\x1e\x38\xfc\x17Whoa, you're already\x11"
```
Eblo: One of these characters seems to be causing a blank textbox to flash before the intended text shows. I'm not sure which.

mckinlee: Should be good now. I simplified and documented the control codes used as well.

Eblo (approved):

PR #1603, opened 2026-03-14
[Fix] Fix Button Env Color

mckinlee:
The original fix only covered the ammo count digits, but the same missing `gDPSetEnvColor` bug shows up in a few other spots in `Interface_DrawBButtonIcons`. I hookified it (`VB_SET_BUTTON_ENV_COLOR`) and expanded the fix to cover all the affected call sites with bug documentation. The fix has been renamed to 'FixButtonEnvColor' and the tooltip updated accordingly.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5926929405.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5926733640.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5926690459.zip)


Eblo (approved):
Checks out to me, though I wonder if we will want some sort of config migration for the renamed setting.

PR #1605, opened 2026-03-18
[Enhancement] Faster Rupee Accumulator 

mckinlee:
Unlike the SoH version which uses a fixed step of 10, I scaled the step proportionally to wallet capacity (`CUR_CAPACITY / 50`) and always transfer`MIN(step, remaining)` per frame. This keeps fill time consistent regardless of wallet size. 

While working on this I also ran into two pre-existing bugs that had to be fixed in order to ship this cleanly:

`GiveItem.cpp` was setting `rupeeAccumulator = CUR_CAPACITY(UPG_WALLET)` (full capacity) instead of the remaining space when giving wallet upgrades. If you already had rupees, the excess accumulator would overflow into Auto Bank Deposit. Changed to `CUR_CAPACITY - current rupees`.

Starting a rando save with wallet starting items + Wallet Full enabled caused an immediate bank deposit. `GiveItem` sets the accumulator when granting starting wallets, but `StartingItems` then sets rupees directly, That leaves a stale accumulator that triggered `VB_DISCARD_EXCESS_RUPEES` on the first frame. Fixed by clearing the accumulator after the direct assignment.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5977401553.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5977320406.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5977229827.zip)

### DIFF
--- a/mm/2s2h/Rando/GiveItem.cpp
+++ b/mm/2s2h/Rando/GiveItem.cpp
@@ -143,7 +143,8 @@ void Rando::GiveItem(RandoItemId randoItemId) {
         case RI_WALLET_GIANT:
             Item_Give(gPlayState, Rando::StaticData::Items[randoItemId].itemId);
             // Fill Rupees to max, this may be opt-in later
-            gSaveContext.rupeeAccumulator = CUR_CAPACITY(UPG_WALLET);
+            // Use remaining space rather than full capacity to avoid excess in the accumulator.
+            gSaveContext.rupeeAccumulator = CUR_CAPACITY(UPG_WALLET) - gSaveContext.save.saveInfo.playerData.rupees;
             break;
         case RI_GS_TOKEN_SWAMP:
             // Set QUEST_QUIVER to match bug mentioned in z_parameter.c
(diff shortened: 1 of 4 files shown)


mm/2s2h/Rando/GiveItem.cpp
```diff
@@ -143,7 +143,8 @@ void Rando::GiveItem(RandoItemId randoItemId) {
         case RI_WALLET_GIANT:
             Item_Give(gPlayState, Rando::StaticData::Items[randoItemId].itemId);
             // Fill Rupees to max, this may be opt-in later
-            gSaveContext.rupeeAccumulator = CUR_CAPACITY(UPG_WALLET);
+            // Use remaining space rather than full capacity to avoid excess in the accumulator.
+            gSaveContext.rupeeAccumulator = CUR_CAPACITY(UPG_WALLET) - gSaveContext.save.saveInfo.playerData.rupees;
```
Eblo, line 147: Will this need `RI_WALLET_TYCOON` as well?

Eblo (commented):
LGTM, one blocking question now that Tycoon's Wallet is merged

mm/2s2h/Rando/GiveItem.cpp
mckinlee: Can you clarify your question? If you mean that tycoon wallet should use the remaining space as well, it already does.

Eblo: You're right; Tycoon's wallet has its own case already.

Eblo (approved):

PR #1631, opened 2026-04-06
Enhancement to add Platformer style jumps

ghost:
Same Idea, New Account, better code, and better execution .

adds 3 new movement options
Jumping with A while moving, with the option to double and triple jump with height based on XZ speed much like the vanilla ledge jump
Wall kicks modeled on sm64's mechanics but with a wider effective angle, 
Spin lift (based on link's Up B in Melee) to get a little extra distance or height (also a fun combat option)

I did a lot of testing on these to make sure the are not OP, don't lead to softlocks, or give free access to glitches (jumping out of a JS for free ISG, or jumping from a ledge grab for free Down A, or shielding a chu mid spin lift for an instant hover)

I did a reasonable amount of testing with zora and FD, and enough with goron an deku to know that they needed a "disable jumping" option.

Did a lot of testing as human, and left some notes about some fun options this opens up in movement that could be consider glitches.

Happy to make any changes or tweaks.

Again AI disclosure, claude wrote the code, but I did a lot of audit and clean up passes, tried to add as many comments as I could and did passes thru a few other LLMs to watch for jank or sloppy code.

also IDK if you guys squash commits on merge, but there is no real reason to keep the full list of commits on this IMO

but I leave that up to you

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6382460375.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6382420824.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6382352204.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1008,12 +1008,14 @@ void BenMenu::AddEnhancements() {
             "wall kicks (A off walls while airborne), and spin lift (B in air for sword helicopter). "
             "Z + A/B still do normal attacks. Sword type scales spin lift height."));
     AddWidget(path, "  Deku Chain Jump", WIDGET_CVAR_CHECKBOX)
-        .CVar("gEnhancements.Player.PlatformerMode.DekuJump")
+        .CVar("gEnhancements.Player.PlatformerModeDekuJump")
+        .PreFunc([](WidgetInfo& info) { info.isHidden = !CVarGetInteger("gEnhancements.Player.PlatformerMode", 0); })
         .Options(CheckboxOptions().Tooltip(
             "Allows Deku Link to chain jump and wall kick in Platformer Mode. "
             "Off by default since Deku jumps are weak."));
     AddWidget(path, "  Goron Chain Jump", WIDGET_CVAR_CHECKBOX)
-        .CVar("gEnhancements.Player.PlatformerMode.GoronJump")
+        .CVar("gEnhancements.Player.PlatformerModeGoronJump")
+        .PreFunc([](WidgetInfo& info) { info.isHidden = !CVarGetInteger("gEnhancements.Player.PlatformerMode", 0); })
         .Options(CheckboxOptions().Tooltip(
             "Allows Goron Link to chain jump and wall kick in Platformer Mode. "
             "Off by default since Goron movement is roll-based."));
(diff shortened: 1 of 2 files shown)


IncomprehensibleNonsense76:
spin-lift
triple-jump
wall-kick


Demo Gifs

mckinlee (commented):
I’m going to be honest:

This is a pretty invasive movement enhancement. It touches player input, player action state, static state, cvar gated behavior, form specific movement, and existing enhancements like Manual Jump. That’s a risky part of the codebase to begin with, and the current implementation raises enough questions that I don’t think leaving a bunch of small review comments is going to get this to a mergeable place.

I appreciate the AI disclosure, but it also makes me less confident here, not more. If Claude wrote the bulk of this and then other LLMs were used to check it, I’d expect the end result to be especially clean and easy to reason about. Right now this feels more like a big experimental patch than something ready for review.

A few examples of what makes me uneasy:
- static state living inside cvar gated hooks
- unclear behavior if the feature is toggled off midair or mid-action
- the tooltip and input suppression logic not seeming to fully line up
- B/C inputs getting suppressed after spin lift
- possible conflict with Manual Jump
- assumptions around sword/held item state

Maybe some of that is explainable. Maybe some of it is fixable. But taken together, I don’t think this should be merged as is.

My preference would be to close this for now and come back with something smaller and more deliberate: clearer goals, clear interaction rules with existing movement enhancements, and changes that are easier to review and test.

mckinlee:
Also, thinking about it more, this might just be better suited as a recomp/mod thing rather than a core 2ship enhancement.

That’s not me saying the idea is bad. It’s a big experimental movement overhaul, and that kind of thing can be cool. I just don’t think the main repo is the right place for this version of it, especially with how much player/input behavior it touches.

ghost:
After reading this I honestly thing I agree with you and this makes much more sense as a mod not a native feature. Also thank you for giving specific feedback and not just "This code looks wrong and I don't like it". C & C++ are pretty much latin to me, I can read the letters and translate them to something I can understand, but Async Graphical Stateful applications are not my forte what so ever (I do web APIs and like.. devops) and I really appreciate your perspective.

If there is like a native modding framework for gameplay modifications that did not expose the core codebase to (TBH IMO) slop code with a fair amount  play testing, I 100% think that would make sense for fun janky mods like this.

Thanks a lot for taking the time to read over this and also just in general thank you guys for making 2ship the absolute best way to play this game, I can't go back to dolphin now XD.

Cheers :)

PS the spinlift is janky as hell and took me 4 times longer to get working that the other 2 parts of this if it was not for how natural it felt as part of the package of movement options I would have scrapped it.

PR #1632, opened 2026-04-07
Erase warning from Owl Save message when persistent option is on

### LINKED ISSUE #1630: [Feature request] Persistent Owl Save skips warning about losing save
When you save at an Owl Statue, it warns you that if you don't save after reloading your game back up, the file will revert to the last Song of Time save. However, with Persistent Owl Save turns on, this will be false.

JordanLongstaff:
Closes #1630

This warning now gets skipped when "Persistent Owl Saves" is turned on:

https://github.com/user-attachments/assets/8ca5ed5a-60b0-4514-8cce-35698ad3b854

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6385898206.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6385858053.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6385819871.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Dialogue/SkipOwlSaveWarning.cpp
+++ b/mm/2s2h/Enhancements/Dialogue/SkipOwlSaveWarning.cpp
@@ -1,7 +1,8 @@
 #include <libultraship/bridge/consolevariablebridge.h>
-#include "2s2h/GameInteractor/GameInteractor.h"
-#include "2s2h/ShipInit.hpp"
 #include "2s2h/CustomMessage/CustomMessage.h"
+#include "2s2h/GameInteractor/GameInteractor.h"
+#include "2s2h/BenPort.hpp"
+#include "2s2h/ShipInit.hpp"
 
 #define CVAR_NAME "gEnhancements.Saving.PersistentOwlSaves"
 #define CVAR CVarGetInteger(CVAR_NAME, 0)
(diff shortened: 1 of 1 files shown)


mm/2s2h/Enhancements/Dialogue/SkipOwlSaveWarning.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+#define CVAR_NAME "gEnhancements.Saving.PersistentOwlSaves"
+#define CVAR CVarGetInteger(CVAR_NAME, 0)
+
+// "You can save your progress and quit here."
+static constexpr u16 TEXT_ID_OWL_SAVE = 0xC01;
+
+// "Warning: If you reopen this Owl File, then reset without saving..."
+static constexpr size_t TEXT_WARNING_BEGIN = 258;
+static constexpr size_t TEXT_WARNING_LENGTH = 261;
```
Eblo: I'm wary of this because this substring range may not be consistent between versions, which will be a concern when PAL and JP GC support get added. So far we've taken care to minimize dependency on text contents with other enhancements and randomizer text. Usually the approach there would be to create brand new text, but I suspect that the dialog choice here might complicate doing that here.

JordanLongstaff: I've just added a version check, so now the message is only modified on NTSC versions. Better?

Eblo (approved):

PR #1634, opened 2026-04-07
Adjust dialogue after failing boat archery with custom health setting

JordanLongstaff:
The boat archery health hook now tweaks the number of times Koume says she hit you (she always said 10 before) to be correct to the actual number of hits.

One hit (singular, edge case):

Between 1 and 10 hits:

More than 10 hits (maxed out to 30 in this case):

Sorry for the torture, Koume, but it had to be done.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6313659262.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6313468851.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6313325945.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Minigames/Archery.cpp
+++ b/mm/2s2h/Enhancements/Minigames/Archery.cpp
@@ -1,4 +1,5 @@
 #include <libultraship/bridge/consolevariablebridge.h>
+#include "2s2h/CustomMessage/CustomMessage.h"
 #include "2s2h/GameInteractor/GameInteractor.h"
 #include "2s2h/ShipInit.hpp"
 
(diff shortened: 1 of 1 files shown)


mm/2s2h/Enhancements/Minigames/Archery.cpp
```diff
@@ -73,8 +76,18 @@ static bool ShouldFailBoatArchery() {
     return gSaveContext.minigameHiddenScore >= BOAT_HEALTH_CVAR;
 }
 
+static void ModifyHitCount(u16* textId, bool* loadFromMessageTable) {
+    auto entry = CustomMessage::LoadVanillaMessageTableEntry(*textId);
+    u16 hits = gSaveContext.minigameHiddenScore;
+    CustomMessage::Replace(&entry.msg, "10 times", std::to_string(hits) + " time" + (hits == 1 ? "" : "s"));
```
Eblo, line 82: Same deal as your owl message PR, although given that there is a numeral at play, it would be acceptable if only the numeral itself was replaced with `std::to_string(hits)`. I don't think "1 times" would be too jarring even if it's incorrect.

JordanLongstaff: At least this time it's a string replace instead of a chunk deletion with hard-coded indices. Nothing should happen if the string "10 times" isn't found.

And, sorry, but I'm gonna insist on making it singular if it's just one.

Eblo (approved):

PR #1647, opened 2026-04-13
[Enhancement] extended keyboard controls (ESS, Quickspins, Half-Stick, Drifted Backflips)

ghost:
Ok this one is actually useful

This lets keyboard players (like me) have access to all the movement options that can't be done with only digital inputs
this adds 4 features, all of which can be individually enabled or disabled

1. ESS Button: Very simple, hold a modifier key and all stick inputs are clamped to 17/127 to give you an ESS position (could also be useful for controller players to get free ESS)
2. Half-Stick Inputs: A bit more complicated but just as important. This let's you input a half-input on any direction, this lets link slow walk, and allow you to use a full digital input + half input to access the up upleft/right needed for twisted backflips or un-target angle changes in hovers.
3. Notch correction: This is a bit of a hack where for 1 frame while holding downleft or downright and pressing A the left/right part of the input is ignored for 1 frame to allow for drifted backfips, since the perfect diagonals on keyboard always resolve to a sidehop. This is important for tricks like the first part of swamp gossips that needs drifted backflips, as does the method of SOT storage in SHT that is done right after getting fire arrows in the old Any% NMG route.
4. Free quick spins: Pretty simple, hold a modifier key and then press B or GFS and you get a quick spin, also works for JS cancels which is great for iron kunkles or Wrath (also for the Tag Strat on incarnation).

Did my best to pile comments into the code and keep changes to z_player.c and the VB hooks as surgical as I could.

Again claude code wrote the code, but I did several audit passes and a boatload of testing.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7581974007.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7581916119.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7581797842.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenGui.cpp
+++ b/mm/2s2h/BenGui/BenGui.cpp
@@ -6,6 +6,7 @@
 #include "UIWidgets.hpp"
 #include "HudEditor.h"
 #include "2s2h/Enhancements/Audio/AudioEditor.h"
+#include "2s2h/Enhancements/ModMenu/ModMenu.h"
 #include "CosmeticEditor.h"
 #include "Notification.h"
 #include "2s2h/Rando/CheckTracker/CheckTracker.h"
@@ -55,6 +56,7 @@ std::shared_ptr<EventLogWindow> mEventLogWindow;
 std::shared_ptr<DLViewerWindow> mDLViewerWindow;
 std::shared_ptr<MessageViewerWindow> mMessageViewerWindow;
 std::shared_ptr<AudioEditor> mAudioEditorWindow;
+std::shared_ptr<ModMenuWindow> mModMenuWindow;
 std::shared_ptr<BenMenu> mBenMenu;
 std::shared_ptr<Notification::Window> mNotificationWindow;
 std::shared_ptr<Rando::CheckTracker::CheckTrackerWindow> mRandoCheckTrackerWindow;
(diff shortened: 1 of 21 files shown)


ghost:
OK so after a month of testing  and 1 fix I think this is pretty solid and matches what I used to be able to do on dolphin (and actually more with quickspins/js cancels). Below are a few examples of usage.

Example of using half inputs for hover angle changes & twisted backflips
RE-Angle and Twist-00 00 00 000-00 00 14 892

Example of ESS positions for HESSIng
HESS

Example of drifted backflips in swamp gossips
Drifted Backflip

Example of using ESS stick for precise aim 
Precise Aim

Example of quickspins
QuickSpins


All of this was done on keyboard, unfortunately due to the way things are implemented the input display does not show show the modified inputs. However I do think this is a useful feature that opens up the full movement expression of the game to people who play without a controller, while also bundling some helpful "Easy Mode" features for people who do play on controller but have issues holding ESS or getting consistent quickspins (TBH I wanted this to make grass farming in rando tolerable XD).

I believe the code is of acceptable quality, but again... C++ is not my native tongue and I am happy to make any and all changes that the team believes are needed.

PS weird colors in the gifs are from reduced color pallets to get them down small enough to upload to github and not how the game looks.

mckinlee (changes requested):
I've not done any in game testing. this is strictly from reading the code.

In the PR description you say each feature can be toggled individually, but that doesn't seem to be the case from what I can see. Only notch has its own toggle; ESS, quick spin, and half stick all run whenever the master toggle is on.

There's a third Player_CanSpinAttack in func_80840CD4 that isn't wrapped (the other two call sites are). Might be fine for the main keyboard quick-spin path since that goes through func_808335F4, but can you confirm behavior with magic spin charge + the modifier held? 

The half-stick key rebinding UI is pretty chunky and not a pattern I usually see in BenMenu. would be cleaner as a helper in the enhancement file and just wired up from the menu.

Also not a huge fan of how verbose the comments are. The code should mostly speak for itself with comments only where something actually needs explaining. otherwise it's a lot of noise. More opinion than requirement though.

ghost:
OK cool I was planning on taking another pass over comments (My god I did not realize how verbose they were, it's been like a month since I was working on this) and over the menu verbiage to be less... bad....

I'll see what I can do for the half stick input system.

Funny enough you and claude both flagged the charge spin while holding quickspin issue. And basically you end up quick spinning on the initial B press then if you keep holding both you will charge a normal spin attack. (Will include a gif in follow up)

I'll confirm the individual toggles, I honestly use all of them together, so toggling them would be an odd case but I did feel it was worth it over just having a giant "everything" switch.

ghost:
Ok made the requested fixes, cleaned up the info text, and caught an edge case where if you are using inverted cam the notch correction would not apply and you would get a sidehop. 

quickspin modifier + held B
quickspin +hold b


I also moved the sub items to use the same ImGui indent from the platforming tricks branch. and I also validated all features respect their toggles.

PR #1659, opened 2026-04-27
Skip Misc Interactions: Couple's Mask cutscene

JordanLongstaff:
https://github.com/user-attachments/assets/8999c4cb-e46b-48ce-91f1-14c1a6e06ed7

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7125155700.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7125137018.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7124993146.zip)


mm/2s2h/GameInteractor/GameInteractor_VanillaBehavior.h
```diff
@@ -764,6 +764,14 @@ typedef enum {
     // - `*EnTalkGibudRequestedItem`
     VB_GIBDO_TRADE_SEQUENCE_TAKE_MORE_THAN_ONE_ITEM,
 
+    // #### `result`
+    // ```c
+    // true
+    // ```
+    // #### `args`
+    // - None
+    VB_GIVE_COUPLES_MASK,
```
Eblo: This VB appears to be unused.

mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipCouplesMaskCutscene.cpp
```diff
@@ -0,0 +1,65 @@
...
+    EnAn* anju = (EnAn*)SubS_FindActor(gPlayState, NULL, ACTORCAT_NPC, ACTOR_EN_AN);
+    if (anju != NULL) {
+        anju->unk_3C0 = true;
+        kafei->player.actor.world.pos = anju->actor.world.pos = POSITION;
+        kafei->player.actor.world.rot = kafei->player.actor.shape.rot = anju->actor.world.rot = anju->actor.shape.rot =
+            ROTATION;
+        kafei->player.yaw = ROTATION.y;
+    }
+
+    if (GameInteractor_Should(VB_GIVE_COUPLES_MASK, true)) {
```
Eblo: As-is, this gives the Couple's Mask in rando in addition to the check. See `SkipRosaSisters.cpp` and `GalleryTwofer.cpp` for other instances of item queues being disregarded in rando. Additionally, this VB appears to be unused.
```suggestion
    if (!IS_RANDO) {
```

mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipCouplesMaskCutscene.cpp
```diff
@@ -0,0 +1,65 @@
...
+                    if (CUSTOM_ITEM_FLAGS & CustomItem::GIVE_ITEM_CUTSCENE) {
+                        CustomMessage::SetActiveCustomMessage(text.msg, text);
+                    } else {
+                        CustomMessage::StartTextbox(text.msg + "\x1C\x02\x10", text);
+                    }
+
+                    Item_Give(gPlayState, ITEM_MASK_COUPLE);
+                    Message_BombersNotebookQueueEvent(play, BOMBERS_NOTEBOOK_EVENT_MET_ANJU);
+                    Message_BombersNotebookQueueEvent(play, BOMBERS_NOTEBOOK_EVENT_MET_KAFEI);
+                    Message_BombersNotebookQueueEvent(play, BOMBERS_NOTEBOOK_EVENT_RECEIVED_COUPLES_MASK);
```
Eblo: Notebook messages can be really finicky with the message state, with multiple moving pieces with the skipped dialog, the player's item notifications setting, whether the player has the notebook, and whether the player is doing a repeat of an event. Your current implementation allows the player to walk freely with the item notification under some configurations, thereby missing the notebook event messages.

This can be and has been a royal pain. One reliable pattern I found was to queue these events from within the last text ID of a given conversation, allowing the notebook and item messages to fire off from an expected state. However, this enhancement skips the entire cutscene, so there is no convenient text box to piggyback off of. Some ad hoc solution may have to be applied here to ensure the player is locked in place when there is actual text to wait for but not when there isn't.

Additionally, following my rando condition suggestion, the notebook events need to be separated from the item queue itself, as they should happen regardless of whether this is rando or not.

Eblo (changes requested):

mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipCouplesMaskCutscene.cpp
JordanLongstaff: What about queuing up a message here with the player frozen? Perhaps the hook opens the "Please take refuge, we are fine here" text, whether the mask was given or not.

mckinlee (commented):
In SkipHandleCouplesMaskCs, does the !IS_RANDO guard around GameInteractor_Should(VB_GIVE_ITEM_FROM_DMCHAR05, true, ITEM_MASK_COUPLE) prevent the rando hook in DmChar05.cpp from marking RC_STOCK_POT_INN_COUPLES_MASK eligible?

My understanding is that the skipped vanilla cutscene would normally reach the VB_GIVE_ITEM_FROM_DMCHAR05 path, and rando relies on that VB firing to credit the randomized Couple’s Mask check. If this skip avoids calling that VB in rando, could rando players miss the check/reward when this enhancement is enabled?

JordanLongstaff:
I don't know, and I'm not willing to test rando to find out. The other places where `VB_GIVE_ITEM_FROM_DMCHAR05` gets used in cutscene skips don't have `IS_RANDO` checks. Is it common practice to put that check after the hook?

mckinlee:
> I don't know, and I'm not willing to test rando to find out. The other places where `VB_GIVE_ITEM_FROM_DMCHAR05` gets used in cutscene skips don't have `IS_RANDO` checks. Is it common practice to put that check after the hook?

I haven't tested it either, but the pattern in `SkipHealingDarmani.cpp`, `SkipHealingMikau.cpp`, and `SkipIkanaCurseCutscenes.cpp` seems to be to call `GameInteractor_Should(VB_GIVE_ITEM_FROM_DMCHAR05, true, ITEM_MASK_*)` unconditionally, then only queue the vanilla item if it returns true. In rando, the hook handles eligibility and returns false, preventing the vanilla item from being queued. 

@Eblo What are your thoughts since this is adjacent to your earlier feedback?

mm/2s2h/Enhancements/Cutscenes/MiscInteractions/SkipCouplesMaskCutscene.cpp
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    anju->unk_3C0 = true;
+    kafei->player.actor.world.pos = anju->actor.world.pos = POSITION;
+    kafei->player.actor.world.pos.z -= 2.0f;
+    kafei->player.actor.world.rot = kafei->player.actor.shape.rot = anju->actor.world.rot = anju->actor.shape.rot =
+        ROTATION;
+    kafei->player.yaw = ROTATION.y;
+
+    if (!IS_RANDO && GameInteractor_Should(VB_GIVE_ITEM_FROM_DMCHAR05, true, ITEM_MASK_COUPLE)) {
```
Eblo: The `!IS_RANDO` is redundant here, as the other processing for `VB_GIVE_ITEM_FROM_DMCHAR05` is handled from within code that only executes within randomizer.
```suggestion
    if (GameInteractor_Should(VB_GIVE_ITEM_FROM_DMCHAR05, true, ITEM_MASK_COUPLE)) {
```

Eblo (changes requested):
I should have caught that condition before. `VB_GIVE_ITEM_FROM_DMCHAR05` is implicitly handled from within randomizer-only code, so it is not necessary to check it again. That is why the current pattern does not check for it. As-is, it does work across vanilla and rando, with and without the cutscene skip, but best to clean up the redundant condition.

Loving the solution to the notebook situation too! It works across all scenarios of vanilla/rando, item get skip on/off, and with/without notebook.

Eblo (approved):

PR #1678, opened 2026-05-08
[Fix] Timer Mode Resetting After Restart

mckinlee:
Fixes the in-game timer option not sticking when set to In-Game Time.

The problem was that `gWindows.DisplayOverlay` was being used for two different things at the same time:

- whether the overlay window is visible
- which timer mode is selected

That worked fine for Off and Real-Time, but In-Game Time uses value `2`. Since the window visibility stuff treats that CVar like a bool, it would turn `2` into `1` on restart, which made the setting come back as Real-Time.

This splits the timer mode out into its own CVar:

- `gWindows.DisplayOverlay` = window visible/hidden
- `gDisplayOverlay.Mode` = Off / Real-Time / In-Game Time

~Also added a small migration for the old values so existing configs keep whatever timer option they had selected.~

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6997070480.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6996916898.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/6996735943.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenGui.cpp
+++ b/mm/2s2h/BenGui/BenGui.cpp
@@ -154,6 +154,7 @@ void SetupGuiElements() {
                                                                              "Item Tracker Settings", ImVec2(800, 400));
     gui->AddGuiWindow(mItemTrackerSettingsWindow);
 
+    DisplayOverlay_Init();
     mDisplayOverlayWindow = std::make_shared<DisplayOverlayWindow>("gWindows.DisplayOverlay", "Display Overlay");
     gui->AddGuiWindow(mDisplayOverlayWindow);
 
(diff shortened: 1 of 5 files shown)


Eblo:
The change itself is good overall and fixes the bug in question. The migration I'm hesitant on because of how boilerplate it is. I think there is a larger discussion to be had on how to implement config migration in 2ship. [SoH has a centralized approach that might be worth following, or at least using for reference](https://github.com/HarbourMasters/Shipwright/blob/a6ceda989eecdfdba2a96c9b530d056218dd5d37/soh/soh/config/ConfigUpdaters.cpp).

mckinlee:
> The change itself is good overall and fixes the bug in question. The migration I'm hesitant on because of how boilerplate it is. I think there is a larger discussion to be had on how to implement config migration in 2ship. [SoH has a centralized approach that might be worth following, or at least using for reference](https://github.com/HarbourMasters/Shipwright/blob/a6ceda989eecdfdba2a96c9b530d056218dd5d37/soh/soh/config/ConfigUpdaters.cpp).

I think it'd just be best to remove the migration then? It felt warranted at the time, but meh. Maybe add the config updater as a nice to have for 4.1.0 release.

Eblo:
I'm in favor of that. There was something else changed since last release that could warrant a migration, but I can't remember what it was exactly.

mckinlee:
> I'm in favor of that. There was something else changed since last release that could warrant a migration, but I can't remember what it was exactly.

iirc, you mentioned you wanted one for when I switched FixAmmoCountEnvColor to FixButtonEnvColor. Maybe that one. But yeh, when I get a sec I'll update this.

Eblo (approved):

Eblo:
FYI I noted a config migration as its own item in the 4.1.0 checklist.

PR #1679, opened 2026-05-08
Disable SFX replacement

### LINKED ISSUE #1599: Custom Audio: Changing certain tracks alters some sound effects.

### Description of the bug:
Changing moon's destruction and tatl & tael tracks to use a different track will cause certain sound effects to change. For example the fire arrow and ice arrow sounds. When using fire or ice arrows they will play different sounds depending on what track they are set to. Landing a hit on igos du ikana when his back is turned may also exhibit this behavior.

1. Set moon's destruction and tatl & tael tracks to any other track.
2. Observe.

### Expected behavior:
For sound effects to sound normal.

### Screenshots or videos:
https://github.com/user-attachments/assets/0c52862a-c161-4dd4-8985-91f8f351e48d

https://github.com/user-attachments/assets/330849b0-3cd9-42bc-9ad6-1fd1a5597251

### System Information:
- OS: LMDE7 Debian 13 Linux
- GFX Backend: OpenGL
- Game Version: Mion Charlie 3.0.2
- Rom Version: US

Eblo:
tl;dr the audio editor's sequence replacement does not differentiate between the types of audio. For example, the Tatl & Tael track and `NA_SE_PL_ARROW_CHARGE_FIRE` both have values of 0x7D, which results in fire arrows attempting to load another sequence when Tatl & Tael is modified. As we do not currently shuffle SFX, I simply commented out the replacement and added a note.

The long term fix, once we do shuffle SFX, may be to separate `mSequenceMap` the different types of sequence.

Fixes #1599

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7046975046.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7046394759.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7046343315.zip)

### DIFF
--- a/mm/src/audio/sfx.c
+++ b/mm/src/audio/sfx.c
@@ -223,7 +223,9 @@ void AudioSfx_ProcessRequest(void) {
     if (req->sfxId == 0) {
         return;
     }
-    u16 newSfxId = AudioEditor_GetReplacementSeq(req->sfxId);
+    // All custom sequences use the same collection, which can result in collisions when different types (e.g. SFX and
+    // BGM) share the same IDs. As we do not yet shuffle SFX, simply do not get the replacement sequence for now.
+    u16 newSfxId = req->sfxId; // AudioEditor_GetReplacementSeq(req->sfxId);
     if (req->sfxId != newSfxId) {
         gAudioCtx.seqReplaced[SEQ_PLAYER_SFX] = 1;
         req->sfxId = newSfxId;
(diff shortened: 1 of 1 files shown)


mckinlee (approved):
Good as is, but the `if (req->sfxId != newSfxId)` block is now unreachable. Ain't harming anything, but it could be commented out until we have proper type aware sfx/seq mapping.

PR #1690, opened 2026-05-14
[Enhancement] Tingle always in Clock Town

### LINKED ISSUE #1682: [Enhancement] Tingle appears in Clock Town during the night like in MM3D
In the N64 version of Majora's Mask, for whatever reason, Tingle only appears in Clock Town during the daytime, completely disappearing during the night. In the 3DS version, however, this was changed so that he always appears in Clock Town no matter what time it is.

Since there are other enhancements in 2Ship that make the game more like the 3DS version, I'm suggesting another one that makes it so Tingle will appear in Clock Town both day and night just like in Majora's Mask 3D. It could at least possibly be useful in randomizer.

As proof of this difference between the two versions of the game, here are two collages of North Clock Town during every half-day on both N64 and 3DS (taken from 2Ship and the Citra emulator, respectively), showcasing how Tingle is always there on 3DS but is only present during the day on N64. Each row in each image is a different day, with daytime on the left and nighttime on the right.

N64 version:
Image

3DS version:
Image

JordanLongstaff:
Closes #1682

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7035130953.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7035070268.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7035003685.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1227,7 +1227,9 @@ void BenMenu::AddEnhancements() {
         .Options(CheckboxOptions().Tooltip("Allows the wallet reward to be collected on any day."));
     AddWidget(path, "Tingle Always in Clock Town", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.Cycle.TingleAlwaysInClockTown")
-        .Options(CheckboxOptions().Tooltip("Tingle will always appear in North Clock Town, not just during the day."));
+        .Options(CheckboxOptions().Tooltip("Tingle will always appear in North Clock Town, not just during the day. If "
+                                           "you turn this off at night while Tingle is present, he will remain until "
+                                           "the scene is reset."));
 
     //// Graphics Enhancements
     path = { "Enhancements", "Graphics", SECTION_COLUMN_1 };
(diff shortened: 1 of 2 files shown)


mckinlee (commented):
I tested this locally and disabling the enhancement does not return Tingle to vanilla behavior until 2ship is fully restarted.

Looks like it's directly modifying Tingle’s setup actor `rot.x` / `rot.z`, which are where the half day spawn bits live. Since that setup actor data appears to remain loaded/cached, the patched schedule bits keep being reused even after the CVAR is disabled.

I'm note sure the best route here, but I think it'd be better to not mutate the actor entry data directly. What are your thoughts on adding a new `GameInteractor` hook inside `Actor_SpawnSetupActors` where the vanilla half day spawn decision is made? That function seems to already compute whether each setup actor should spawn based on the halfday bits before calling `Actor_SpawnEntry`.

JordanLongstaff:
I like my solution for its elegance, but I can confirm the issue that you mentioned. I've now updated the mod to have a second hook that reverts the setup actor data once Tingle is spawned.
I also updated it to spawn Tingle instantly if the player turns the option on at night while in North Clock Town. If the player turns it off, Tingle won't despawn until the scene is reset. (I tried, but he'd also despawn during the day, which is unwanted behaviour, and ATM I'm too lazy to attempt a night check.)

JordanLongstaff:
> I also updated it to spawn Tingle instantly if the player turns the option on at night while in North Clock Town. If the player turns it off, Tingle won't despawn until the scene is reset. (I tried, but he'd also despawn during the day, which is unwanted behaviour, and ATM I'm too lazy to attempt a night check.)

Update: added the night check, now he does despawn.

mckinlee (approved):

PR #1692, opened 2026-05-16
[Enhancement] SoT does not reset Scarecrow's Song

JordanLongstaff:
Closes out the first half of discussion #684 (the second half is already done).

I also split up all the EoC Save hooks by the appropriate setting so that one change doesn't reevaluate all of them.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7038277374.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7038250149.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7038186257.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1212,6 +1212,9 @@ void BenMenu::AddEnhancements() {
         .CVar("gEnhancements.Cycle.DoNotResetChateau")
         .Options(CheckboxOptions().Tooltip(
             "Playing the Song of Time will not reset the infinite magic status granted by Chateau Romani."));
+    AddWidget(path, "Do not reset Scarecrow's Song", WIDGET_CVAR_CHECKBOX)
+        .CVar("gEnhancements.Cycle.DoNotResetScarecrowSong")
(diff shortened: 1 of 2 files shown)


balloondude2:
Does this work if you close 2ship between cycles?

JordanLongstaff:
> Does this work if you close 2ship between cycles?

I just tested, and no. 😞
I've just figured out the reason why, though. While the game does correctly load a Scarecrow's Song that was saved on an Owl Save, it does not do so for a Cycle Save. This means that, if you were to have this option turned on, and you:
1. Opened File 1
2. Recorded a Scarecrow's Song
3. Played the Song of Time (to supposedly save the Scarecrow's Song)
4. Reset
5. Opened File 2
6. Recorded a different Scarecrow's Song
7. Reset
8. Opened File 1 again
9. Tried to play the Scarecrow's Song

The Scarecrow's Song you recorded for File 1 would not work, but the one you recorded for File 2 would. The same is also true if File 2 was an Owl Save that already had a Scarecrow's Song recorded; that one would be used for File 1 as well, instead of the correct one.
I believe I know how to fix this. It will require adding another hook.

Eblo (approved):

PR #1711, opened 2026-05-25
[Enhancement] Port mod menu from SoH

Eblo:
image

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7225152973.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7224954738.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7224766311.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -718,7 +718,7 @@ void BenMenu::AddSettings() {
     AddSidebarEntry("Settings", path.sidebarName, 2);
     AddWidget(path, "Mod Menu", WIDGET_SEPARATOR_TEXT);
     AddWidget(path, "Popout Mod Menu Window", WIDGET_WINDOW_BUTTON)
-        .CVar("gWindows.InputViewer")
+        .CVar("gWindows.ModMenu")
         .WindowName("Mod Menu")
         .HideInSearch(true)
         .Options(ButtonOptions().Tooltip("Enables the separate Mod Menu Window."));
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -711,6 +711,17 @@ void BenMenu::AddSettings() {
         .CVar("gWindows.InputViewerSettings")
         .WindowName("Input Viewer Settings")
         .Options(ButtonOptions().Tooltip("Enables the separate Input Viewer Settings Window."));
+
+    // Mod Menu
+    path.sidebarName = "Mod Menu";
+    path.column = SECTION_COLUMN_1;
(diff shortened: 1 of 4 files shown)


Jameriquiah:
nice choice of mod to showcase it off with mr 2ship o7

mm/2s2h/BenGui/BenMenu.cpp
```diff
@@ -711,6 +711,17 @@ void BenMenu::AddSettings() {
...
         .WindowName("Input Viewer Settings")
         .Options(ButtonOptions().Tooltip("Enables the separate Input Viewer Settings Window."));
+
+    // Mod Menu
+    path.sidebarName = "Mod Menu";
+    path.column = SECTION_COLUMN_1;
+    AddSidebarEntry("Settings", path.sidebarName, 2);
+    AddWidget(path, "Mod Menu", WIDGET_SEPARATOR_TEXT);
+    AddWidget(path, "Popout Mod Menu Window", WIDGET_WINDOW_BUTTON)
+        .CVar("gWindows.InputViewer")
```
mckinlee: Why is this coupled with the InputViewer?

mm/2s2h/Enhancements/ModMenu/ModMenu.cpp
```diff
@@ -0,0 +1,399 @@
...
+                        changed = true;
+                    }
+                }
+            }
+        }
+        if (changed) {
+            SetEnabledModsCVarValue();
+        }
+    }
+}
```
mckinlee, line 180: What happens when two files share the same name but in different folders.. say two files like `mods/a/foo.o2r` and `mods/b/foo.o2r`?

It looks like both would get added to `enabledModFiles`, but `filePaths` only stores the first path?

mm/2s2h/BenGui/BenMenu.cpp
Eblo: Big goof, thanks for the catch

mm/2s2h/Enhancements/ModMenu/ModMenu.cpp
Eblo: It prefers whichever one it finds first, not loading the other one. This may be a bug for SoH's mod menu as well.

Eblo: Given that the chances are low of somebody having identically named files in different subfolders, and that this goes by the immediate filename, I think it is acceptable to leave this quirk be for now to be returned to later.

mckinlee: Yeh agreed.

PR #1719, opened 2026-05-29
add audio dependencies to build instructions

stellarkookies:
these four (libopus, libopusfile, libvorbis, and libogg) were required for building the default target on Debian 13, and were not included in the dependency lines. i do not use Fedora or openSUSE very actively, so the package names are based on whatever I could find on their respective package portals. feel free to correct me on this!

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7282821908.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7282744333.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7282650041.zip)

### DIFF
--- a/docs/BUILDING.md
+++ b/docs/BUILDING.md
@@ -89,34 +89,34 @@ C:\Program Files\CMake\bin\cmake.exe --build build-cmake --target ExtractAssetHe
 #### Debian/Ubuntu
 ```sh
 # using gcc
-apt-get install gcc g++ git cmake ninja-build lsb-release libsdl2-dev libpng-dev libsdl2-net-dev libzip-dev zipcmp zipmerge ziptool nlohmann-json3-dev libtinyxml2-dev libspdlog-dev libopengl-dev
+apt-get install gcc g++ git cmake ninja-build lsb-release libsdl2-dev libpng-dev libsdl2-net-dev libzip-dev zipcmp zipmerge ziptool nlohmann-json3-dev libtinyxml2-dev libspdlog-dev libopengl-dev libopus-dev libopusfile-dev libogg-dev libvorbis-dev
 
 # or using clang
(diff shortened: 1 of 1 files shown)


docs/BUILDING.md
```diff
@@ -89,34 +89,34 @@ C:\Program Files\CMake\bin\cmake.exe --build build-cmake --target ExtractAssetHe
...
 
 # or using clang
-pacman -S clang git cmake ninja lsb-release sdl2 libpng libzip nlohmann-json tinyxml2 spdlog sdl2_net
+pacman -S clang git cmake ninja lsb-release sdl2 libpng libzip nlohmann-json tinyxml2 spdlog sdl2_net libogg libvorbis opus opusfile
 ```
 #### Fedora
 ```sh
 # using gcc
-dnf install gcc gcc-c++ git cmake ninja-build lsb_release SDL2-devel libpng-devel libzip-devel libzip-tools nlohmann-json-devel tinyxml2-devel spdlog-devel
+dnf install gcc gcc-c++ git cmake ninja-build lsb_release SDL2-devel libpng-devel libzip-devel libzip-tools nlohmann-json-devel tinyxml2-devel spdlog-devel libogg-devel libvorbis-devel opus-devel opusfile-devel
```
Eblo, line 108: Is there a significant difference between `libogg-dev` and `libogg-devel`? What about `opusfile-devel` vs `libopusfile-dev`?

stellarkookies: the `-devel` suffix is used on fedora (and openSUSE?) to denote development packages, while the `-dev` suffix is used on Debian-based systems. as far as i know, there is no overlap between the two conventions on a given distribution (i.e. there is either `libogg-dev` or `libogg-devel`, but not both, depending on the distro family)

Eblo (approved):

PR #1723, opened 2026-06-01
[Enhancement] Port multiple warp points from SoH

Eblo:
image

- Multiple warp points with buttons to warp, set as boot point, and delete (Boot to Warp is no longer hidden)
- Added tooltip to Skip to File Select to clarify that it takes priority over boot to warp
- Added migration from singular warp point to new structure, with a default name of "Debug Warp Point"
- Set all nlohmann json conversions to `inline` because I couldn't get the `WarpPoint` struct conversion to compile otherwise

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7340448230.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7340227181.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7340032017.zip)

### DIFF
--- a/mm/2s2h/DeveloperTools/WarpPoint.cpp
+++ b/mm/2s2h/DeveloperTools/WarpPoint.cpp
@@ -1,9 +1,11 @@
 #include "2s2h/BenGui/UIWidgets.hpp"
+#include <ship/config/Config.h>
 #include <ship/window/gui/IconsFontAwesome4.h>
 #include "2s2h/GameInteractor/GameInteractor.h"
 #include "2s2h/DeveloperTools/DeveloperTools.h"
 #include "2s2h/ShipInit.hpp"
 #include "2s2h/BenGui/BenGui.hpp"
+#include "BenJsonConversions.hpp"
 
 extern "C" {
 #include "z64.h"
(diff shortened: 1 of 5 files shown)


mm/2s2h/DeveloperTools/WarpPoint.cpp
```diff
@@ -64,105 +73,129 @@ void Warp() {
...
+            ImGui::BeginDisabled();
+        }
+
+        if (UIWidgets::Button(ICON_FA_PLUS)) {
+            Player* player = GET_PLAYER(gPlayState);
+
+            std::string warpName = Ship_GetSceneName(gPlayState->sceneId);
+            if (gPlayState->roomCtx.curRoom.num != 0) {
+                warpName += " (" + std::to_string(gPlayState->roomCtx.curRoom.num) + ")";
+            }
```
mckinlee: warpName is computed here but I don't see where its used. Is this dead code?

mckinlee (approved):
lgtm with one question

mm/2s2h/DeveloperTools/WarpPoint.cpp
Eblo: Good catch. This could be used as a default name for the warp point. I'll make that change.

PR #1726, opened 2026-06-04
Fix GC US assets conflicting with N64 US HD assets

### LINKED ISSUE #610: GC Texture Differences Need Different Names
Textures that differ slightly (format/height/width/language) between the N64 and GC release need different names so that texture packs overriding these, do not cause issues with wrapping/format/overriding the wrong language

GhostlyDark:
Turn the copyright text (different width) and some of the File Select text (different texture format: ia4 vs. ia8) into uniquely named texture files. This is required to make the same HD texture pack work for both ROMs without exhibiting issues. Fixes #610.

_before (GC US)_
image
image

_after (GC US with updated HD pack)_
image
image

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7414355101.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7414084727.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7413869432.zip)

### DIFF
--- a/mm/assets/misc/title_static/title_static.h
+++ b/mm/assets/misc/title_static/title_static.h
@@ -6,17 +6,32 @@
 #define dgFileSelNoFileToCopyENGTex "__OTR__misc/title_static/gFileSelNoFileToCopyENGTex"
 static const ALIGN_ASSET(2) char gFileSelNoFileToCopyENGTex[] = dgFileSelNoFileToCopyENGTex;
 
+#define dgFileSelNoFileToCopyIA4ENGTex "__OTR__misc/title_static/gFileSelNoFileToCopyIA4ENGTex"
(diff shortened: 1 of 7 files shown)


smartgamer701:
I tried your build artifact on Windows and this is what I see on the options screen with MM_Reloaded_v11.0.2_HD.o2r
image
This is the title screen
image
I tried the same exact thing on mac, and same results
I could be doing something wrong, I don't know about textures and stuff

GhostlyDark:
The last texture pack release doesn't contain the new textures necessary to show them in HD. The textures looking alright but SD with v11.0.2 means the PR is working - the textures aren't corrupted, which is what the fix is about.

smartgamer701:
Oh, so the texture pack has to be updated to make it look like your screenshots? Ok, thank you.

Eblo (approved):

Issue #1580, opened 2026-02-27
4.1.0 Release


garrettjoecox:
Anyone with access is free to edit this list.

## Must have
- [ ] Changelog
- [ ] Update menu credits as needed
- [ ] Inconsistent softlock happening when opening grave in graveyard (potentially fixed with #1602 
- [ ] Separate toggle for glitch useful cutscenes (Moon crash specifically)
- [ ] #1504 (need to make sure we dont break texture packs)
- [ ] Static traps not actually static

## Should have
- [ ] Modifying starting health should affect how many pieces are added to pool
- [ ] Option to disable infinite purchasing of rupees in shops
- [ ] Replace remaining TextDrawNotification usage with notifications (save manager mostly)
- [ ] Hint text on stray fairy return message (So they dont have to leave and re-enter for hint)
- [ ] Default final hours start time to 30 seconds till moon crash
- [x] #1594 
- [x] #1603 
- [ ] Randomizer option to have tingle share maps like vanilla
- [ ] The gossip stones on the moon should hint transformation mask locations
- [x] #1595
- [ ] Better handling of moving scrubs and multiple scrubs being in the same location
- [ ] Better hinting on scrubs
- [ ] When using FD anywhere, allow using transformation mask
- [ ] Disable faster song playback if we detect you are trying song storage (check -y velocity?)
- [ ] Re-introduce save-delay in more scenarios (when voiding out outside snowhead)
- [ ] Disable moon crash cutscene skip if user is attempting 4th day glitch
- [ ] Option to remove trade items from inventory when used until song of time reset
- [x] #1592 
- [ ] #1628
- [ ] Disable one point cutscene skip on left-most switch outside stone tower temple
- [ ] More granular NPC hint options (to toggle them)
- [ ] If shops not shuffled, dont place bomb bag vanilla (and bigger bomb bag?)
- [ ] Hinting your starting items shouldnt be possible
- [ ] Disable explosive limit enhancement
- [x] #1600
- [x] #1597
- [ ] Swamp archery speedup should apply without custom score (misc interactions probably)
- [ ] #1329
- [ ] #1564 
- [x] #1567 
- [ ] #1538 (someone take this from pat)
- [ ] #1531 (someone take this from pat)
- [ ] #1496 (needs thorough QA for features it touches)
- [ ] #1484 (Needs linux debugging)
- [ ] #1588 

## Nice to have
- [ ] Giant’s mask anywhere
[...]

PR #1495, opened 2026-01-23
swap forms from fd mask

Jameriquiah:
extension to #1455
realised i forgot to allow fd to use other transformation masks, this is a feature in mm3d so im just adding it to this enhancement. initially caused a softlock when switching from fd, and then to human, so i had to rewrite some stuff to fix that. also renamed the cpp bc i forgot to when i first did the name change on the enhancement so the cpp now matches what its actually for

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5425534510.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5425575803.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/5425631724.zip)

PR #1734, opened 2026-06-12
Fierce Deity can use transformation masks

JordanLongstaff:
While in Fierce Deity form, allows other transformation masks to be used. However, they appear _huge_ on FD Link's head.

FD to Deku
FD to Goron
FD to Zora

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7599198152.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7599026897.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7598797010.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
+++ b/mm/2s2h/Enhancements/Masks/FierceDeityAnywhere.cpp
@@ -234,6 +234,15 @@ static void RegisterFierceDeityAnywhere() {
         ItemId* itemId = va_arg(args, ItemId*);
         AllowTransformationMask(itemId, should);
     });
+
+    /*
+     * Needed to allow FD Link to use transformation masks while in water.
+     */
+    COND_VB_SHOULD(VB_FD_ALWAYS_WIELD_SWORD, CVAR, {
+        if (Player_GetEnvironmentalHazard(gPlayState) >= PLAYER_ENV_HAZARD_UNDERWATER_FLOOR) {
+            *should = false;
+        }
+    });
 }
 
 static RegisterShipInitFunc initFunc(RegisterFierceDeityAnywhere, { CVAR_NAME });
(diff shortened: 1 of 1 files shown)


Jameriquiah:
oh i already did this in #1495

JordanLongstaff:
Your way requires 3DS Mask Equipping. Mine doesn't.
My way connects this feature to Fierce Deity Anywhere, which maybe isn't optimal either, but it fulfills the task listed in #1580.

Jameriquiah:
oh so its always on when fd anywhere is turned on?

JordanLongstaff:
Yes.

grande1900:
very nice! does this also allow you to equip zora mask underwater? That was a big painpoint when trying to do certain skips with FD.

Eblo (approved):
> Your way requires 3DS Mask Equipping. Mine doesn't. My way connects this feature to Fierce Deity Anywhere, which maybe isn't optimal either, but it fulfills the task listed in #1580.

Yeah, there may be some overlap, but I think that's fine. It does make sense to add this as part of FDA.

> very nice! does this also allow you to equip zora mask underwater? That was a big painpoint when trying to do certain skips with FD.

#1268 addressed this by allowing the FD mask itself to be usable in water, though this new PR allows the player to go straight from FD to Zora without needing to go to human first.

JordanLongstaff:
> > very nice! does this also allow you to equip zora mask underwater? That was a big painpoint when trying to do certain skips with FD.
> 
> #1268 addressed this by allowing the FD mask itself to be usable in water, though this new PR allows the player to go straight from FD to Zora without needing to go to human first.

Actually, unfortunately, if you are FD, you can't use any of the transformation masks underwater ATM. Deku and Goron are understandable and appear disabled, but Zora and FD look usable but aren't. Seriously - you can't take FD off underwater.

I'm going to investigate this, but I think it can be fixed in a separate PR.

JordanLongstaff:
Actually, never mind, I'll add it into this PR.

JordanLongstaff:
There, done. I had to add an extra hook that disables the sword in water, because otherwise, the game refused to recognize the D-Pad inputs.

Side note: I see that the FD mask is disabled while walking underwater as Zora Link. This seems weird to me. Why can't we just allow it? It'd be a very easy change.

JordanLongstaff:
Fixes #1736

Eblo (approved):

PR #1735, opened 2026-06-12
Add Boss Health Multiplier option and fix Z-Targeting health bar overflow

MarvinCoto12:
## Description ## 

This pull request introduces a new option in the difficulty settings: a **Boss Health Multiplier** dropdown menu with precise increments (1x, 1.25x, 1.50x, 1.75x, 2x). This multiplier adjusts the health of the game's four main bosses and all phases of the final battle against Majora.

It also fixes a rendering issue where, when multiplying an enemy's health, the red fill bar of the Z-Targeting system would overflow its white box.

## Changes Made

* **`BenMenu.cpp`**: Added a checkbox to a dropdown menu (combobox) for easy selection of different difficulty options.

* **`BossHealthMultiplier.cpp`**: Created a new file that handles boss health. It utilizes ID-specific hooks (`COND_ID_HOOK`) to ensure zero performance overhead on regular actors. Memory lifecycle is safely handled via the `OnActorDestroy` hook, clearing the tracking map automatically whenever a boss unloads to prevent stale data or accumulation.

* **`EnemyHealthBars.cpp`**: The UI renderer has been updated to automatically increase the maximum health if the boss's current health exceeds it. This prevents the red bar from extending beyond its white border.

## Tests Performed

* **Odolwa**, **Goht**, **Gyorg**, and **Twinmold** were tested via the debug mode. It was verified that the health scaling is applied correctly.

* **Goht** was fought multiple times; it was confirmed that the multiplier activates correctly after the thawing cutscene and does not exhibit memory leaks or accumulate indefinitely upon the player's death/scene reload thanks to the destroy hooks.

* **Majora** was fought in all three phases; it was confirmed that Majora's Rage (Phase 3) inherits the correct health bar.

* It was verified that, with Z-Targeting enabled, the red health bar empties smoothly and remains completely within the white box.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7771721694.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7771687469.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7771597987.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1711,7 +1711,7 @@ void BenMenu::AddEnhancements() {
         .CVar("gEnhancements.DifficultyOptions.HyperEnemies")
         .Options(CheckboxOptions().Tooltip("Double the rate at which enemies are updated, making them more difficult"));
     static const std::unordered_map<int32_t, const char*> bossHealthOptions = {
-        { 1, "1.0x (Default)" }, { 2, "1.25x" }, { 3, "1.50x" }, { 4, "1.75x" }, { 5, "2.0x" }
+        { 1, "1x (Default)" }, { 2, "1.25x" }, { 3, "1.50x" }, { 4, "1.75x" }, { 5, "2x" }
     };
     AddWidget(path, "Boss Health Multiplier", WIDGET_CVAR_COMBOBOX)
         .CVar("gEnhancements.DifficultyOptions.BossHealthMultiplier")
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1703,13 +1703,21 @@ void BenMenu::AddEnhancements() {
                      .Max(40)
                      .DefaultValue(0));
 
-    // Difficulty Options
+   // Difficulty Options
     path = { "Enhancements", "Difficulty Options", SECTION_COLUMN_1 };
     AddSidebarEntry("Enhancements", "Difficulty Options", 3);
     AddWidget(path, "Combat", WIDGET_SEPARATOR_TEXT);
     AddWidget(path, "Hyper Enemies", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.DifficultyOptions.HyperEnemies")
         .Options(CheckboxOptions().Tooltip("Double the rate at which enemies are updated, making them more difficult"));
+    static const std::unordered_map<int32_t, const char*> bossHealthOptions = { { 1, "1.0x (Default)" },
+                                                                                { 2, "1.5x" },
+                                                                                { 3, "2.0x" } };
+    AddWidget(path, "Boss Health Multiplier", WIDGET_CVAR_COMBOBOX)
+        .CVar("gEnhancements.DifficultyOptions.BossHealthMultiplier")
+        .Options(ComboboxOptions()
+                     .Tooltip("Multiply the health of all bosses (Restart Required).")
+                     .ComboMap(&bossHealthOptions));
     AddWidget(path, "Damage Multiplier", WIDGET_CVAR_COMBOBOX)
         .CVar("gEnhancements.DifficultyOptions.DamageMultiplier")
         .Options(ComboboxOptions()
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1716,7 +1716,7 @@ void BenMenu::AddEnhancements() {
     AddWidget(path, "Boss Health Multiplier", WIDGET_CVAR_COMBOBOX)
         .CVar("gEnhancements.DifficultyOptions.BossHealthMultiplier")
         .Options(ComboboxOptions()
-                     .Tooltip("Multiply the health of all bosses (Restart Required).")
+                     .Tooltip("Multiply the health of all bosses. Requires a Scene Reload to take effect.")
                      .ComboMap(&bossHealthOptions));
     AddWidget(path, "Damage Multiplier", WIDGET_CVAR_COMBOBOX)
         .CVar("gEnhancements.DifficultyOptions.DamageMultiplier")
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1703,7 +1703,7 @@ void BenMenu::AddEnhancements() {
                      .Max(40)
                      .DefaultValue(0));
 
-   // Difficulty Options
+    // Difficulty Options
     path = { "Enhancements", "Difficulty Options", SECTION_COLUMN_1 };
     AddSidebarEntry("Enhancements", "Difficulty Options", 3);
     AddWidget(path, "Combat", WIDGET_SEPARATOR_TEXT);
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -61,6 +61,10 @@ static const std::vector<const char*> alwaysWinDoggyraceOptions = {
     "Always",                    // ALWAYS_WIN_DOGGY_RACE_ALWAYS
 };
 
+static const std::unordered_map<int32_t, const char*> bossHealthOptions = {
+    { 0, "1x (Default)" }, { 1, "1.25x" }, { 2, "1.50x" }, { 3, "1.75x" }, { 4, "2x" }
+};
+
 static const std::vector<const char*> cremiaRewardOptions = {
     "Vanilla", // CREMIA_REWARD_RANDOM
     "Hug",     // CREMIA_REWARD_ALWAYS_HUG
@@ -1710,9 +1714,6 @@ void BenMenu::AddEnhancements() {
     AddWidget(path, "Hyper Enemies", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.DifficultyOptions.HyperEnemies")
         .Options(CheckboxOptions().Tooltip("Double the rate at which enemies are updated, making them more difficult"));
-    static const std::unordered_map<int32_t, const char*> bossHealthOptions = {
-        { 1, "1x (Default)" }, { 2, "1.25x" }, { 3, "1.50x" }, { 4, "1.75x" }, { 5, "2x" }
-    };
     AddWidget(path, "Boss Health Multiplier", WIDGET_CVAR_COMBOBOX)
         .CVar("gEnhancements.DifficultyOptions.BossHealthMultiplier")
         .Options(ComboboxOptions()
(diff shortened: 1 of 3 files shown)


mm/2s2h/Enhancements/DifficultyOptions/BossHealthMultiplier.cpp
```diff
@@ -0,0 +1,58 @@
...
+#include "functions.h"
+}
+
+#define CVAR_NAME "gEnhancements.DifficultyOptions.BossHealthMultiplier"
+
+static std::unordered_map<Actor*, int> sLastBossHealth;
+static uint32_t sLastFrames = 0;
+
+void RegisterBossHealthMultiplier() {
+    COND_HOOK(OnActorUpdate, 1, [](Actor* actor) {
```
Eblo: The way you have this set up with the `1` means this will always run, regardless of the CVar's value. Ideally, the "off" setting (1 in this case) should avoid running the code:
```suggestion
#define CVAR_NAME "gEnhancements.DifficultyOptions.BossHealthMultiplier"
#define CVAR CVarGetInteger(CVAR_NAME, 1)

static std::unordered_map<Actor*, int> sLastBossHealth;
static uint32_t sLastFrames = 0;

void RegisterBossHealthMultiplier() {
    COND_HOOK(OnActorUpdate, CVAR > 1, [](Actor* actor) {
```

mm/2s2h/Enhancements/DifficultyOptions/BossHealthMultiplier.cpp
```diff
@@ -0,0 +1,58 @@
...
+        if (actor->id == ACTOR_BOSS_01 || actor->id == ACTOR_BOSS_HAKUGIN || actor->id == ACTOR_BOSS_03 ||
+            actor->id == ACTOR_BOSS_02 || actor->id == ACTOR_BOSS_07) {
```
Eblo: You have an `OnActorUpdate` that effectively runs for all actors, which can be cumbersome for something that only needs to run for five actors. My recommendation here is to put lines 18-55 into their own function (e.g. `HandleBossHealthMultiplier` and change the `COND_HOOK` to a `COND_ID_HOOK` that calls that function for each of the five actor IDs here.

mm/2s2h/Enhancements/DifficultyOptions/BossHealthMultiplier.cpp
```diff
@@ -0,0 +1,58 @@
...
+                sLastFrames = gPlayState->gameplayFrames;
+            }
+
+            int currentHealth = actor->colChkInfo.health;
+
+            if (currentHealth > 0) {
+                // Every time a boss phases or unfreezes (e.g. Goth, Majora's transformations), the engine overwrites
+                // its health with hardcoded values, so we just watch for health surges and override it again.
+                if (sLastBossHealth.count(actor) == 0 || currentHealth > sLastBossHealth[actor]) {
+                    int option = CVarGetInteger(CVAR_NAME, 1);
```
Eblo:
```suggestion
                    int option = CVAR;
```

mm/2s2h/Enhancements/DifficultyOptions/BossHealthMultiplier.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+                    if (option == 2)
+                        multiplier = 1.5f;
+                    if (option == 3)
+                        multiplier = 2.0f;
+
+                    if (multiplier > 1.0f) {
+                        currentHealth = (s8)(currentHealth * multiplier);
+                        actor->colChkInfo.health = currentHealth;
+                    }
```
Eblo: Per my `CVAR > 1` suggestion, this condition would no longer be needed.
```suggestion
                    currentHealth = (s8)(currentHealth * multiplier);
                    actor->colChkInfo.health = currentHealth;
```

mm/2s2h/BenGui/BenMenu.cpp
```diff
@@ -1710,6 +1710,14 @@ void BenMenu::AddEnhancements() {
...
         .CVar("gEnhancements.DifficultyOptions.HyperEnemies")
         .Options(CheckboxOptions().Tooltip("Double the rate at which enemies are updated, making them more difficult"));
+    static const std::unordered_map<int32_t, const char*> bossHealthOptions = { { 1, "1.0x (Default)" },
+                                                                                { 2, "1.5x" },
+                                                                                { 3, "2.0x" } };
+    AddWidget(path, "Boss Health Multiplier", WIDGET_CVAR_COMBOBOX)
+        .CVar("gEnhancements.DifficultyOptions.BossHealthMultiplier")
+        .Options(ComboboxOptions()
+                     .Tooltip("Multiply the health of all bosses (Restart Required).")
+                     .ComboMap(&bossHealthOptions));
```
Eblo: I wonder if this should be a combobox, or if a float slider would make more sense. What are your thoughts?

mm/2s2h/BenGui/BenMenu.cpp
```diff
@@ -1710,6 +1710,14 @@ void BenMenu::AddEnhancements() {
     AddWidget(path, "Hyper Enemies", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.DifficultyOptions.HyperEnemies")
         .Options(CheckboxOptions().Tooltip("Double the rate at which enemies are updated, making them more difficult"));
+    static const std::unordered_map<int32_t, const char*> bossHealthOptions = { { 1, "1.0x (Default)" },
+                                                                                { 2, "1.5x" },
+                                                                                { 3, "2.0x" } };
+    AddWidget(path, "Boss Health Multiplier", WIDGET_CVAR_COMBOBOX)
+        .CVar("gEnhancements.DifficultyOptions.BossHealthMultiplier")
+        .Options(ComboboxOptions()
+                     .Tooltip("Multiply the health of all bosses (Restart Required).")
```
Eblo:
```suggestion
                     .Tooltip("Multiply the health of all bosses. Requires a Scene Reload to take effect.")
```

Eblo (changes requested):

mm/2s2h/BenGui/BenMenu.cpp
MarvinCoto12: I think that a float slider would serve to be able to put even more options (like 1.25x etc) and keep the space more visually organized, it seems like a good idea to me.

edit: I tried implementing and testing the float slider you mentioned, but the boss health multiplication wasn't working correctly, and I couldn't have fixed, direct options like (1.0x, 1.25x, 1.50x, 1.75x, 2.0x), which I can put in a combo box. This is to ensure a completely accurate and predictable user experience for everyone.

mm/2s2h/Enhancements/DifficultyOptions/BossHealthMultiplier.cpp
MarvinCoto12: You're right, I'll try to implement it.

mm/2s2h/Enhancements/DifficultyOptions/BossHealthMultiplier.cpp
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    // If the gameplay frames reset or go backwards, it means the player reloaded the scene,
+    // died, or started a new fight. We wipe the map clean to avoid stale health data carrying over.
+    if (gPlayState != NULL) {
+        if (gPlayState->gameplayFrames < sLastFrames || gPlayState->gameplayFrames == 0) {
+            sLastBossHealth.clear();
+        }
+        sLastFrames = gPlayState->gameplayFrames;
+    }
```
Eblo: Instead of running this check every frame, you could just use the `OnActorDestroy` hook to clear the value when the actor of interest gets unloaded.

mm/2s2h/Enhancements/DifficultyOptions/BossHealthMultiplier.cpp
```diff
@@ -0,0 +1,77 @@
...
+                case 4:
+                    multiplier = 1.75f;
+                    break;
+                case 5:
+                    multiplier = 2.0f;
+                    break;
+                default:
+                    multiplier = 1.0f;
+                    break;
+            }
```
Eblo: With this many paired options, it would be cleaner to define a map or array and simply retrieve the multiplier based on the CVar value.

mm/2s2h/Enhancements/DifficultyOptions/BossHealthMultiplier.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+extern "C" {
+#include "variables.h"
+#include "functions.h"
+}
+
+#define CVAR_NAME "gEnhancements.DifficultyOptions.BossHealthMultiplier"
+#define CVAR CVarGetInteger(CVAR_NAME, 1)
+
+static std::unordered_map<Actor*, int> sLastBossHealth;
```
Eblo, line 14: I'm wondering if ObjectExtension would be better suited for this.

mm/2s2h/BenGui/BenMenu.cpp
```diff
@@ -1710,6 +1710,14 @@ void BenMenu::AddEnhancements() {
     AddWidget(path, "Hyper Enemies", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.DifficultyOptions.HyperEnemies")
         .Options(CheckboxOptions().Tooltip("Double the rate at which enemies are updated, making them more difficult"));
+    static const std::unordered_map<int32_t, const char*> bossHealthOptions = {
+        { 1, "1x (Default)" }, { 2, "1.25x" }, { 3, "1.50x" }, { 4, "1.75x" }, { 5, "2x" }
+    };
```
Eblo: Please move this to the top of the file with the other combobox options (`alwaysWinDoggyraceOptions`, `cremiaRewardOptions`, etc.). I also think it makes more sense to zero-index these options.

mm/2s2h/Enhancements/DifficultyOptions/BossHealthMultiplier.cpp
```diff
@@ -0,0 +1,77 @@
...
+    int currentHealth = actor->colChkInfo.health;
+
+    if (currentHealth > 0) {
+        // Every time a boss phases or unfreezes (e.g. Goth, Majora's transformations), the engine overwrites
+        // its health with hardcoded values, so we just watch for health surges and override it again.
+        if (sLastBossHealth.count(actor) == 0 || currentHealth > sLastBossHealth[actor]) {
+            int option = CVAR;
+            float multiplier = 1.0f;
+
+            switch (option) {
```
Eblo:
```suggestion
            float multiplier = 1.0f;

            switch (CVAR) {
```

Eblo (changes requested):

CallaboDev:
@Eblo Hi, I've just implemented the changes you suggested. I'm really enjoying contributing to this project. If you see any other changes that need to be made, please let me know.

Eblo (approved):

Issue #1741, opened 2026-06-18
[Feature Request] Allow unbinding masks while wearing them


ShawnSigma:
By default the game will prevent you from unbinding a mask that you're already wearing from your C buttons or d-pad and play a sound to indicate that. This also happens when you try to bind something else in its place and the other buttons are already bound. I think it would be nice to get rid of this as it's just kind of annoying to have to take off a mask manually every time you want to use a different item

ShawnSigma:
I had another idea. How about an option to make it so masks can be equipped directly from the inventory screen by pressing A without having to bind them first?

CallaboDev:
I prefer the first idea, and yes, it would be a good option to be able to equip another item in the slot where you have the mask equipped, even while wearing it, and for Link to detransform when equipping something else.

Eblo:
I'd say #1744 addresses this concern.

PR #755, opened 2024-08-11
[Enhancement] EasyMaskEquip

mckinlee:
EasyMaskEquip is a feature that allows players to quickly equip masks without needing to assign them to C-buttons or D-pad slots. It simplifies the process by letting players equip masks directly from the pause menu. The enhancement also provides visual feedback by highlighting the currently equipped mask and graying out masks that can't be equipped in certain situations (e.g., underwater or mid-transformation).

~~This PR introduces the EasyMaskEquip enhancement, allowing players to equip masks directly from the pause menu by pressing the A button.~~

A new OnKaleidoClose hook was added to handle mask equipping when the pause menu is closed. This allows players to change their mask selection if they accidentally select the wrong one, ensuring the mask is only equipped once the pause menu is closed.

~~This enhancement also utilizes Proxy's BeforeKaleidoDrawPage hook from PR #537 to allocate and update vertex data for the mask equip border.~~

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2878890813.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2878926775.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/2879020032.zip)

PR #1744, opened 2026-06-19
[Enhancement] Easy Mask Equip

mckinlee:
Pressing A on a mask now equips or unequips it directly from the pause menu without needing to assign it to a button first.

This covers regular masks, transformation masks, Giant/Fierce Deity restrictions, active mask highlighting, and compatibility with Persistent Bunny Hood / Fast Transformation.

This supersedes my older transformation mask only PR #755. That version was from when I was less familiar with this part of the codebase and took a more custom, narrower approach. This version uses the existing player item action flow more directly, covers the full mask page behavior, and handles more edge cases, so I’ll be closing the older PR in favor of this one. This would also resolve issue #1741.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7884443338.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7884127783.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7884010676.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1400,7 +1400,7 @@ void BenMenu::AddEnhancements() {
     AddWidget(path, "Easy Mask Equip", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.Masks.EasyMaskEquip")
         .Options(CheckboxOptions().Tooltip("In the pause menu, press A on any owned mask to put it on or take it off "
-                                           "without assigning it to a C-button. Mask-specific restrictions still "
+                                           "without assigning it to a button. Mask-specific restrictions still "
                                            "apply."));
     AddWidget(path, "Fierce Deity's Mask Anywhere", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.Masks.FierceDeitysAnywhere")
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -1397,6 +1397,11 @@ void BenMenu::AddEnhancements() {
         .CVar("gEnhancements.Masks.3DSMaskEquip")
         .Options(CheckboxOptions().Tooltip("Allows equipping masks while in other forms, returning you to human form "
                                            "with the mask immediately equipped, like in MM3D."));
+    AddWidget(path, "Easy Mask Equip", WIDGET_CVAR_CHECKBOX)
+        .CVar("gEnhancements.Masks.EasyMaskEquip")
+        .Options(CheckboxOptions().Tooltip("In the pause menu, press A on any owned mask to put it on or take it off "
+                                           "without assigning it to a C-button. Mask-specific restrictions still "
+                                           "apply."));
     AddWidget(path, "Fierce Deity's Mask Anywhere", WIDGET_CVAR_CHECKBOX)
         .CVar("gEnhancements.Masks.FierceDeitysAnywhere")
         .Options(CheckboxOptions().Tooltip("Allow using Fierce Deity's mask outside of boss rooms."));
--- a/mm/2s2h/Enhancements/Masks/EasyMaskEquip.cpp
+++ b/mm/2s2h/Enhancements/Masks/EasyMaskEquip.cpp
@@ -1,14 +1,9 @@
-#include <libultraship/bridge/consolevariablebridge.h>
-#include "2s2h/GameInteractor/GameInteractor.h"
 #include "2s2h/Enhancements/FrameInterpolation/FrameInterpolation.h"
 #include "2s2h/ShipInit.hpp"
 
 extern "C" {
 #include "assets/interface/parameter_static/parameter_static.h"
 #include "src/overlays/kaleido_scope/ovl_kaleido_scope/z_kaleido_scope.h"
-#include "variables.h"
-#include "z64interface.h"
-#include "z64player.h"
 
 void Player_UseItem(PlayState* play, Player* player, ItemId item);
 void func_8082E1F0(Player* player, u16 sfxId);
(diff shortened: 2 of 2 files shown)


mm/2s2h/BenGui/BenMenu.cpp
```diff
@@ -1397,6 +1397,11 @@ void BenMenu::AddEnhancements() {
         .CVar("gEnhancements.Masks.3DSMaskEquip")
         .Options(CheckboxOptions().Tooltip("Allows equipping masks while in other forms, returning you to human form "
                                            "with the mask immediately equipped, like in MM3D."));
+    AddWidget(path, "Easy Mask Equip", WIDGET_CVAR_CHECKBOX)
+        .CVar("gEnhancements.Masks.EasyMaskEquip")
+        .Options(CheckboxOptions().Tooltip("In the pause menu, press A on any owned mask to put it on or take it off "
+                                           "without assigning it to a C-button. Mask-specific restrictions still "
```
Eblo: cuz dpad exists
```suggestion
                                           "without assigning it to a button. Mask-specific restrictions still "
```

mm/2s2h/Enhancements/Masks/EasyMaskEquip.cpp
```diff
@@ -0,0 +1,717 @@
...
+        RestoreAButtonPrompt();
+        RestoreButtonGhost();
+        CancelPendingAction(player);
+        sButtonGhost.mask = PLAYER_MASK_NONE;
+    }
+
+    COND_HOOK(OnKaleidoUpdate, CVAR, OnKaleidoUpdate);
+    COND_ID_HOOK(BeforeKaleidoDrawPage, PAUSE_MASK, CVAR, OnBeforeKaleidoDrawPage);
+    COND_ID_HOOK(ShouldActorUpdate, ACTOR_PLAYER, CVAR, OnShouldPlayerUpdate);
+    COND_ID_HOOK(OnActorUpdate, ACTOR_PLAYER, CVAR, OnPlayerUpdate);
```
Eblo: These function names are similar (or in one case, identical) to the hook names. They should be renamed to be more descriptive of what they do for this enhancement.

mm/2s2h/Enhancements/Masks/EasyMaskEquip.cpp
```diff
@@ -0,0 +1,717 @@
...
+
+    gDPSetPrimColor(POLY_OPA_DISP++, 0, 0, 80, 255, 160, pauseCtx->alpha);
+    gSPVertex(POLY_OPA_DISP++, (uintptr_t)activeMaskVtx, 4, 0);
+    POLY_OPA_DISP = Gfx_DrawTexQuadIA8(POLY_OPA_DISP, (TexturePtr)gEquippedItemOutlineTex, 32, 32, 0);
+
+    CLOSE_DISPS(gfxCtx);
+}
+
+static void OnBeforeKaleidoDrawPage(PauseContext* pauseCtx, u16 pauseIndex) {
+    (void)pauseIndex;
```
Eblo: What does this do?

mm/2s2h/Enhancements/Masks/EasyMaskEquip.cpp
```diff
@@ -0,0 +1,717 @@
...
+
+static bool PlayerCanStartFollowUpAction(Player* player) {
+    return (player != nullptr) && (gPlayState->pauseCtx.state == PAUSE_STATE_OFF) &&
+           (player->transformation == PLAYER_FORM_HUMAN) && (player->currentMask == PLAYER_MASK_NONE) &&
+           (player->unk_AA5 == PLAYER_UNKAA5_0) &&
+           !(player->stateFlags1 & (PLAYER_STATE1_2 | PLAYER_STATE1_100 | PLAYER_STATE1_20000000));
+}
+
+static bool QueuePlayerAction(Player* player, ItemId item, PlayerMask selectedMask, PlayerItemAction itemAction,
+                              PlayerMask targetMask, PendingMode pendingMode, bool playMenuSfx = true);
```
Eblo: Is this forward declaration necessary?

mm/2s2h/Enhancements/Masks/EasyMaskEquip.cpp
```diff
@@ -0,0 +1,717 @@
...
+            break;
+    }
+
+    if ((sButtonGhost.mask != PLAYER_MASK_NONE) && (player->currentMask != sButtonGhost.mask)) {
+        sButtonGhost.mask = PLAYER_MASK_NONE;
+    }
+}
+
+static void OnGetItemOnButton(bool* should, va_list args) {
+    (void)should;
```
Eblo: Ditto.

mm/2s2h/Enhancements/Masks/EasyMaskEquip.cpp
```diff
@@ -0,0 +1,717 @@
...
+            queued = QueueReturnToHumanThenRegular(player, item, targetMask, itemAction);
+        }
+
+        if (!queued) {
+            ResetPendingAction();
+        }
+        return;
+    }
+
+    if ((targetMask == PLAYER_MASK_GIANT) && (player->transformation != PLAYER_FORM_HUMAN)) {
```
Eblo, line 744: There need to be additional conditions for processing transformation masks. If the player tries to (un)equip a transformation mas from the menu while in a state where transformation masks cannot be used (e.g. while jumping), the transformation will not occur, but the EME item state will reflect as if it did.

Eblo (changes requested):
I like how refined this is over your previous iteration. Applying to all masks will definitely be helpful for those one-off uses, particularly in rando.

mm/2s2h/Enhancements/Masks/EasyMaskEquip.cpp
mckinlee: It's a unused param, but we don't need pauseIndex since the hook already filters to the mask page. I'll make that clearer by just leaving the unused param unnamed instead.

mm/2s2h/Enhancements/Masks/EasyMaskEquip.cpp
mckinlee: Nope, I'll just move the function somewhere before its used and get rid of that.

mm/2s2h/Enhancements/Masks/EasyMaskEquip.cpp
mckinlee: I'll do the same thing here.

Eblo (approved):

PR #1745, opened 2026-06-20
Fix silent SFX and several 64-bit-host audio crashes (non-PIE builds)

billsix:
These fixes come from getting audio working on a native Linux build (Fedora 44, gcc-16). Two things to flag up front for reviewers:

- The **unfixed code builds and runs in CI (Ubuntu 22.04, gcc-12) with working audio.** That initially looked like a compiler/toolchain problem. It is not — none of these changes depend on the compiler version. They are 64-bit-host portability bugs, plus one latent bug that CI happens to mask via memory layout (explained below). Several are present in `develop` for *any* 64-bit build.
- We've tried to be explicit about what we **verified** vs. what is a **stopgap** or an **inspection-only** finding, and to defer to maintainers where the proper fix is in an area we're less familiar with (the asset pipeline). Where we're confident, we say so; where we're not, we say that too.

`AudioPlayback_ProcessNotes` had a port guard:

    if ((uintptr_t)playbackState->parentLayer < 0x7FFFFFFF) { continue; }

Its intent — skip layer pointers "below where N64 RAM would be" (valid N64 pointers are >= 0x80000000) — is reasonable on the console. On a 64-bit host it misfires in a **layout-dependent** way: `SequenceLayer`s are sub-allocated from `gAudioHeap`, a static array, so a PIE/ASLR build places them above 2 GB (guard harmless), but a non-PIE build places them below (~0x02dc9268 on the build where we saw this) — and the guard then skips **every note's playback update**. The ADSR envelope never advances past INITIAL, note volume stays 0, and sound effects play at zero amplitude. Removing the guard restored SFX; verified by ear and by tracing that the envelope now advances to its target. OoT/SoH has no equivalent check, and NO_LAYER is already handled just above.

This is also why CI "works": a PIE build keeps the audio heap above 2 GB and never trips the guard. Same latent bug, masked by layout — not a compiler issue.

Honest caveats: (a) if this guard was protecting against something we couldn't reproduce, a narrower validity check may be preferable to outright removal; (b) we did **not** fully characterize why music was less affected than SFX under the bug — the fix is verified regardless, but that asymmetry is an open question.

`osRecvMesg` copies a full `OSMesg` (8 bytes on a 64-bit host — it holds a `void*`). Several call sites received into a 4-byte `u32`, overflowing the stack; `AudioLoad_ProcessScriptLoads` was an observed crash. Fixed by receiving into an `OSMesg` and reading `.data32`. (Maintainers: other `osRecvMesg` sites may share this narrower-than-`OSMesg` pattern — a sweep would be worthwhile.)

- `sched.c`: read scheduler messages via the 64-bit `.ptr` union member instead of `.data32`, which truncates the `OSScMsg` pointer on a 64-bit host.
- `z_scene.c`: widen a truncated object segment address from `u32` to `uintptr_t`.

We found these by inspection rather than a reproduced crash, but the truncation is unambiguous on a 64-bit host.

- `synthesis.c`: clamp `numSamplesUntilEnd` to `>= 0`. We saw crashes where a reused note carried a stale `samplePosInt` past a new (shorter) sample's end, making this negative and cascading into a software-RSP DMEM overflow (seen under ASan). The clamp prevents that, but it is a **guard, not a fix**: our best guess is `samplePosInt` should be reset on note reuse, and it is possible this was a downstream symptom that no longer triggers after fix #1 — we have not re-verified. (An in-code comment also records a tempting *wrong* fix to avoid.)
- `AudioSampleFactory.cpp` + `AudioSample.h`: initialize `AdpcmLoop.sampleEnd` (offset 0x0C). The `.o2r`/ZAPD don't carry this field because OoT/sm64's audio engines never read it — it's MM-specific (only the `count==2` "loop N then stop" path uses it). For MM it was left uninitialized and crashed `count==2` samples (read as poison -> out-of-bounds). As a **format-compatible stopgap** we set it to the loop end (exact for one-shots, possibly short for `count==2` with a tail). The correct fix — extracting the real value from the ROM (ZAPD + OTRExporter + a resource-version bump, then re-extract) — belongs to the asset pipeline; we defer to maintainers on the format/version conventions.

- Sweep remaining `osRecvMesg` call sites for the 8-vs-4-byte receive-buffer bug.
- Find the real cause of the `samplePosInt` overshoot (reset on reuse?) and determine whether the clamp is still needed after fix #1.
- Extract the real `AdpcmLoop.sampleEnd` through the asset pipeline (a V2 -> V3 resource bump) so MM `count==2` samples get their true tail length, replacing the stopgap.
- Decide whether removing the `parentLayer` guard should instead be a narrower validity check.

These changes are logically separable — (1) the SFX fix, (2) the osRecvMesg crashes, (3) the pointer-width fixes, (4) the two stopgaps — and could be split into multiple commits if preferred. Squashed here for review.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7761239390.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7761214700.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/7761114554.zip)

### DIFF
--- a/mm/2s2h/resource/importer/AudioSampleFactory.cpp
+++ b/mm/2s2h/resource/importer/AudioSampleFactory.cpp
@@ -216,6 +216,20 @@ ResourceFactoryBinaryAudioSampleV2::ReadResource(std::shared_ptr<Ship::File> fil
     audioSample->loop.end = reader->ReadUInt32();
     audioSample->loop.count = reader->ReadUInt32();
 
+    // 2S2H [Port] AdpcmLoop.sampleEnd (offset 0x0C, the clip's total s16-sample count) isn't
+    // carried by the .o2r, and isn't extracted by ZAPD. As far as we can tell that's because the
+    // OTR pipeline was authored for OoT, whose audio engine never reads this field (sm64's doesn't
+    // either — it names 0x0C `pad`). MM is the exception: its count==2 "loop N times then stop"
+    // path in synthesis.c reads it. So for MM the field was left uninitialized, which on the build
+    // where we hit this read as poison (0xBEC0FEBE) -> huge sampleEndPos -> out-of-bounds sample
+    // read -> crash. As a minimal, format-compatible stopgap we initialize it to the loop end:
+    // exact for one-shots, and a safe (possibly slightly short) value for count==2 samples. The
+    // proper fix — extracting the real 0x0C from the ROM (ZAPD ZAudio + OTRExporter + a
+    // resource-version bump, then re-extract) — really belongs to whoever maintains the asset
+    // pipeline; it's the only way to recover the true tail length for count==2 samples, and we'd
+    // defer to the maintainers on the format/version conventions for it.
+    audioSample->loop.sampleEnd = audioSample->loop.end;
+
     // This always seems to be 16. Can it be removed in V3?
     uint32_t loopStateCount = reader->ReadUInt32();
     for (int i = 0; i < 16; i++) {
@@ -263,6 +277,7 @@ ResourceFactoryXMLAudioSampleV0::ReadResource(std::shared_ptr<Ship::File> file,
         audioSample->loop.start = loopRoot->UnsignedAttribute("Start");
         audioSample->loop.end = loopRoot->UnsignedAttribute("End");
         audioSample->loop.count = loopRoot->UnsignedAttribute("Count");
+        audioSample->loop.sampleEnd = audioSample->loop.end; // 2S2H [Port]: see binary path above
         tinyxml2::XMLElement* predictor = loopRoot->FirstChildElement("Predictor");
         while (predictor != nullptr) {
             audioSample->loop.state[i++] = predictor->IntAttribute("State");
--- a/mm/2s2h/resource/type/AudioSample.h
+++ b/mm/2s2h/resource/type/AudioSample.h
@@ -9,7 +9,13 @@ typedef struct {
     /* 0x00 */ u32 start;
     /* 0x04 */ u32 end;
     /* 0x08 */ u32 count;
-    /* 0x0C */ char unk_0C[0x4];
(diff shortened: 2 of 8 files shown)


garrettjoecox:
@billsix  Thank you for the PR! Just wanted to at least get you a response so you didn't feel completely ignored.

Basically anything audio is considered witchcraft by most of the active developers on this project, so it's difficult for us to thoroughly vet & review these types of changes. So this may sit around until we can get some eyes on it from people who are knowledgeable on this side of things, or if you can recruit people to playtest this per platform so we know it's not going to cause any new bugs that would also help a lot!

billsix:
@garrettjoecox thanks for the feedback!  A person on discord said that this PR fixes their issues from 2 years ago on aarch64 (which I've never tried) which was pretty sweet to hear! 
I totally understand your point of view.  I know for me, for native compilation on Fedora, this patch made it run instead of break, so in the future, maybe this patch may be useful if newer build breaks

SproedKartoffelChip:
> @garrettjoecox thanks for the feedback! A person on discord said that this PR fixes their issues from 2 years ago on aarch64 (which I've never tried) which was pretty sweet to hear! I totally understand your point of view. I know for me, for native compilation on Fedora, this patch made it run instead of break, so in the future, maybe this patch may be useful if newer build breaks

Yep, that person is me. I've been playing through the build on a linux handheld and have suffered no audio issues so far personally.

bdunahu:
applying this patch to my recipe for 4.0.2 did fix these segfaults on file select, thanks! game's running without other issues the first time I'm playing it
[2Ship2Harkinian.log](https://github.com/user-attachments/files/30289802/2.Ship.2.Harkinian.log)
```
uname -a
Linux garmr 7.0.14-gnu #1 SMP PREEMPT_DYNAMIC 1 x86_64 GNU/Linux
```

PR #1771, opened 2026-07-04
Fix some shadow flickering

garrettjoecox:
**Disclaimer**, not only do I not fully understand our interpolation mechanism, but I also don't fully understand this fix. This was what was suggested after some back and forth with claude, and it does indeed resolve the issue in-game without any known side effects.

### Claude summary:
Cause: The cow's body scale wobbles slightly (breathing animation) each frame, and its scale.x keeps crossing back and forth over equal-to-scale.z. The shadow code treats "circle with equal x/z scale" as a special fast path that skips rotation. So the shadow's draw call keeps flip-flopping between "rotated" and "not rotated" code paths.

The interpolation system stores those two paths' matrices differently — one is a finished matrix, the other is a half-finished matrix waiting on an extra step. When one frame used path A and the next used path B, the code blended these two different-shaped matrices together like they were compatible. They're not, so you get garbage for a frame — which shows up as the shadow vanishing.

Fix: Always also store the finished matrix, untouched, regardless of which path was taken. When adjacent frames took different paths, just use that finished matrix from the new frame instead of trying to blend.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8083054884.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8082979314.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8082904382.zip)

### DIFF
--- a/mm/2s2h/Enhancements/FrameInterpolation/FrameInterpolation.cpp
+++ b/mm/2s2h/Enhancements/FrameInterpolation/FrameInterpolation.cpp
@@ -121,6 +121,10 @@ union Data {
     struct {
         Mtx* dest;
         MtxF src;
+        // The matrix as it actually was for this frame, before any actor-relative adjustment. Kept around
+        // (regardless of has_adjusted) so a has_adjusted mismatch between frames has a correct, non-relative
+        // matrix to fall back to instead of blending across incompatible spaces.
+        MtxF raw;
         bool has_adjusted;
     } matrix_to_mtx;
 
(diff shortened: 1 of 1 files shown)


Eblo (approved):

garrettjoecox:
Going to merge this acknowledging it will be the first thing we revert if we run into any new GFX/interpolation issues in the near future.

PR #1787, opened 2026-07-09
Add extra frames to new day title card to simulate load time

### LINKED ISSUE #1786: "Dawn of the X Day" title screen plays at incorrect speed (too fast)
- [x] I have checked checked the known issues and searched Github to ensure this issue has not already been reported.

### Description of the inaccuracy:
Whenever a day title screen appears (e.g. Dawn of the First Day - 72 Hours Remain), it displays for a noticeably shorter period than it did on the original hardware. It also cuts off the audio that is meant to play during the transition (gentle bird chirps in the background). Technically the code is accurate to the decompile, but the acceleration is probably due to decreased load times on modern machines.

This exact bug was noticed and reported [here](https://github.com/Zelda64Recomp/Zelda64Recomp/issues/198) in the Zelda64 recomp project, and later fixed in [this patch](https://github.com/Zelda64Recomp/Zelda64Recomp/pull/266) - so Ship of Harkinian could probably take a very similar approach.

### Screenshots or videos of 2Ship's behavior:
Most easily reproduced by simply starting a new game with the "skip intro" setting enabled. 

### Screenshots or videos of the original hardware's behavior:
Here's a video showing the title card running on a Wii U Virtual Console
https://youtu.be/kUMTRLhw3rw?si=Tue6Ex6G8zI4C-vA&t=1412

### System Information:
- OS: Windows 11
- GFX Backend: DirectX
- Game Version: Keiichi Charlie
- Rom Version: US 1.0

sirrus233:
Fixes https://github.com/HarbourMasters/2ship2harkinian/issues/1786

This is a very simple tweak to add additional frames to the "new day" title card duration, so that its perceived length is closer to what it was on real hardware.

The same approach is taken [in this PR, in the Zelda64 recomp project](https://github.com/Zelda64Recomp/Zelda64Recomp/pull/266) which also displayed the bug.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8272572998.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8272360049.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8272146882.zip)

### DIFF
--- a/mm/src/overlays/gamestates/ovl_daytelop/z_daytelop.c
+++ b/mm/src/overlays/gamestates/ovl_daytelop/z_daytelop.c
@@ -246,7 +246,7 @@ void DayTelop_Init(GameState* thisx) {
     View_Init(&this->view, this->state.gfxCtx);
     this->state.main = DayTelop_Main;
     this->state.destroy = DayTelop_Destroy;
-    this->transitionCountdown = 260;
+    this->transitionCountdown = 140;
     this->fadeInState = DAYTELOP_HOURSTEXT_OFF;
 
     if (gSaveContext.save.day < 9) {
--- a/mm/src/overlays/gamestates/ovl_daytelop/z_daytelop.c
+++ b/mm/src/overlays/gamestates/ovl_daytelop/z_daytelop.c
@@ -246,7 +246,7 @@ void DayTelop_Init(GameState* thisx) {
     View_Init(&this->view, this->state.gfxCtx);
     this->state.main = DayTelop_Main;
     this->state.destroy = DayTelop_Destroy;
-    this->transitionCountdown = 140;
+    this->transitionCountdown = 260;
     this->fadeInState = DAYTELOP_HOURSTEXT_OFF;
 
     if (gSaveContext.save.day < 9) {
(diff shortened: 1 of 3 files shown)


mm/src/overlays/gamestates/ovl_daytelop/z_daytelop.c
```diff
@@ -246,7 +246,7 @@ void DayTelop_Init(GameState* thisx) {
     View_Init(&this->view, this->state.gfxCtx);
     this->state.main = DayTelop_Main;
     this->state.destroy = DayTelop_Destroy;
-    this->transitionCountdown = 140;
+    this->transitionCountdown = 260;
```
Eblo: This is a neat restoration, but we try to avoid directly modifying source if possible. In this case, your enhancement will always enforce the new duration. What we would prefer would be an opt-in enhancement that can be toggled from the menu.

In fact, for something like this, you could get away without modifying the source at all, using an `OnActorInit` hook for DayTelop that sets `transitionCountdown = 260` purely within the enhancement file.

sirrus233: Thanks for the feedback! I'll take a look at that approach.

sirrus233: I took a shot at updating this per your review. It seems the title card isn't an actor, so I don't think `OnActorInit` works. I ended up pairing `OnGameStateMainStart` with a first-frame detection guard. It worked in my testing, but please let me know if that's a valid approach!

Eblo: Oops my bad on the actor, but indeed that is as valid an approach as any, I'd say. Thanks for making that change!

Eblo (approved):

PR #1793, opened 2026-07-15
Fix mod load time regression

### LINKED ISSUE #1789: Regression in mod loading times
- [X] I have checked the known issues and searched Github to ensure this issue has not already been reported.
- [X] This is not a feature request, I know those belong on Discord/Github Discussions.

### Description of the bug:
There appears to be a regression in mod load times on the latest nightly https://github.com/HarbourMasters/2ship2harkinian/commit/a7e3e7c29be41cf67d89bdcb2f501d36d6106544 when compared against Keiichi Charlie [4.0.2](https://github.com/HarbourMasters/2ship2harkinian/tree/4.0.2)

When running the [latest version](https://github.com/GhostlyDark/MM-Reloaded-2S2H/releases/tag/v11.0.2) of the 4k Majora's Mask Reloaded texture pack (about ~22 GB o2r file), the startup time of 2ship is increased by about 10 seconds on my machine.

- [X] Yes, here are the steps to reproduce the bug:

1. Download the MM Reloaded 4k textures and place in mods directory
2. Open 2ship latest public release and measure startup time
3. Open 2ship latest nightly and measure startup time
4. Weep.

### Expected behavior:
Similar load times between two versions running the same mod.

### System Information:
- OS: Windows 11
- GFX Backend: DirectX
- Game Version: See issue description
- Rom Version: 1.0.0 US

sirrus233:
Resolves https://github.com/HarbourMasters/2ship2harkinian/issues/1789

The long startup time turned out to be caused by eager loading of archived resources for custom cosmetics.

**What was causing long load times?**
Prior to this PR, when `ScanDynamicCosmetics` was called, it did a `archiveManager->ListFiles("*");` to retrieve every path of every loaded resource from any mods, and then indiscriminately called `TryLoadCustomDisplayListXml` on each path. With a large texture pack such as MM Reloaded, this means allocating memory and decompressing many gigabytes of texture files - only to determine they are not XML and do nothing with them. In practice this ends up translating to > 60 second boot times when using the 4k version of MM Reloaded on a Steam Deck.

**Fix Approach**
Really, this issue has already been handled reasonably well in [Shipwright](https://github.com/HarbourMasters/Shipwright/blob/585530f68d8c3c94fef4e08d8616aa06f1d19feb/soh/soh/Enhancements/cosmetics/DynamicCosmeticsEditor.cpp#L209). My changes here simply reflect what Shipwright was already doing, which is to read a cosmetics manifest from a known path in the archive, and use that to find the resources that actually need to be processed.

It should be noted that this *will* require that 2ship mods providing custom cosmetics will need to provide this manifest file.

**Differences from Shipwright**
In this PR I opted to aggregate all the manifest entries in one pass, and then process their contents in another. This isn't a behavioral change and the outcome should be the same. But I found this produced a nicer diff and read rather cleanly (IMO).

I also fixed a pre-existing bug I noticed when comparing Shipwright and 2ship: in the stable sort 2ship was initializing the sort order to `2` (probably because that's correct for Shipwright), but it should really be `6` due to the fact 2ship has more categories.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8357214522.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8357124671.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8356877033.zip)

### DIFF
--- a/mm/2s2h/BenGui/DynamicCosmeticEditor.cpp
+++ b/mm/2s2h/BenGui/DynamicCosmeticEditor.cpp
@@ -416,14 +416,73 @@ void ScanDynamicCosmetics() {
     auto resourceManager = Ship::Context::GetInstance()->GetResourceManager();
     auto archiveManager = resourceManager->GetArchiveManager();
     RefreshCustomModelActiveFlags(archiveManager.get());
-    auto materialPaths = archiveManager->ListFiles("*");
+    auto archives = archiveManager->GetArchives();
+    struct ManifestEntry {
+        std::string materialPath;
+        std::string cosmeticEntry;
+        std::string cosmeticCategory;
+        bool hasCosmeticCategory = false;
+        bool isPrimColor = false;
+    };
+    std::vector<ManifestEntry> manifestEntries;
     std::unordered_map<std::string, size_t> entryIndicesByKey;
 
-    for (const auto& materialPath : *materialPaths) {
-        if (!IsCustomArchive(archiveManager->GetArchiveFromFile(materialPath))) {
+    for (const auto& archive : *archives) {
+        if (!IsCustomArchive(archive)) {
             continue;
         }
 
+        auto manifestFile = archive->LoadFile("CosmeticEntries");
+        if (manifestFile == nullptr || !manifestFile->IsLoaded || manifestFile->Buffer == nullptr) {
+            continue;
+        }
+
+        tinyxml2::XMLDocument manifestDocument;
+        manifestDocument.Parse(manifestFile->Buffer->data(), manifestFile->Buffer->size());
+        if (manifestDocument.Error()) {
+            continue;
+        }
+
+        tinyxml2::XMLElement* manifestRoot = manifestDocument.FirstChildElement();
+        if (manifestRoot == nullptr) {
+            continue;
+        }
+
+        for (auto* manifestEntry = manifestRoot->FirstChildElement(); manifestEntry != nullptr;
+             manifestEntry = manifestEntry->NextSiblingElement()) {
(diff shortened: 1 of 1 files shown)


Jameriquiah (approved):
this was on my todo list for 2ship, just hadnt got around to it yet lol, but tbh this looks good. functionality is left intact and is behaviorally identical to my soh implementation. good work!

mm/2s2h/BenGui/DynamicCosmeticEditor.cpp
```diff
@@ -416,14 +416,80 @@ void ScanDynamicCosmetics() {
...
-    auto materialPaths = archiveManager->ListFiles("*");
+    auto archives = archiveManager->GetArchives();
+    struct ManifestEntry {
+        std::string materialPath;
+        std::string cosmeticEntry;
+        std::string key;
+        std::string cosmeticCategory;
+        bool hasCosmeticCategory = false;
+        bool isPrimColor = false;
+    };
```
Eblo: Move this struct definition to near the top where `CustomCosmeticBinding` and `CustomCosmeticEntry` are.

Eblo (commented):
Just a teensy cleanup comment, but otherwise LGTM. This fixes the issue on my end.

mm/2s2h/BenGui/DynamicCosmeticEditor.cpp
sirrus233: Done!

Eblo (approved):

PR #1817, opened 2026-07-26
Improvements to excluded checks

garrettjoecox:
A few things:
- Converts excluded checks to be stored in an array in the json
- Uses `RC_` strings instead of the numbers so they survive updates
- Fix a bug that prevented arrays from applying correctly in presets
- Add buttons to excluded checks interface for quickly excluding Termina Field & Cow Grotto grass, along with some common exclusions.
- Adjust heading of check excludes lists

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8641600712.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8641578290.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8641576926.zip)

### DIFF
--- a/mm/2s2h/PresetManager/PresetManager.cpp
+++ b/mm/2s2h/PresetManager/PresetManager.cpp
@@ -7,6 +7,7 @@
 #include "2s2h/BenGui/UIWidgets.hpp"
 #include "2s2h/BenGui/Notification.h"
 #include <ship/window/FileDropMgr.h>
+#include <ship/config/Config.h>
 
 std::unordered_map<std::string, std::string> tagMap = {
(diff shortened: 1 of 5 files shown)


garrettjoecox:
Additional change made here, if a check is excluded and it's vanilla item is junk, we don't even shuffle it. We only shuffle and force it to be junk if it's vanilla item is not junk (which is the current behavior for all excluded checks in develop)

PR #1818, opened 2026-07-26
[Enhancement] Goron Rolling Spikes requires Shield button to be pressed

### LINKED ISSUE #1747: [Enhancement] Goron roll R button magic saver
I think it would be cool if it was possible to roll with goron Link without using the spikes unless you press or hold down the R button, upon which the regular spike charge-up would start. This would work similarly to the zora swimming shield, and it would synergise very well with the Fast Spikes option

I know there's already an option to make it so the spikes don't consume magic, but I think this idea is a little more balanced

JordanLongstaff:
Closes #1747

https://github.com/user-attachments/assets/4e748607-b724-4468-bb3f-da102d5106cd

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8822950072.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8822926930.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8822839785.zip)


mm/2s2h/Enhancements/Masks/GoronRollingIgnoresMagic.cpp
```diff
@@ -17,18 +20,27 @@ void RegisterGoronRollingIgnoresMagic() {
...
+    COND_VB_SHOULD(VB_GORON_ROLL_DISABLE_SPIKE_MODE, CVAR || CVAR_R, {
         Player* player = GET_PLAYER(gPlayState);
-        bool disableSpikes = (player->stateFlags3 & PLAYER_STATE3_80000) &&
+        bool disableSpikes = CVAR && (player->stateFlags3 & PLAYER_STATE3_80000) &&
                              (!CHECK_BTN_ALL(sPlayerControlInput->cur.button, BTN_A) ||
                               ((player->av1.actionVar1 == 4) && (player->unk_B08 < 12.0f)));
         if (!disableSpikes) {
             *should = false;
         }
+        if (CVAR_R && !CHECK_BTN_ALL(sPlayerControlInput->cur.button, BTN_R)) {
```
garrettjoecox, line 41: This appears to let me continue to spike roll after my magic depletes when holding R, going to assume it's this condition or something in the logic above

garrettjoecox (changes requested):

mm/2s2h/Enhancements/Masks/GoronRollingIgnoresMagic.cpp
JordanLongstaff: I think you're misreading this. The `should` here is controlling whether spikes should be _disabled_, and they should if R isn't held. It's not continuing to do the spikes with R being held if it would normally be disabled.

JordanLongstaff: Unless you think I put `CVAR` in the wrong place in the `disableSpikes` condition, which could be true.

garrettjoecox: I'm not sure exactly where the logic is wrong, but what I mean is in game, the bug is occuring. I continue to have spikes after my magic is depleted if I'm holding R, which normally would make them disappear.

JordanLongstaff: Well, I've just confirmed the bug and fixed it. Thanks for pointing it out.

PR #1823, opened 2026-07-28
Add categorized sequence randomization

frogssoldseparately:
Currently, music shuffling is done by groups as defined for the audio editor UI (battle music <-> battle music, event music <-> event music, etc.). While this is a fine implementation of music shuffling as it is, it does get a bit weird at times with custom songs that were intended only for specific types of events or regions, for example. What if every background song *couldn't* replace every other background song?

## What's Changed

- Added support for custom sequence categories that are distributed with .zseq and .mmrs files, and takes these into account while shuffling music.
- The vanilla MM sequences have been categorized under the same standard, so both custom sequences and vanilla sequences will (attempt to) be placed more appropriately upon shuffling.
- This categorized shuffling can be toggled in the Audio Editor settings, falling back to the original shuffle method with it disabled.
- Maintains support for the current format (`_bgm`/`_fanfare`) for custom sequence names. The same .o2r file will be compatible with and without this change. Sequences with these suffixes will inherit all background music categories or fanfare categories accordingly.
- Events are now fanfares.

## How's It Done

I left the original typing of sequences intact, as it's used by the audio editor UI to group them into different tabs. For the sake of familiarity, maybe this should be left as is. I've added additional information to the `SequenceInfo` struct so that sequences can be sorted based on the following categories:

```
0 -> Field
1 -> Town
2 -> Dungeons
3 -> Indoors
4 -> Minigames
5 -> Action
6 -> Calm
7 -> Bosses
8 -> Get Item Fanfares
9 -> Game Over Fanfares
10 -> Event Clear Fanfares
16 -> Title Screen
```

as well as sequence specific replacements of the format `1[sequenceIdHex]` (e.g., `120` for the Game Over sequence).

When a specific sequence slot is being filled, a random sequence with *any* of its categories (or a replacement target of this sequence's id) will be chosen. If there is no such sequence that hasn't already been used, it will disregard categorization.

## Considerations

For the time being, I left ocarina songs uncategorized as there's not really an appropriate category from above that fits them. This just means they'll be shuffled as they were.

Sequence specific replacements (categories above `100`) may disproportionately take up a slot as a byproduct of how random sequences are picked. I couldn't think of a great way to deal this without drastically increasing how long shuffling takes. This may need addressing.

While retro will just accept numeric categories in place of "bgm" or "fanfare" in the .meta files during o2r packaging, cache policy for a custom sequence is set differently when it's fanfare. Providing numeric categories currently defaults to a cache policy of `PERSISTENT`, regardless of the sequence being fanfare or not. This could be a thing, and the inclusion of categorization here may necessitate changes to retro in turn. I've already modified my bank packaging tool to account for this, but most are going to be using retro.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8913554177.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8913521470.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8913361791.zip)


mm/2s2h/Enhancements/Audio/AudioCollection.cpp
```diff
@@ -249,14 +327,48 @@ void AudioCollection::AddToCollection(char* otrPath, uint16_t seqNum) {
...
+            std::string category;
+            // typeString can be assumed to be dash separated list of categories
+            try {
+                while ((dashPos = typeString.find('-')) != std::string::npos) {
+                    category = typeString.substr(0, dashPos);
+                    ParseSequenceCategory(category, compositeCategory, seqIdReplacements);
+                    typeString = typeString.substr(dashPos + 1);
+                }
+                ParseSequenceCategory(typeString, compositeCategory, seqIdReplacements);
+            } catch (std::invalid_argument e) { compositeCategory = SEQ_CAT_BGM; }
```
garrettjoecox: This can throw more than just invalid_argument here, should catch all exceptions

mm/2s2h/Enhancements/Audio/AudioEditor.cpp
```diff
@@ -187,60 +188,152 @@ void UpdateCurrentBGM(u16 seqKey, SeqType seqType) {
...
+                // Stop looking for a valid category after all applicable ones have been seen, or when one is found.
+                while (compositeCategory != targetFilter && (!matchFlag || seqPools[currentCategory].empty())) {
+                    if (randomCategory == includedCategories.end()) {
+                        std::shuffle(includedCategories.begin(), includedCategories.end(), g);
+                        randomCategory = includedCategories.begin();
+                    }
+                    currentCategory = *randomCategory++;
+                    matchFlag = targetFilter & currentCategory;
+                    compositeCategory |= matchFlag;
+                }
```
garrettjoecox, line 309: A double nested while loop inside of the for loop here. I don't fully understand what all is going on here, but I have a feeling some of this can be cleaned up/optimized

PR #1824, opened 2026-07-31
Ios

PatFromSplatt:


garrettjoecox:
@PatFromSplatt going to assume this was an accident, if you intended on sharing this I'd suggest PR'ing into your own fork and sharing it in our `2s2h-modding` forums on the discord!

Jameriquiah:
if this is an ios port then a `2s2h-chat` thread would be preferable over `2s2h-modding` wouldnt it?

PR #1840, opened 2026-08-03
Change how skulltula shuffle works

garrettjoecox:
This was originally requested by zfg a while ago, I'm guessing it's probably consistent with how MMR or OOTXMM does things.

Current behavior:
- Enabling Gold Skulltula Shuffle shuffles _all_ spiders in both houses
- 1 Slider to configure how many tokens are required for spider house rewards
- 1 Slider to configure how many tokens end up in the item pool.

New Behavior:
- Enabling Gold Skulltula Shuffle shuffles _a configured number_ of spiders in both houses.
- 1 Slider to configure how many tokens are required for spider house rewards
- 1 Slider to configure how many spiders are shuffled. The remaining spiders have their vanilla tokens. The selection of spiders is chosen at random per seed.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8861832880.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8861775713.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8861739069.zip)


Eblo (commented):
~~This seems to have introduced a bug where `RC_SWAMP_SKULLTULA_MAIN_ROOM_JAR` can yield two swamp tokens instead of one. This doesn't seem to affect other checks from my testing, but the issue may not be limited to that one. It appears to bump the swamp token by 1 immediately upon contact, then once more upon picking up the item. This does not happen my local build in a seed where `RC_SWAMP_SKULLTULA_MAIN_ROOM_JAR` is still a swamp token.~~

Edit: This was happening consistently in one runtime between resets, but now I cannot reproduce this. Nothing in this PR seems like it would cause such a thing, so disregard this as relevant to the changes made here.

Eblo (approved):

PR #1850, opened 2026-08-05
Time Moves When You Move: fix saving zero time speed

### LINKED ISSUE #1841: "Time Moves When You Move" freezes clock when loading game
### Description of the bug:
I recently played 2ship2harkinian on my SteamDeck and wanted to try the "Time Moves When You Move" enhancement. Only a while later I noticed that my time was entirely frozen, even when moving and such.
It seems whenever that enhancement is active and I'd restart the game, the setting would get "frozen" or so and not turn the time back on.

1. Enable “Time Moves When You Move”
2. quit 2Ship completely
3. relaunch it and load a save
4. Time Speed is forced to 0 and remains frozen even while Link moves
5. Disabling and re-enabling the enhancement fixes it until the next full application restart (Resetting from within 2Ship does not reproduce the problem.)

### Expected behavior:
When "Time Moves When You Move" is active, time should only freeze when standing still.

### Screenshots or videos:

https://imgur.com/a/4cRQwoX

In this video I also opened the cheat editor sidebar and you can see how the "Time Speed" setting would freeze when I start to move.  I turned the time speed up to 3 manually several times, to show how it turns it down to 0. I also disabled the enhancement to show how it would work again then.

### System Information:
- OS: SteamOS
- GFX Backend: OpenGL
- Game Version: 4.0.2
- Rom Version: Can't remember, anywhere to check that inside the game?

JordanLongstaff:
Fixes #1841

Two main issues contributed to this bug:
1. The enhancement stores a backup time speed variable that is used to restore the correct time speed whenever the player moves around. However, that variable gets set for the first time when a save is loaded and does not get reset if another save is loaded.
2. Owl Saves remember the time speed when they are saved, so if that speed is 0, then that will be the value stored in the backup, meaning that both the backup _and_ the main time speed variables are 0, so time will never start moving.

I fixed the first issue by adding an `OnSaveLoad` hook to reset the backup variable, and I fixed the second issue by adding more conditions to the "is time moving" check. This ensures that the time speed is correct when it gets saved. I also factored Pause Save into this. All that left was Autosave, which I fixed by simply disabling it when the player isn't moving. It'll immediately trigger again when the player starts moving.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8949407251.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8947238577.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/8947145135.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
+++ b/mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
@@ -19,11 +19,12 @@ static void UpdateTimeSpeedOffset(PauseContext* pauseCtx) {
     // Time is considered to be moving when any of the following is true:
     // - The player is moving (since that's the basis of this enhancement)
     // - The player is playing the Ocarina (so that Inverted SoT still works properly)
-    // - The player is in a choice dialog (so that Owl Saving doesn't save the wrong time speed)
+    // - The player is choosing whether to save at an Owl Statue (so that it doesn't save the wrong time speed)
     // - The pause menu save prompt is open (so that Pause Save doesn't either)
-    bool timeShouldMove = (player->stateFlags2 & PLAYER_STATE2_USING_OCARINA) || player->speedXZ != 0.0f ||
-                          Message_GetState(&gPlayState->msgCtx) == TEXT_STATE_CHOICE ||
-                          pauseCtx->state == PAUSE_STATE_SAVEPROMPT;
+    bool timeShouldMove =
+        (player->stateFlags2 & PLAYER_STATE2_USING_OCARINA) || player->speedXZ != 0.0f ||
+        (Message_GetState(&gPlayState->msgCtx) == TEXT_STATE_CHOICE && gPlayState->msgCtx.currentTextId == 0xC01) ||
+        pauseCtx->state == PAUSE_STATE_SAVEPROMPT;
 
     if (timeShouldMove && sStoredTimeOffset != DEFAULT_TIME_OFFSET) {
         gSaveContext.save.timeSpeedOffset = sStoredTimeOffset;
--- a/mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
+++ b/mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
@@ -21,10 +21,11 @@ static void UpdateTimeSpeedOffset(PauseContext* pauseCtx) {
     // - The player is playing the Ocarina (so that Inverted SoT still works properly)
     // - The player is choosing whether to save at an Owl Statue (so that it doesn't save the wrong time speed)
     // - The pause menu save prompt is open (so that Pause Save doesn't either)
+    // - The Game Over save prompt is open (not reachable right now, but there for future-proofing)
     bool timeShouldMove =
         (player->stateFlags2 & PLAYER_STATE2_USING_OCARINA) || player->speedXZ != 0.0f ||
         (Message_GetState(&gPlayState->msgCtx) == TEXT_STATE_CHOICE && gPlayState->msgCtx.currentTextId == 0xC01) ||
-        pauseCtx->state == PAUSE_STATE_SAVEPROMPT;
+        pauseCtx->state == PAUSE_STATE_SAVEPROMPT || pauseCtx->state == PAUSE_STATE_GAMEOVER_SAVE_PROMPT;
 
     if (timeShouldMove && sStoredTimeOffset != DEFAULT_TIME_OFFSET) {
         gSaveContext.save.timeSpeedOffset = sStoredTimeOffset;
--- a/mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
+++ b/mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
@@ -88,7 +88,12 @@ static void RegisterTimeMovesWhenYouMove() {
 
     COND_HOOK(OnKaleidoUpdate, CVAR, UpdateTimeSpeedOffset);
 
-    COND_HOOK(OnSaveLoad, CVAR, [](s16) { sStoredTimeOffset = DEFAULT_TIME_OFFSET; });
+    COND_HOOK(OnSaveLoad, CVAR, [](s16) {
+        sStoredTimeOffset = DEFAULT_TIME_OFFSET;
+        if (gSaveContext.save.timeSpeedOffset == -R_TIME_SPEED) {
+            gSaveContext.save.timeSpeedOffset = 0;
+        }
+    });
 }
 
 static RegisterShipInitFunc initFunc(RegisterTimeMovesWhenYouMove, { CVAR_NAME });
--- a/mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
+++ b/mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
@@ -89,13 +89,16 @@ static void RegisterTimeMovesWhenYouMove() {
     COND_ID_HOOK(OnActorUpdate, ACTOR_PLAYER, CVAR, [](Actor* actor) { UpdateTimeSpeedOffset(&gPlayState->pauseCtx); });
 
     COND_HOOK(OnKaleidoUpdate, CVAR, UpdateTimeSpeedOffset);
+}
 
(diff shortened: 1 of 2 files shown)


mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
```diff
@@ -60,21 +84,11 @@ void RegisterTimeMovesWhenYouMove() {
...
-        } else if (!timeShouldMove && sStoredTimeOffset == DEFAULT_TIME_OFFSET) {
-            sStoredTimeOffset = gSaveContext.save.timeSpeedOffset;
-            gSaveContext.save.timeSpeedOffset = -R_TIME_SPEED;
-        }
-    });
+    COND_ID_HOOK(OnActorUpdate, ACTOR_PLAYER, CVAR, [](Actor* actor) { UpdateTimeSpeedOffset(&gPlayState->pauseCtx); });
+
+    COND_HOOK(OnKaleidoUpdate, CVAR, UpdateTimeSpeedOffset);
+
+    COND_HOOK(OnSaveLoad, CVAR, [](s16) { sStoredTimeOffset = DEFAULT_TIME_OFFSET; });
```
mckinlee: Does resetting only `sStoredTimeOffset` also repair saves that were already written with `timeSpeedOffset == -R_TIME_SPEED`. My assumption is that after loading one of those saves, the frozen value can become the new backup and continue being restored indefinitely. If not, this might need a migration.

mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
```diff
@@ -13,7 +13,31 @@ extern "C" {
...
+    Player* player = GET_PLAYER(gPlayState);
+
+    // Time is considered to be moving when any of the following is true:
+    // - The player is moving (since that's the basis of this enhancement)
+    // - The player is playing the Ocarina (so that Inverted SoT still works properly)
+    // - The player is in a choice dialog (so that Owl Saving doesn't save the wrong time speed)
+    // - The pause menu save prompt is open (so that Pause Save doesn't either)
+    bool timeShouldMove = (player->stateFlags2 & PLAYER_STATE2_USING_OCARINA) || player->speedXZ != 0.0f ||
+                          Message_GetState(&gPlayState->msgCtx) == TEXT_STATE_CHOICE ||
+                          pauseCtx->state == PAUSE_STATE_SAVEPROMPT;
```
mckinlee: What about `PAUSE_STATE_GAMEOVER_SAVE_PROMPT`? The game over save path also serializes `SaveContext`, so I assume while the player is stationary it can still persist the frozen time offset. Although it might be preferable to restore/reapply the real offset around all save serialization through a central save hook.

mm/2s2h/Enhancements/Saving/AutoSave.cpp
```diff
@@ -47,6 +47,11 @@ void HandleAutoSave() {
         return;
     }
 
+    // Time must be moving for autosave to work. Fixes an issue with Time Moves When You Move.
+    if (gSaveContext.save.timeSpeedOffset == -R_TIME_SPEED) {
```
mckinlee: Isn't this condition is also true for both `R_TIME_SPEED == 0` and `timeSpeedOffset == 0`?

mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
```diff
@@ -13,7 +13,31 @@ extern "C" {
...
+static void UpdateTimeSpeedOffset(PauseContext* pauseCtx) {
+    Player* player = GET_PLAYER(gPlayState);
+
+    // Time is considered to be moving when any of the following is true:
+    // - The player is moving (since that's the basis of this enhancement)
+    // - The player is playing the Ocarina (so that Inverted SoT still works properly)
+    // - The player is in a choice dialog (so that Owl Saving doesn't save the wrong time speed)
+    // - The pause menu save prompt is open (so that Pause Save doesn't either)
+    bool timeShouldMove = (player->stateFlags2 & PLAYER_STATE2_USING_OCARINA) || player->speedXZ != 0.0f ||
+                          Message_GetState(&gPlayState->msgCtx) == TEXT_STATE_CHOICE ||
```
mckinlee: `TEXT_STATE_CHOICE` applies to every choice dialogue, not specifically the Owl Statue save prompt from my understanding. This may have some unintended consequences to look into.

mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
JordanLongstaff: It's true that if you do this with a save that has 0 saved as its time speed, it'll be stuck unless it gets changed (either in Dev Tools or with SoT). I guess I could add a conditional reset to the default, which is 3.

mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
JordanLongstaff: Is there a way to reach that pause state? Do we have an enhancement that allows the player to save during Game Over?

mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
JordanLongstaff: Does it? I found (by testing) that if you select No to saving, the time speed does correctly go back to 0, so that tells me that there shouldn't be any unintended consequences. Is there a situation that comes to mind for you? Do you want me to add a text ID check for sanity?

mm/2s2h/Enhancements/Saving/AutoSave.cpp
JordanLongstaff: Oh, that's a good point. I should fix that.

mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
garrettjoecox: I don't think we do no, that's a holdover from oot

mckinlee: I withdraw my comment lol I coulda swore we used it for something. I don't see anything.

mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
mckinlee: Up to you. I'm just being overly cautious.

mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
```diff
@@ -60,19 +86,14 @@ void RegisterTimeMovesWhenYouMove() {
...
 
-            // This is for the section above, lets arrows continue flying after they were fired with time frozen
-            // player->unk_D57 = 4;
-        } else if (!timeShouldMove && sStoredTimeOffset == DEFAULT_TIME_OFFSET) {
-            sStoredTimeOffset = gSaveContext.save.timeSpeedOffset;
-            gSaveContext.save.timeSpeedOffset = -R_TIME_SPEED;
+    COND_HOOK(OnSaveLoad, CVAR, [](s16) {
+        sStoredTimeOffset = DEFAULT_TIME_OFFSET;
+        if (gSaveContext.save.timeSpeedOffset == -R_TIME_SPEED) {
+            gSaveContext.save.timeSpeedOffset = 0;
```
mckinlee: Questions for ya. I believe this will also run on file select right? I saw in `z_file_choose_NES.c` that R_TIME_SPEED is being set to 10. My concern is a save with a already bad offset will get skipped. Is this something you can confirm? Additionally, this is gated by the cvar so it won't work with it disabled.

mckinlee (commented):
Looks great! but a few more questions for the repair part.

mm/2s2h/Enhancements/Modes/TimeMovesWhenYouMove.cpp
JordanLongstaff: I can't confirm it, but I will reconfigure this to be more guaranteed and not gated.

mckinlee (approved):
Github actions seem to be down, but everything lgtm. Thanks!

PR #1854, opened 2026-08-06
Owl Statue skips to File Select

### LINKED ISSUE #1852: "Skip to File Select" doesn't include Owl Saves

### Description of the bug:
"Skip to File Select" is supposed to skip over the intro cutscene, and it does when you boot up the game. But if you save at an Owl Statue, it quits out and then plays the intro cutscene, even with the option to skip it turned on.

1. Turn on "Skip to File Select"
2. Load a game where Link has a sword (can easily be done using the Skip First Cycle option)
3. Save at an Owl Statue
4. Intro cutscene plays after the game quits out

### Expected behavior:
When you save at an Owl Statue with "Skip to File Select" turned on, it should go right back to File Select, not play the intro cutscene.

### System Information:
- OS: Windows 11
- GFX Backend: DirectX
- Game Version: Keiichi Charlie
- Rom Version: US v1.0

JordanLongstaff:
Fixes #1852

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9287103617.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9287017850.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9287010011.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Cutscenes/SkipToFileSelect.cpp
+++ b/mm/2s2h/Enhancements/Cutscenes/SkipToFileSelect.cpp
@@ -11,7 +11,27 @@ extern "C" {
 #define CVAR_NAME "gEnhancements.Cutscenes.SkipToFileSelect"
 #define CVAR CVarGetInteger(CVAR_NAME, 0)
 
-void RegisterSkipToFileSelect() {
+static constexpr int32_t OPENING_CS_ID = 0x1C00;
+
+// Normally the following is called by the opening cutscene
+static void SkipToFileSelect() {
+    Sram_InitNewSave();
+    gSaveContext.save.time = CLOCK_TIME(8, 0);
+    gSaveContext.save.day = 1;
+    gSaveContext.save.playerForm = PLAYER_FORM_HUMAN;
+    gSaveContext.gameMode = GAMEMODE_FILE_SELECT;
+
+    STOP_GAMESTATE(gGameState);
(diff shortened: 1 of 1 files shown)


Eblo:
LGTM, but could you set this to target `develop`?

JordanLongstaff:
I think this should have made it into the new release, but didn't because it targeted `develop-keiichi`. I did that because I regard it as a bugfix. So shouldn't it target `develop-battler` now instead?

garrettjoecox:
I'd be okay with this going into battler if it gets some thorough QA. I was too worried of regressions to merge it in last minute for the release. 

Reason being anything to do with changing game state like that is sort of fragile, so we have to make sure it doesn't break anything accidentally

JordanLongstaff:
I don't know how thorough you want it, but it seems to still work. All it does is just apply the same skip logic that's already there for the title screen at startup to after the owl save.

Feel free to patch and test this if you want to. Here's a recording that I took of it in its present state.

https://github.com/user-attachments/assets/a3091a0b-acaf-4e0e-8d80-0d631eced264

Eblo (approved):

garrettjoecox:
@JordanLongstaff this seems to be causing playing song of time to send you to file select. I can have a look tomorrow if you aren't able to get to it

JordanLongstaff:
Fixed with #1897. Thanks for bringing that to my attention.

PR #1859, opened 2026-08-08
Allow free look when climbing on objects

lightmanLP:
Literally

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9022553517.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9022528349.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9022437310.zip)

### DIFF
--- a/mm/2s2h/BenGui/BenMenu.cpp
+++ b/mm/2s2h/BenGui/BenMenu.cpp
@@ -888,11 +888,6 @@ void BenMenu::AddEnhancements() {
                      .Min(-89.0f)
                      .Max(89.0f)
                      .DefaultValue(-49.0f));
-    AddWidget(path, "Allow Camera Control During Climb", WIDGET_CVAR_CHECKBOX)
-        .CVar("gEnhancements.Camera.FreeLook.AllowClimb")
-        .PreFunc([](WidgetInfo& info) { info.isHidden = mBenMenu->disabledMap.at(DISABLE_FOR_FREE_LOOK_OFF).active; })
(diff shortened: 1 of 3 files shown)


Jameriquiah:
tbh i think this should just be by default how free look works rather than a separate enhancement

lightmanLP:
Removed requirement in separate commit, choose wisely

garrettjoecox:
I'm not a freecam user, so I have no opinion here, @Jameriquiah do you usually use freecam?

Jameriquiah:
> I'm not a freecam user, so I have no opinion here, @Jameriquiah do you usually use freecam?

always. and free cam works when hanging and climbing in soh so id be happy to have it here, im just not too sure the `z_camera.c` footprint is the best way to do it

lightmanLP:
Last commit does it without requirement of any z_camera changes. I just keept names if anyone would need them

Jameriquiah:
ah perfect then, yea id get rid of that then, and the `gEnhancements.Camera.FreeLook.AllowClimb` cvar probably isnt needed either, could just get rid of that and just have the code go when it needs to for `CAM_FUNC_PARALLEL1`

lightmanLP:
Seems to work with minimal edit now

Jameriquiah (approved):

PR #1863, opened 2026-08-09
16 Bit Sound Font Ids

frogssoldseparately:
Changes some infrastructure to use 16 bit font id's for sequence data, as opposed to the original 8 bit. This provides more sound font slots than could practically be used up.

Allows sequenced music to use crc64 hashes of sound font file names to reference them. This was originally only available for streamed music.

This .o2r file can be used to test. Just change the zip extension to o2r. It has 827 custom sound fonts, and 1923 custom sequences (the current custom sequence limit).

[1786312534.zip](https://github.com/user-attachments/files/30880379/1786312534.zip)

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9820758370.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9820621313.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9820579771.zip)

### DIFF
--- a/mm/2s2h/resource/importer/AudioSequenceFactory.cpp
+++ b/mm/2s2h/resource/importer/AudioSequenceFactory.cpp
@@ -34,8 +34,16 @@ ResourceFactoryBinaryAudioSequenceV2::ReadResource(std::shared_ptr<Ship::File> f
     for (uint32_t i = 0; i < 16; i++) {
         audioSequence->sequence.fonts[i] = 0;
     }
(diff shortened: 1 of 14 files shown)


garrettjoecox (approved):
I thoroughly tested this and committed some of my own changes/tweaks, we're likely going to find more issues in the future with it but will be good to get more widespread testing on develop for a while

PR #1870, opened 2026-08-10
[Enhancement] Trigger active item on B

lightmanLP:
This enhancement allows to use deku sticks, gfs, bow and hookshot with B button. Pressing button again unequips it. Hookshot and bow functions intentionally limited to aim mode.
It's functionally replaces GFS enhancement.

- [x] fix compat issues with arrow cycle on R
- [ ] fix compat with bomb arrows
- [x] strip gfs enhancement after verification?
- [ ] validate dpad
- [ ] validate epona

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9237733517.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9237699821.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9237571845.zip)


mckinlee:
> strip gfs enhancement after verification?

Yes please. No sense in keeping it when this gets in.

JordanLongstaff:
There should be a config migration to transfer the GFS option to this new one, shouldn't there?

mckinlee:
> There should be a config migration to transfer the GFS option to this new one, shouldn't there?

I'm not sure what would be ideal, but it may be confusing to the player to have this enabled when it effects more than just GFS. I'm curious what others opinions might be, but maybe we just disclose the option was removed in the release notes and point them to this one as a alternative.

mckinlee:
Actually, if the toggle for this was broken up per item then the migration would make more sense. Just wanted to throw that out there.

PR #1875, opened 2026-08-11
rando grass optimization

Jameriquiah:
null

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9092050877.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9092005619.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9091823419.zip)


Eblo (approved):

Jameriquiah:
oh this wasnt something i was actually intending to get in yet, was just for proxy to test something on when he got around to it

Eblo:
Oops. Well, it works in my eyes 🤷

Jameriquiah:
lol i mean u can merge it if u want bc it is more optimized but it is subject to change again in the future lol

garrettjoecox:
It seems like this entirely removed the custom grass texture that was applied to all models, which was the only indicator of junk vs already checked, was this intentional?

Jameriquiah:
yes it was intentional, i didnt think they were different enough that most people would be noticing the difference anyway lol, could probably use a new flower texture for it tho instead, i think the token one kind of looks like it should be junk given its texture, could use that and put a different texture on the token one, or just restore the rando grass anyway. up to u, another reason i took it out was that its geometry was identical to the vanilla grass dl and so i figured maybe could be seen as shipping a vanilla asset

garrettjoecox:
While I do think we could and probably should discuss/revisit that issue, I think it should be separate from optimizing the grass tri's, unless the texture itself is what's causing the bloat issues.

Jameriquiah:
texture isnt the cause, i think its a combination of the mesh having been double sided by duplicating the geometry rather than flagging cull back to off, and just how much grass is actually in termina field

Jameriquiah:
which i should clarify i didnt actually change the flower models at all, its still the same model, just handled the backface culling through the material rather than the geometry

Jameriquiah:
also could be the dls not having been inlined since fast64 doesnt natively export them like that, 200+ dls that branch to and from other dls for triangle and material data, probably also causing more load

PR #1882, opened 2026-08-12
properly remove all incompatible o2rs

garrettjoecox:
`std::filesystem::remove("mm.o2r");` was not working on mac release, nor on my local dev because multiple locations were checked for the o2r and it only removed one of them.

This ensures it removes all of them so that the "redirect" actually redirects.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9277534116.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9277345328.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9277299864.zip)

### DIFF
--- a/mm/2s2h/BenPort.cpp
+++ b/mm/2s2h/BenPort.cpp
@@ -238,6 +238,19 @@ bool PathTestCleanup(FILE* tfile) {
     return true;
 }
 
+static bool RemoveArchiveAcrossAppDirs(const std::string& fileName) {
+    for (const std::string& path : { Ship::Context::GetPathRelativeToAppDirectory(fileName, appShortName),
+                                     Ship::Context::GetPathRelativeToAppBundle(fileName), "./" + fileName }) {
+        std::error_code err;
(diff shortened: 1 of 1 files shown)


Malkierian:
Why do you have o2rs in multiple usable locations to begin with?

Malkierian:
Oh, wait, it wasn't actually removing from the locations that were checked is what you're saying.

Malkierian:
As a side note, you might want to change the plural "archives" to "archive" here and elsewhere as well, since the plural was a leftover from SoH.

Malkierian:
Oh, and maybe change it from "and it could not be automatically removed" to "but" instead. Bit more correct grammar.

Malkierian:
Also, you might want to grab my duplicate-parsing fix from https://github.com/HarbourMasters/Shipwright/pull/7063 which incorporated this PR for SoH.

Issue #1651, opened 2026-04-19
[Request] Bring back cosmetic color options for Double Defense hearts and Chateau magic only


SuperBro64:
I recently learned that a previous pull request (https://github.com/HarbourMasters/2ship2harkinian/pull/1562) significantly simplified the heart and magic meter cosmetic color options by reducing them down to just one for each.

While I understand the point made in that pull request about how most players would just randomize everything regardless, and the separate options for the heart/magic borders, beating heart, and consumed magic do seem excessive, I still think the pull request went too far in also lumping together the Double Defense hearts with normal hearts and the Chateau magic meter with the normal magic meter.

Compared to the other removed options, the DD hearts and Chateau magic feel different enough that I think they should still be available to change for greater customizability. Additionally, connecting the Chateau magic to the normal magic removes the fact that the original game used a separate color for the Chateau magic to make it easier to recognize beyond just the flashing border.

So all I'm suggesting here is to at least bring back in the cosmetic options for the Double Defense hearts and the Chateau magic meter.

SuperBro64:
Completed by #1891.

PR #1891, opened 2026-08-17
Moar Cosmetics

garrettjoecox:
woohoo.

- new cosmetics
  - Kaleido
  - Epona
  - Rupees
  - FD Tunic
  - Dungeon Items (stray/boss key/small key) (I sort of want to lock these by default.. but leaving them in default for now)
  - a few more
- New silly options
  - average texture shader (first custom shader example!)
  - longoron
  - a few more
- Fixes for recoloring some HD textures
  - I can't really guarantee this is going to work with all texture packs, it generally assumes your texture is going to be colored like a vanilla texture (eg your hair is a gradient of yellow). But it is at least verified to work with @GhostlyDark 's reloaded.
  - Zora recoloring is more complicated.. it doesn't look horrible but it doesn't look great, I will have to revisit it another time.
  - Rainbow mode with HD textures should be _way_ less laggy (still not great maybe, but way better)
- Moved the texture shading functions to a separate file

Will be blocked by https://github.com/Kenix3/libultraship/pull/1247

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9785938012.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9785924707.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9785840578.zip)

### DIFF
--- a/mm/src/code/z_player_lib.c
+++ b/mm/src/code/z_player_lib.c
@@ -50,6 +50,7 @@
 #include "2s2h/BenPort.h"
 #include "2s2h/GameInteractor/GameInteractor.h"
 #include "2s2h/Enhancements/FrameInterpolation/FrameInterpolation.h"
+#include "2s2h/BenGui/CosmeticEditor.h"
 #include <libultraship/bridge/consolevariablebridge.h>
 #include <libultraship/bridge/resourcebridge.h>
 
@@ -2501,6 +2502,8 @@ s32 Player_OverrideLimbDrawGameplayCommon(PlayState* play, s32 limbIndex, Gfx** 
         }
 
         if (limbIndex == PLAYER_LIMB_HEAD) {
+            CosmeticEditor_ApplySillyLimbScale(CVAR_COSMETIC("Silly.LinkHeadScale"), pos);
+
             rot->x += player->headLimbRot.z;
             rot->y -= player->headLimbRot.y;
             rot->z += player->headLimbRot.x;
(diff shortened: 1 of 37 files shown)


Jameriquiah:
bout time we got fd tunic lol good stuff

SuperBro64:
Thanks for your work on this! Would it be possible to also bring back the cosmetics for Double Defense hearts and Chateau magic that got removed in (https://github.com/HarbourMasters/2ship2harkinian/pull/1562), as I mentioned in this issue (https://github.com/HarbourMasters/2ship2harkinian/issues/1651)?

garrettjoecox:
> Thanks for your work on this! Would it be possible to also bring back the cosmetics for Double Defense hearts and Chateau magic that got removed in (#1562), as I mentioned in this issue (#1651)?

Sure

mm/src/code/z_player_lib.c
```diff
@@ -2699,6 +2703,8 @@ s32 Player_OverrideLimbDrawGameplayDefault(PlayState* play, s32 limbIndex, Gfx**
                 }
 
                 *dList = rightHandDLists[sPlayerLod];
+                CosmeticEditor_SplitHeldItem(dList, CVAR_COSMETIC("Silly.LinkShieldScale"));
```
Eblo, line 2706: Unless I'm missing something, this does not seem to actually scale shields.

garrettjoecox: Are you holding the shield? I didn't really want to bother with it on his back but maybe I should try
Screenshot 2026-08-24 at 11 00 12 PM

Eblo: oops, that would be it

PR #1896, opened 2026-08-20
Add vanilla option for moon trials

garrettjoecox:
This becomes the new default option.

For context, the reason I even added the 2-6-12-20 option was because I thought this was the closest we could get to the intent of vanilla, as for some reason I thought you couldn't play SoT on the moon in vanilla (I figured we specifically allowed it in rando). Turns out you're able to play SoT on the moon in vanilla, so my interpretation was flawed.

This also includes a fix for #1894 because now that bug could effect rando saves as well with this new option.

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9389679463.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9389648640.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9389627722.zip)

PR #1908, opened 2026-08-25
Reset masksGivenOnMoon on file load

### LINKED ISSUE #1894: masksGivenOnMoon not clearing with 2ship reset
image

JordanLongstaff:
Fixes #1894

I cut out that small piece of #1896 so it could be merged into the Battler branch.


Eblo (approved):
Thanks for being quick with this! I'll include this in 5.0.1.

PR #1916, opened 2026-08-29
Move some graveyard wonder items to correct region

garrettjoecox:
null

### Build Artifacts
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9719939192.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9719909089.zip)
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9719904951.zip)

### DIFF
--- a/mm/2s2h/Rando/Logic/Regions/East.cpp
+++ b/mm/2s2h/Rando/Logic/Regions/East.cpp
@@ -247,6 +247,9 @@ static RegisterShipInitFunc initFunc([]() {
             CHECK(RC_IKANA_GRAVEYARD_WONDER_ITEM_07, true),
             CHECK(RC_IKANA_GRAVEYARD_WONDER_ITEM_08, true),
             CHECK(RC_IKANA_GRAVEYARD_WONDER_ITEM_09, true),
+            CHECK(RC_IKANA_GRAVEYARD_WONDER_ITEM_10, true),
+            CHECK(RC_IKANA_GRAVEYARD_WONDER_ITEM_11, true),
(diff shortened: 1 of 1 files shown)


mm/2s2h/Rando/Logic/Regions/East.cpp
```diff
@@ -267,10 +270,6 @@ static RegisterShipInitFunc initFunc([]() {
     Regions[RR_IKANA_GRAVEYARD_UPPER] = RandoRegion{ .name = "Upper", .sceneId = SCENE_BOTI,
         .checks = {
             CHECK(RC_IKANA_GRAVEYARD_CAPTAIN_MASK, CanKillEnemy(ACTOR_EN_SKB) && CanKillEnemy(ACTOR_EN_BSB)),
-            // Rocks can be picked up and thrown at this:
```
Eblo, line 270: This should be the sign above Dampe's house, meaning this region is actually correct, right?

garrettjoecox: this is another one of those naming things where the name could probably be better. It is yes, but this region is for the part past keeta, with the giveaway being access to it is gated by sonata

Eblo: In that case, I *believe* the Bad Bat drop can be removed, since they don't appear in this particular region.

garrettjoecox: yup that sounds right, done

Eblo (approved):

PR #1918, opened 2026-08-30
Options to invert telescope aiming

### LINKED ISSUE #1684: Telescope y axis isn’t inverted when invert y axis in first person mode is on
I’m guessing the control schemes are really different under the hood but this setting should control both imo

JordanLongstaff:
Fixes #1684

I've linked the telescope aiming axes to the first-person view inversion options. That makes the most sense to me for the moment. However, the issuer has also suggested:

> Or maybe it should be its own toggle? Whatever you guys feel makes more sense :)

So I can add options specific to telescope mode if that's preferred.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9735456559.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9735445718.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9735443978.zip)


garrettjoecox:
I agree with it being mixed in with first-person

PR #1935, opened 2026-09-06
Reordered if checks to prevent null pointer reference and crash

CobraChicken997:
Ensures the If statement checks that the actor.parent is !NULL before checking if the parent has an update. Prevents a null pointer reference and a crash.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9989544454.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9989397851.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/9989384639.zip)


garrettjoecox:
Thanks for the PR! I assume you ran into this spawning this enemy with sail?

CobraChicken997:
Yeah that's exactly what happened! I'll probably look into spawning the angry cucco hordes more in the future too.

PR #1940, opened 2026-09-09
[Enhancement] Alien speed modifier

JordanLongstaff:
Added a slider to control how fast the aliens move during the Ranch Invasion minigame. Minimum 0.5x, maximum 5.0x.

Why 0.5x minimum? Because the aliens normally take 1 hour 20 minutes to reach the barn, and the minigame lasts for 2 hours 45 minutes. So lowering the minimum any further would literally make it impossible to fail. At that point, just use the skip checkbox right below.

Why 5.0x maximum? Because that reduces the aliens' travel time to the barn to just 16 minutes, which factors nicely. I play tested and was able to beat it with Inverted Song of Time. The next fastest I had considered was 8.0x, which gives you only 10 minutes to defeat each alien, but I decided that it was too fast, even with my computer being slow. Partly because of that, I think.

Anyone else is welcome to play test this if they want. 8.0x might be winnable if you're a video game god. Infinite Ammo attempts don't count IMO, though.

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/10090674538.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/10090670993.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/10090569590.zip)


JordanLongstaff:
@Eblo No conflicts with #1942 🙂

Eblo (approved):

PR #1947, opened 2026-09-10
[Enhancement] Difficulty options for Fisherman's Jumping Game

### LINKED ISSUE #1929: Difficulty Options for Fisherman's Minigame
maybe a slider for the amount of times you need to jump

@JordanLongstaff feel free to tackle this if you are looking for work, I'll get to it later this week if not

JordanLongstaff:
Fixes #1929

Options
Updated Text

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/10306082543.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/10305997926.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/10305573334.zip)

### DIFF
--- a/mm/2s2h/Enhancements/Minigames/FishermanJumpingGame.cpp
+++ b/mm/2s2h/Enhancements/Minigames/FishermanJumpingGame.cpp
@@ -4,10 +4,6 @@
 #include "2s2h/GameInteractor/GameInteractor.h"
 #include "2s2h/ShipInit.hpp"
 
-extern "C" {
-#include "overlays/actors/ovl_En_Jgame_Tsn/z_en_jgame_tsn.h"
-}
-
 static constexpr s32 TARGET_CVAR_DEFAULT = 20;
 #define TARGET_CVAR_NAME "gEnhancements.Minigames.FishermanJumpingGame.TargetScore"
 #define TARGET_CVAR CVarGetInteger(TARGET_CVAR_NAME, TARGET_CVAR_DEFAULT)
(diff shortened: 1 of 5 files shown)


JordanLongstaff:
FYI I had suggested changing the time limit for the minigame itself, but with the option to change the target score, I decided that would be redundant.

mm/2s2h/Enhancements/Minigames/FishermanJumpingGame.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+static constexpr s16 DOUBLE_TIME_LIMIT = 200;
+
+static bool ShouldWinJumpingGame() {
+    return gSaveContext.minigameScore >= TARGET_CVAR;
+}
+
+static void ModifyTargetScore(u16* textId, bool* loadFromMessageTable) {
+    auto entry = CustomMessage::LoadVanillaMessageTableEntry(*textId);
+    CustomMessage::Replace(&entry.msg, "20 or", std::to_string(TARGET_CVAR) + " or");
```
Eblo: When using `CustomMessage::Replace` we try to shy away from replacing words, as that will become a localization headache when more ROM versions are supported. Numerals alone would be fine, but in this case, `20` appears for both the target and the price, which we only want to modify one of in this case. The color codes are different however, so maybe this would work?

```suggestion
    CustomMessage::Replace(&entry.msg, "\x0120", "%r" + std::to_string(TARGET_CVAR));
```

Eblo (changes requested):

mm/2s2h/Enhancements/Minigames/FishermanJumpingGame.cpp
Eblo: My bad, grouping the escape sequence next to a number like that makes the compiler think it's a single escape sequence, and an invalid one at that. This should do it:

```suggestion
    CustomMessage::Replace(&entry.msg, "\x01" "20", "%r" + std::to_string(TARGET_CVAR));
```

Eblo (approved):

PR #1951, opened 2026-09-13
[Enhancement] Faster Hookshot

Eblo:
null

### Build Artifacts
  - [2ship-mac.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/10325310505.zip)
  - [2ship-linux.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/10325020483.zip)
  - [2ship-windows.zip](https://nightly.link/HarbourMasters/2ship2harkinian/actions/artifacts/10324209992.zip)


JordanLongstaff:
Why not make this a float slider instead?

Eblo:
> Why not make this a float slider instead?

The `timer` variable governs how long the Hookshot can extend and has a default value of 26. This normally decrements every frame as the Hookshot extends to find a target. The game checks if `timer` is equal to 0 to see if the Hookshot needs to retract. I tried playing with other values, but those would result in issues for others parts of the code that deal with `timer` and expectations against the actor's Y velocity.

These are things that could be addressed in theory, but given that doing so would require adding complexity, and that most players probably just want a faster Hookshot and 2x speed is decently faster, I decided to just go with the flat 2x multiplier.

